PR #9, opened 2019-12-08
Add semester modal and functionality

tcho6319:
### Summary
This pull request is for the implementation of the add semester feature.

- [x] refactored the add semester modal to match designs
- [x] implemented add semester functionality
- [x] moved the add semester modal from semester to semesterview
- [x] implemented add semester functionality

### Notes 
Some minor additional styling needed, such as adding emojis to an added semester's name, but will be resolved later.


src/components/Modals/Modal.vue
```diff
@@ -110,6 +113,14 @@ export default {
...
       // clear input and close modal when complete
       dropdown.value = '';
+      this.closeCurrentModal();
+    },
+    addSemester() {
+      const seasonInput = document.getElementById(`season-placeholder`);
+      const yearInput = document.getElementById(`year-placeholder`);
+
+      this.$parent.addSemester(seasonInput.innerHTML.split(' ')[1], parseInt(yearInput.innerHTML, 10));
+
```
SamChou19815: Why are we reading raw DOM html here? Vue should already have provided good solution to avoid this.

src/components/Modals/NewSemester.vue
```diff
@@ -2,63 +2,141 @@
...
+      if (contentShown) { // clicked box when content shown. So then hide content
+        content.style.display = 'none';
+        box.style.borderColor = '#C4C4C4';
+        arrow.style.borderTopColor = '#C4C4C4';
+      } else {
+        content.style.display = 'block';
+        box.style.borderColor = '#32A0F2';
+        arrow.style.borderTopColor = '#32A0F2';
+      }
+    },
```
SamChou19815: Do not use raw DOM manipulation to set style. 
This doc shows how to do it with Vue's style binding: https://vuejs.org/v2/guide/class-and-style.html

src/components/Modals/NewSemester.vue
```diff
@@ -2,63 +2,141 @@
...
-        }
+
+      const placeholderText = document.getElementById(placeholderID);
+      placeholderText.style.color = '#757575';
+
+      const arrow = document.getElementById(arrowID);
+      arrow.style.borderTopColor = '#C4C4C4';
+
+      const content = document.getElementById(contentID);
+      content.style.display = 'none';
```
SamChou19815: Same here, do not use DOM manipulation.

src/components/Modals/Modal.vue
SamChou19815: Fixing of this is more involved. It depends on better cross-component communication. Hacky HTML is a non-solution.

PR #14, opened 2019-12-11
Merge requirements

bryantleee:
### Merging the entire backend 

This pull request is for merging the requirement tracker functionality into the main course plan repo.

- [x] implemented the requirement tracker

### Test Plan 

- Testing with user feedback in alpha to see how accurate requirement tracker is


.eslintrc.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
     // TODO: fix copied js code that makes this warn necessary
     'no-use-before-define': ['warn'],
     'linebreak-style': 0,
+    //TODO: make requests asynchronous
+    'no-await-in-loop': ['warn'],
```
SamChou19815: This is a bad sign. It usually implies that you can do
```js
await Promise.all(
  list.map(
    item => expensiveAsyncOps(item)
  )
);
```
instead of
```js
for (const item of list) {
  await expensiveAsyncOps();
}
```

hanywang2:
@SamChou19815 removed the file

PR #26, opened 2020-01-16
Landing Page

ayesha4698:
### Summary 
The redesign of the login page to match Chris's Designs and with email configuration to firebase.


src/components/Landing.vue
```diff
@@ -0,0 +1,283 @@
+<template>
```
hanywang2: Why is there a landing and login component? Aren't they the same thing?

src/components/Login.vue
```diff
@@ -5,41 +5,114 @@
...
 
-          <!-- <div class="extras">
-          <a>Forgot Password</a>
-          <a>Create an Account</a>
-        </div> -->
-        </form>
-      </div>
-    </section>
+  </div>
   </div>
```
hanywang2: There should also be a footer

Issue #39, opened 2020-02-01
Represent insufficient permissions in the UI


SheaBelsky:
## Issue
If I log into CoursePlan and have insufficient permissions to do something, I should see that in the browser.

## Replication
1. Go to the web page
2. Log in with Google
3. When you're brought back to the site, nothing happens, but a console warning is logged:

```
FirebaseError: Missing or insufficient permissions.
    at new e (https://courseplan.io/js/chunk-vendors.f54b8f75.js:1664:524)
    at t.fromRpcStatus (https://courseplan.io/js/chunk-vendors.f54b8f75.js:2116:457)
    at t.fromWatchChange (https://courseplan.io/js/chunk-vendors.f54b8f75.js:2116:8934)
    at e.onMessage (https://courseplan.io/js/chunk-vendors.f54b8f75.js:2019:5977)
    at https://courseplan.io/js/chunk-vendors.f54b8f75.js:2019:4742
    at https://courseplan.io/js/chunk-vendors.f54b8f75.js:2019:5553
    at https://courseplan.io/js/chunk-vendors.f54b8f75.js:1745:9894
```

Issue #37, opened 2020-02-01
Mobile landing overlaps icons with text


SheaBelsky:
## Issue
On mobile, the icons on the landing page overlap with the text they are next to.

## Replication
1. Go to the landing page on a mobile device (or emulate with Inspect Element/Developer Tools)
2. Scroll down the landing page

<img width="307" alt="Screen Shot 2020-02-01 at 4 24 42 PM" src="https://user-images.githubusercontent.com/15164473/73599243-7c45da00-450f-11ea-805a-51fcf8790fc7.png">

PR #43, opened 2020-02-04
Landing page enhancements

willespencer:
### Summary 

This pull request deals with landing page issues, mainly #39 to alert users if they do not have alpha access, and #37 to fix mobile styling of the task images overlapping with text.

- [x] include an alert if a user logs in with an account without alpha access
- [x] remove overlap of task images on mobile
- [x] fix other responsive design issues like too small of email input on mobile, and awkward image sizes on medium breakpoints
- [x] fix security problems

reqs

alert

### Test Plan 

Look over the landing page on a variety of breakpoints to ensure everything looks fine. Also ensure emails without access are given an alert after attempting to login.

### DIFF
--- a/src/components/Login.vue
+++ b/src/components/Login.vue
@@ -159,15 +159,7 @@ export default {
         .signInWithPopup(provider)
         .then(user => {
           // Check whitelist emails to ensure user can log in
-          alphaWhitelistCollection.get().then(querySnapshot => {
-            if (!this.checkEmailAccess(querySnapshot, user)) {
-              return;
-            }
-
-            this.$store.commit('setCurrentUser', user.user);
-            this.$store.dispatch('fetchUserProfile');
-            this.$router.push(`${process.env.BASE_URL}/`);
-          });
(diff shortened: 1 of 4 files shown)


src/components/Login.vue
```diff
@@ -212,6 +175,22 @@ export default {
...
+      });
+      this.performingRequest = false;
+      if (!isAlphaEmail) {
+        fb.auth.signOut();
+        alert('Sorry, but you do not have alpha access.\nPlease sign up below for email updates on when the platform is available and for a chance to test the platform early.');
+        return false;
+      }
+
+      return true;
+    },
```
SamChou19815: This looks bad. You are implicitly exposing the alpha whitelist email to the entire world, since you need to loop over the entire snapshot.

SamChou19815: My suggestion:
1. Write firebase security rules so that user can only access his/her own whitelist document
2. Try to access that document
3. If caught insufficient permission error, then we know that the user is not whitelisted.

PR #44, opened 2020-02-04
Type annotate requirements functions

SamChou19815:
### Summary

Type annotate requirement functions so that they are easier to understand, with syntax described here: https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html
In this process, I found that the code is abusing JavaScript's feature of making `undefined` as falsy, so I tweaked the `RequirementFulfillment` object initializer to make them initialize to false so that it's more explicit.

### Test Plan

Add `@ts-check` to the top of the script section. Open VSCode. You will see that most the things are well-typed, except the first few imports. To make VSCode and TypeScript stop complaining about those imports, a more complete TS setup is necessary, but it should not be the focus of this diff. That's why `@ts-check` is not included in the diff.

### Notes

I think the requirement computation code should be moved to a separate file and make it a pure function, so that it's easier to reason and unit test.

### DIFF
--- a/src/components/Requirements.vue
+++ b/src/components/Requirements.vue
@@ -266,10 +266,18 @@ export default {
   },
 
   methods: {
+    /**
+     * @param {number} index
+     */
     toggleDetails(index) {
       this.reqs[index].displayDetails = !this.reqs[index].displayDetails;
     },
 
+    /**
+     * @param {number} index
+     * @param {'ongoing' | 'completed'} type
+     * @param {number} id
+     */
     toggleDescription(index, type, id) {
       if (type === 'ongoing') {
         const currentBool = this.reqs[index].ongoing[id].displayDescription;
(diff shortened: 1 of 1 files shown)


src/components/Requirements.vue
```diff
@@ -344,16 +364,44 @@ export default {
 
       return finalRequirementJSONs;
 
+      /**
+       * @typedef {Object} Requirement
+       * @property {string} name
+       * @property {string} description
+       * @property {string} source
+       * @property {string} search
+       * @property {string[][]} includes
```
tcho6319: @hanywang2 not entirely sure what the prop includes is for. I saw that below (line 565), there was some explanation but do you think we could put those when listing the properties of the different objects or would that be too verbose?

hanywang2: I don't think being too verbose is generally a bad thing for code documentation. `includes` works in conjunction with `search` to scan for courses that satisfy the requirement. For example:
- search = 'code' and includes = 'CS 2110' means the requirement is satisfied when the course code is CS 2110. 
- search = 'subject' and includes = 'INFO' means the requirement is satisfied when the subject is INFO. 
- search = 'catalogSatisfiesReq.' and includes = 'First-Year Writing Seminar.' means the reqyurement is satisfied when the phrase 'First-Year Writing Seminar.' is included in the catalogSatisifesReq attribute of a course's data.

The reason it's called include is because the way it's checked off is by if it's a substring of the course's data (ie if it's included). For example, 'First-Year Writing' can be used instead of 'First-Year Writing Seminar.' to check off the same requirement when search = 'catalogSatisifesReq'.

hanywang2: You can read more doc on this in the README.md in src/requirements

PR #45, opened 2020-02-08
Adding course fixes

tcho6319:
### Summary 

This pull request resolves a couple bug fixes found for adding courses from the alpha launch. 

- [x] Debugged adding supplementary course when trying to add main course (i.e. CHEM 1007 added when trying to add CHEM 2070)
- [x] Fixed adding courses for edge cases where null data props exist for fetched data from course roster API (i.e. unable to add MEDVL 1101)

### Notes
Read through requirements functions and referred to Cornell course roster API to diagnose the bugs. Implemented a secondary check when selecting fetched course to avoid accidentally adding supplementary courses and handled cases where fetched course data had null props that needed to be parsed.

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -152,8 +152,9 @@ export default {
       const credits = course.credits || course.enrollGroups[0].unitsMaximum;
 
       // Semesters: remove periods and split on ', '
-      // alternateSemesters option in case catalogWhenOffered for the course is null
-      const alternateSemesters = (course.catalogWhenOffered != null) ? course.catalogWhenOffered.replace(/\./g, '').split(', ') : [];
+      // alternateSemesters option in case catalogWhenOffered for the course is null, undef, or ''
+      const catalogWhenOfferedDoesNotExist = course.catalogWhenOffered === undefined || course.catalogWhenOffered === null || course.catalogWhenOffered === '';
+      const alternateSemesters = (catalogWhenOfferedDoesNotExist) ? [] : course.catalogWhenOffered.replace(/\./g, '').split(', ');
       const semesters = course.semesters || alternateSemesters;
 
       // Get prereqs of course as string (). '' if neither available because '' is interpreted as false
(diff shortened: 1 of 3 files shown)


src/components/Dashboard.vue
```diff
@@ -192,7 +194,9 @@ export default {
       }
 
       // Distribution of course (e.g. MQR-AS)
-      const distributions = course.distributions || course.catalogDistr.split(',');
+      // alternateDistributions option in case catalogDistr for the course is null
+      const alternateDistributions = (course.catalogDistr != null) ? course.catalogDistr.split(', ') : [""];
```
willespencer: Clean out parentheses

tcho6319: Used a regex to clean out the parentheses!

src/components/Dashboard.vue
```diff
@@ -192,7 +194,9 @@ export default {
       }
 
       // Distribution of course (e.g. MQR-AS)
-      const distributions = course.distributions || course.catalogDistr.split(',');
+      // alternateDistributions option in case catalogDistr for the course is null
+      const alternateDistributions = (course.catalogDistr != null) ? /\(([^)]+)\)/.exec(course.catalogDistr)[1].split(', ') : [""];
```
hanywang2: course.catalogDistr != null does not account for when course.catalogDistr is undefined. Consider making it either just (course.catalogDistr) ? ... or (course.catalogDistr === undefined || course.catalogDistr === nulll)

willespencer: The first suggestion @hanywang2 made will work as both null and undefined evaluate to false. Did not know this before, thanks for the suggestion!

tcho6319: Thanks for the suggestions! I had to end up doing the second way
(course.catalogDistr || course.catalogDistr === '') because there were some cases when course.catalogDistr could have been ''. I did the same changes for catalogWhenOffered.

PR #62, opened 2020-02-27
404 page

rmz38:
### Summary 

Made 404 page

- [x] Made 404 Vue component
- [x] Set up the router to display the 404 page

### Test Plan 

We put a random path in the url and made sure the 404 page pops up.

image


src/components/404.vue
```diff
@@ -0,0 +1,373 @@
...
+  width: 28px;
+  height: 23px;
+  left: 579.52px;
+  top: 320.16px;
+  background: #105351;
+}
+/* end second four */
+
+#oops {
+  position: absolute;
```
willespencer: Using absolute positioning here is really not great, especially because it looks way different on my screen than it did on yours (so it is not centered). Try to put both this and the new 404 image within the same div and center them on the screen. One way I would suggest to do that is by using flex, which if you've never used before, you should try my favorite tutorial ever. https://flexboxfroggy.com/

src/components/404.vue
```diff
@@ -0,0 +1,373 @@
...
+  width: 553px;
+  height: 134px;
+  left: 670px;
+  top: 207px;
+  font-family: Proxima Nova;
```
willespencer: I know Figma has font-family's like this, but it is a bit more complicated than this. We had to import font files to CoursePlan ourselves, and we set font-family in _global.scss. Check this file if you want to see how we set up the font families. You never need to set the font-family attribute in any other css file in our code because we set it globally in that file, so just delete this line (and the other ones throughout this file).

src/components/404.vue
```diff
@@ -0,0 +1,373 @@
...
+  font-style: normal;
+  font-weight: bold;
+  font-size: 60px;
+  line-height: 73px;
+  color: #0a2327;
+}
+#back_to_home {
+  /* Back to home */
+
+  position: absolute;
```
willespencer: Again, try to find a way to center this without using position absolute.

src/components/404.vue
```diff
@@ -1,373 +1,179 @@
...
+              </div>
+        </div>
     </div>
-    <div id="band">
-      <div id="first_four">
-        <div id="first_box1" class="color_box"></div>
-        <div id="second_box1" class="color_box"></div>
-        <div id="third_box1" class="color_box"></div>
-        <div id="fourth_box1" class="color_box"></div>
+    <!-- <img data-v-ef68022e="" src="img/logo.4d084161.svg" alt="logo" class="logo"> -->
```
willespencer, line 12: Please remove the commented out code here as well, and further in the file like in the CSS.

src/components/404.vue
```diff
@@ -1,373 +1,179 @@
...
+      <a class = "back_to_home_link" href = "/login">Back to home</a>
     </div>
-  </body>
-</html>
+    <div class = "footer-container">
+    </div>
+    <div class="container-fluid footer p-0 page_404">
+        <div class= "row footer justify-content-center m-0">
+            <div class="col-3 footer">
+              <!-- <a href ="https://www.cornelldti.org" class= "footer"> Cornell DTI @ </a> -->
```
willespencer, line 30: Commented out code here

src/requirements/reqs.json
```diff
@@ -2137,5 +2137,31 @@
                 }
             ]
         }
-    }
+    },
```
willespencer, line 2140: Make sure not to include this in this pull request since it is not related to the 4040 page. Could you remove the changes in this file?

src/components/404.vue
```diff
@@ -1,373 +1,179 @@
...
-    <div id="bottom_links">
-      <div id="copyright">Cornell DTI @ 2019</div>
-      <div id="love">Built with LOVE</div>
-      <div id="privacy">Privacy Policy</div>
+    <div class="back_to_home">
+      <a class = "back_to_home_link" href = "/login">Back to home</a>
     </div>
-  </body>
-</html>
+    <div class = "footer-container">
```
willespencer, line 25: Since its been a while, can you remind me why you did not end up just importing the Footer component? This works and looks good, but would definitely be better to avoid duplicating code.

rmz38: I remember that when just copying over the footer html + css there were issues with formatting so I had to change some of the css to work on this page. Should I still import it + is there a way to import and overwrite its css?

PR #63, opened 2020-02-28
Bottom bar updates

tcho6319:
### Summary 
This pull request is the updates for the bottom bar, mainly for the added implementation of tabs, debugging bottom bar issues from Alpha stage, and more closely matching the original designs.

- [x] Implemented Classes Tabs (also handling duplicate tabs)
- [x] Implemented See More Tab
- [x] Implemented X and arrows for closing, expanding, collapsing bottom bar
- [x] Style fixes (i.e. smaller width) to better match original designs
- [x] Fixed Bug: Bottom Bar expands when changing course color


src/components/BottomBar.vue
```diff
@@ -1,50 +1,87 @@
...
+    position:fixed;
+    bottom: 2.5rem;
+    left: 29.5rem;
+    width: calc(100vw - 29.5rem);
+  }
+
+}
+.expandedTabView{
+    position: fixed;
+    bottom: 18.81rem;
```
willespencer: Can this be rounded to something better?

src/components/BottomBarCourse.vue
```diff
@@ -134,69 +110,56 @@ export default {
...
-    }
-
     &-bar {
       width: 100%;
 
       &-info {
         float: left;
-        height: 261px;
-        width: 35%;
+        height: 16.31rem;
```
willespencer: Same as above. Should start with at least an even number pixel value (260 or 262) before dividing by 16. Thanks for switching to rem though :)

src/components/BottomBarTabView.vue
```diff
@@ -0,0 +1,239 @@
...
+              v-for="seeMoreCourse in seeMoreCourses"
+              :key="seeMoreCourse.id"
+              class="seeMoreCourse-option"
+            >
+                <span class="seeMoreCourse-option-text" @click="moveToBottomBar(seeMoreCourse)">{{ seeMoreCourse.subject }} {{ seeMoreCourse.number}}</span>
+                <img class="seeMoreCourse-option-delete" src="@/assets/images/x-blue.svg" @click="deleteSeeMoreCourse(seeMoreCourse)"/>
+            </div>
+        </div>
+      </div>
+
```
willespencer: Ok nitpick but a few random spaces between div tags here

src/components/BottomBarTabView.vue
```diff
@@ -0,0 +1,239 @@
...
+  },
+
+  methods: {
+    bottomBarTabToggle(courseObj) {
+      this.$emit('bottomBarTabToggle', courseObj);
+    },
+
+    deleteBottomTab(subject, number) {
+      for (let i = 0; i < this.bottomCourses.length; i += 1) {
+        if (this.bottomCourses[i].subject === subject && this.bottomCourses[i].number === number) {
```
willespencer: This code relies on there only being one version of a course in the expanded view of the course. Is this a problem if we are now fine with allowing people to have multiple of the same course? Especially if they are different colors?

src/components/Dashboard.vue
```diff
@@ -15,27 +15,40 @@
...
           :semesters="semesters"
           :user="user"
-          :isBottomPreview="bottomBar.isPreview"
-          :isBottomBar="bottomBar.isExpanded"
           :key="requirementsKey"
           @requirementsMap="loadRequirementsMap"
          />
       </div>
-      <semesterview v-if="loaded"
+      <!-- TODO: isBottomBar fix -->
```
willespencer: What is this commented code for?

src/components/Dashboard.vue
```diff
@@ -360,23 +372,56 @@ export default {
...
 
+      // expand bottombar if first course added
+      if (this.bottomCourses.length === 0) {
+        this.bottomBar.isExpanded = true;
+      }
+
+      // if course already exists in bottomCourses, first remove course
+      for (let i = 0; i < this.bottomCourses.length; i += 1) {
+        // if colorJustChanged and course already exists, just update course color
+        if (this.bottomCourses[i].subject === course.subject && this.bottomCourses[i].number === course.number && colorJustChanged) {
```
willespencer: Same concern as above and in a lot of places in this file.

src/components/BottomBar.vue
tcho6319: I used an even pixel size and reconverted to rem and got 18.75rem. Is that ok?

src/components/BottomBarCourse.vue
tcho6319: Same as above and now it's 16.25rem

src/components/Dashboard.vue
tcho6319: Removed

src/components/BottomBarTabView.vue
tcho6319: At the moment, no dupe courses are allowed for bottom tab including see more. If a user does have multiples of the same course (say CS 2110) and they click on the course card again, if CS 2110 is already a bottom bar tab, it is brought to the very front (with the color of that clicked course card). I don't think it would be very useful to have dupe bottom tabs even though they can have dupe course cards since the expanded bottom bar course details would be the same. Since there would only be one ver of a course, my delete functionality relied on this. Or are we saying that expanded course details could differ for the same course?

tcho6319: If we want course identification to be more unique, we could assign/use a unique id prop?

src/components/BottomBar.vue
tcho6319: This is for the See More screenshot but I noticed that the scrollbar might cover a part of the X on hover if a user wants to delete a See More course. Should I move the X more inward?

willespencer: This all sounds good to me, thanks Theresa

src/components/BottomBarTabView.vue
willespencer: Discussed over slack - Want a unique id integer that starts at 0, and increases every time a course is added, so then there is never a duplicate. We can store and retrieve it from firebase per user, so that it always increases for a single user. Can be done in a separate PR, because it functionally works fine for bottom bar, but not across the rest of the platform.

PR #66, opened 2020-02-29
Update freshman writing seminars in requirements

hanywang2:
### Summary 

Update freshman writing seminars requirement data to read from titleLong (as "FWS:") rather than catalogReqSatisfies.

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -144,7 +144,6 @@ export default {
      * Creates a course on frontend with either user or API data
      */
     createCourse(course) {
-      // TODO: id?
       const randomId = Math.floor(Math.random() * Math.floor(1000));
 
       const subject = (course.code && course.code.split(' ')[0]) || course.subject;
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -144,6 +144,7 @@ export default {
      * Creates a course on frontend with either user or API data
      */
     createCourse(course) {
+      // TODO: id?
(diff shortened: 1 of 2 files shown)


src/components/Dashboard.vue
```diff
@@ -144,7 +144,6 @@ export default {
      * Creates a course on frontend with either user or API data
      */
     createCourse(course) {
-      // TODO: id?
       const randomId = Math.floor(Math.random() * Math.floor(1000));
```
willespencer: Can you actually leave a TODO here and make an item to look at it? This random ID thing seems very concerning to me

src/requirements/reqs.json
```diff
@@ -365,13 +365,13 @@
                     "fulfilledBy": "self-check"
                 },
                 {
-                    "name": "Freshman Writing Seminars",
+                    "name": "First-Year Writing Seminars",
                     "description": "2 writing classes. Any two writing classes from the following: First-Year Writing Seminars (FWS), ENGL 2880, ENGL 2890",
                     "source": "https://aap.cornell.edu/academics/art/undergraduate/curriculum",
-                    "search": ["catalogSatisfiesReq"],
+                    "search": ["titleLong"],
```
willespencer: Do we now only want to search titleLong instead of both the req and title?

src/components/Dashboard.vue
hanywang2: O lmao yea I just thought we were using randomId for some reason.

PR #72, opened 2020-03-05
Strictify type annotation for requirement function

SamChou19815:
### Summary

Thanks to Will's effort in #71, we are only a few steps away from making the requirement function in TypeScript. The first step is to enable type checking based on JSDoc annotation, which is implemented in this diff. Previously this is impossible due to other non-strict functions in `Requirements.vue` file.

Now with a stronger type definition of requirements established in #68, TypeScript is able to find more type errors due to incorrect type definition setup in #44. I fixed these errors.

I have to create a separate `typed-requirement-json.ts` that basically just re-exported the json with a type assersion, and make `reqs-functions.js` import this file instead. This hack is necessary because TypeScript widens the `fulfillBy` enum to string when it's inferring the type of json (more background in #68).

### Test Plan

`yarn tsc` still type checks, but now we have more type coverage!

### Notes

In the future, we can

- Convert `reqs-functions.js` to TypeScript simply by changing the file extension and moving jsdoc type annotations to function parameters
- Strictify the types of these functions even more. e.g. Replace all `any`s with actual type definitions.
- Make `iterateThroughRequirements` non-async. It's probably async due to some legacy code.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -86,6 +86,20 @@ module.exports = {
           }
         ]
       }
+    },
+    {
+      files: ['*.js'],
+      rules: {
+        'import/extensions': [
+          'error',
+          'always',
+          {
+            js: 'never',
+            ts: 'never',
+            vue: 'never'
+          }
+        ]
+      }
     }
   ]
 };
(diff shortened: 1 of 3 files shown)


SamChou19815:
Already contained in #73

PR #73, opened 2020-03-07
Introduce proof-of-concept requirement pre-computation code

SamChou19815:
### Summary

According to our discussion in the work session today, pre-computation can help to eliminate the invocations of Cornell's course API on frontend, as well as automatically get a satisfiable course list.

Creating a diff that directly changes everything is hard. Therefore, I decide to create a proof-of-concept implementation. More specifically, this diff does the following stuff:

- [x] Fetch all classes from all semesters, drop unnecessary fields, and store them in a json.
- [x] Setup Infrastructure to compute and store a "decorated" requirement json, where the each requirement is decorated with a list of class names that can satisfy it.

What's not implemented is the function that computes whether a course can satisfy the requirement. I stubbed it with a function that returns true with 1% probability. I mostly want to use it to estimate the size of the generated json. I will talk about the json size later. The idea is that implementing these checks are not hard: we can just call the existing functions in `src/requirement/req-functions.js`. However, making it work will involving moving some files around, which don't plan to touch in this diff.

Now we get to the size of the json. The `fetcher.ts` script generates all courses and its size is 2.0M (with current configuration with a lot of fields filtered away). This size actually does not matter, because it is never served to the end user. What's more important is `decorated-requirements.json`, which is less than 800K. This is a very promising result, if you look at it: each requirement is attached 100+ courses with the current randomized setup. Real requirements that we plan to deal with might only have 10 courses that are associated with it, so the real generated `decorated-requirements.json` might be less than 200K, which can be served to end user easily. What's even more amazing is that running `requirement-satisfiable-course-generator.ts` finishes almost instantly, so we can experiment with the script very fast!

Hence, according to the analysis above, I believe it's a good idea to continue along the path of requirement pre-computation, and this diff is a significant first step.

### Test Plan

To reproduce my finding, do the following:

1. Run `npm run build:req-gen` to compile all TS files in `requirement-generator`.
2. Run `node requirement-generator/dist/requirement-generator/fetcher.js` to fetch all classes (this may take 20 minutes)
3. run `node requirement-generator/dist/requirement-generator/requirement-satisfiable-course-generator.js` to generate the decorated requirement json.

### DIFF
--- a/requirement-generator/fetcher.ts
+++ b/requirement-generator/fetcher.ts
@@ -1,0 +1,123 @@
+// Disabled because we want to log progress.
+/* eslint-disable no-await-in-loop */
+// Disabled because we want the inefficiency to rate-limit ourselves.
+/* eslint-disable no-console */
+import fetch from 'node-fetch';
+import { writeFileSync } from 'fs';
+import { Course } from './types';
+
+const PREFIX = 'https://classes.cornell.edu/api/2.0';
+
+const wait = (time: number) => new Promise(resolve => setTimeout(() => resolve(), time));
+
+const getSemesters = async (): Promise<readonly string[]> => {
+  const response = await fetch(`${PREFIX}/config/rosters.json`);
+  const json = await response.json();
+  return json.data.rosters.map((it: { slug: string }) => it.slug);
+};
+
+const getSubjects = async (semester: string): Promise<readonly string[]> => {
+  try {
+    const response = await fetch(`${PREFIX}/config/subjects.json?roster=${semester}`);
+    const json = await response.json();
+    return json.data.subjects.map((it: { value: string }) => it.value);
+  } catch {
+    return [];
+  }
+};
+
+type CourseFieldFilter<T extends keyof Course> = (course: Course) => Pick<Course, T>;
+
+const getCourseFieldFilter = <T extends keyof Course>(allowedFields: T[]): CourseFieldFilter<T> => (
+  course: Course
+) => {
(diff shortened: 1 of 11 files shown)


requirement-generator/requirement-satisfiable-course-generator.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const courseCanSatisfyCollegeRequirement = (
+  course: FilteredCourse,
+  majorRequirement: CollegeRequirement<StrictFulfilledByType>
+): boolean => Math.random() < 0.01;
+
+const courseCanSatisfyMajorRequirement = (
+  course: FilteredCourse,
+  majorRequirement: MajorRequirement<StrictFulfilledByType>
+): boolean => Math.random() < 0.01;
```
hanywang2: Don't want to keep the randomized method for filtering courses on master for too long

requirement-generator/fetcher.ts
```diff
@@ -0,0 +1,123 @@
...
+      console.log(`We fetched ${semesterCount} out of ${semesters.length} semesters.`);
+    }
+  }
+  if (doPrintDebuggingInfo) {
+    console.log(`Total Running Time: ${new Date().getTime() - startTime}ms.`);
+  }
+  return courses;
+};
+
+const courseFieldFilter = getCourseFieldFilter(['subject', 'catalogNbr']);
```
hanywang2: add every field that is a search code that is currently used on reqs.json

PR #74, opened 2020-03-08
Edit Credits for Courses with Credit Ranges

tcho6319:
### Summary 
This work-in-progress pull request is for the functionality of editing credit values for courses with credit ranges (i.e. user setting the course credit value for INFO 1998). 

### Issues
- [] Accessing course ranges (using enrollGroups) for user data where enrollGroups/credit range does not exist.


rmz38:
Added ability to edit the credits of courses with a credit range, such as CS 4999 and INFO 1998. The option to edit a course will appear in the course menu if the course has a range, otherwise it will not appear. This was done by adding a creditRange field to createCourse in Dashboard, updating the firebase course with the credit Range, and adding a section with a dropdown in a course menu.

I've tested this on courses where the range starts at 1 (CS 4999's range is 1-4) and courses where the range does not start at 1 (PSYCH 3420's range is 3-4).

Issues: Fulfilled credits in the Requirements Menu don't update, but when I worked off the page404 branch I remember it did, so maybe this will not be an issue when merged.

After merging, developers have to delete and readd all their courses as well to update firebase courses with the new credit range.

image
image

hanywang2:
Wait @rmz38 I don't believe that the ability to edit credits is in the page404 branch. Also, does it support courses that have been entered prior to this change?

rmz38:
> Wait @rmz38 I don't believe that the ability to edit credits is in the page404 branch. Also, does it support courses that have been entered prior to this change?

I made some commits meant for this branch to page404 but I removed them and put them here, so the ability to edit credits shouldn't be on page404. Unfortunately it doesn't support courses that were entered before because a course in firebase didn't store the credit range before.

hanywang2:
> > Wait @rmz38 I don't believe that the ability to edit credits is in the page404 branch. Also, does it support courses that have been entered prior to this change?
> 
> I made some commits meant for this branch to page404 but I removed them and put them here, so the ability to edit credits shouldn't be on page404. Unfortunately it doesn't support courses that were entered before because a course in firebase didn't store the credit range before.

I asked because of the issue with the requirements bar credits not updating. Can we look into that and get it fixed cause I believe that's all we need for customizing credits to get merged in

src/components/Dashboard.vue
```diff
@@ -43,6 +43,7 @@
 <script>
 import Vue from 'vue';
 
+// import undefined from 'firebase/empty-import';
```
willespencer, line 46: Remove this comment?

src/components/Dashboard.vue
```diff
@@ -140,11 +141,30 @@ export default {
...
+        return [course.credits, course.credits];
+      }
+      return course.creditRange;
+    },
+
     /**
      * Creates a course on frontend with either user or API data
      */
     createCourse(course) {
-      // TODO: id?
```
willespencer, line 147: Can you leave this TODO? It's a problem that still needs to be addressed.

src/components/Dashboard.vue
```diff
@@ -140,11 +141,30 @@ export default {
...
     },
 
+    /**
+     * Creates credit range based on course
+     * Example: [1, 4] is the credit range for the given course
+     */
+    createCourseCreditRange(course) {
+      const courseCreditRange = [];
+      if (typeof course.creditRange !== 'undefined') {
+        const ccr = course.creditRange; // the linter won't let me return with course.creditRange so im just doing this
```
willespencer, line 151: What are you trying to do here? Just return course.creditRange here?

src/components/Modals/CourseMenu.vue
```diff
@@ -29,6 +29,30 @@
...
         </div>
       </div>
+      <div
+        class="courseMenu-section"
+        @mouseover="setDisplayEditCourseCredits(true)"
+        @mouseleave="setDisplayEditCourseCredits(false)"
+        v-if="this.getCreditRange[0] != this.getCreditRange[1]"
+        >
+        <div class="courseMenu-left">
+          <img class="courseMenu-icon" src="../../assets/images/edit-credits.svg" />
```
willespencer, line 39: Better to use the "@" symbole to get to the src directory than "../../"

src/components/Modals/CourseMenu.vue
```diff
@@ -144,10 +184,14 @@ export default {
...
     }
   }
-
   &-colors {
     position: absolute;
     right: -9rem;
   }
+  &-editCredits {
+    position: absolute;
+    width: 2.7rem;
```
willespencer, line 193: Good practice to use more precise rem values - so something like 2.5 rem or 2.75 rem that match px values (since 16 px = 1 rem)

src/components/Semester.vue
```diff
@@ -229,6 +230,15 @@ export default {
...
       this.$emit('updateBar', course);
     },
+    editCourseCredit(credit, courseCode) {
+      for (let i = 0; i < this.courses.length; i += 1) {
+        if (`${this.courses[i].subject} ${this.courses[i].number}` === courseCode) {
+          this.courses[i].credits = credit;
+          break;
+        }
+      }
+      this.$parent.$parent.updateRequirementsMenu();
```
willespencer, line 240: Can you make this an $emit instead of referencing parent twice?

src/components/Dashboard.vue
rmz38: yeah but eslint doesn't let that

src/components/Modals/CourseMenu.vue
rmz38: I was using the same filepaths as the other icons should I change those too?

src/components/Dashboard.vue
```diff
@@ -178,6 +198,7 @@ export default {
         const enrollmentMap = {};
         const lectureTimesMap = {};
         const instructorsMap = {};
+        console.log(course.enrollGroups);
```
willespencer, line 200: Console 🙃

rmz38: O:

PR #75, opened 2020-03-08
Multiple Majors and Minors

ayesha4698:
### Summary 
This pull request is the first step towards implementing feature of having multiple majors and minors on the requirements bar and having an edit section. 

- [x] implemented dual tab majors and minors.
- [x] implemented the WIP input form for users to edit their requirements.

### Test Plan 
Nothing seems to break now


src/components/Modals/Onboarding.vue
```diff
@@ -194,7 +193,7 @@
 <script>
 import reqsData from '@/requirements/typed-requirement-json';
 
-const placeholderText = 'Select one';
+const placeholderText = ['Select one'];
```
willespencer, line 196: Any reason why the placeholder text includes brackets? Looks a bit off on the onboarding modal

src/components/Modals/Onboarding.vue
```diff
@@ -319,7 +350,22 @@ export default {
...
+      const minorJSON = reqsData.minor;
+      for (const key in minorJSON) {
+        // make sure name defined
+        if ('name' in minorJSON[key]) {
+          // only show majors for schools the user is in
+          for (let i = 0; i < this.displayOptions.college.length; i += 1) {
+            const college = this.displayOptions.college[i];
+            if (minorJSON[key].schools.includes(college.acronym)) {
+              minors[key] = minorJSON[key].name;
+              // continue;
```
willespencer, line 363: Delete commented code if it is not being used.

src/components/Modals/Onboarding.vue
```diff
@@ -465,8 +509,20 @@ export default {
         arrowColor: '',
         placeholderColor: '',
         placeholder: placeholderText,
+        // fixxxxx
```
willespencer, line 512: hmmmm 🤔

src/components/Requirements.vue
```diff
@@ -1,149 +1,199 @@
...
-      <div class="progress">
-        <div
-          class="progress-bar"
-          :style="{ 'background-color': `#${req.color}`, width: `${(req.fulfilled/req.required)*100}%`}"
-          role="progressbar"
-        ></div>
+      <!-- TODO change for multiple colleges -->
+      <div v-if="index<=2 || index == 2 + majors.length" class="row top">
+        <p class="name col p-0">{{ req.name }} <span class="specific" v-if="req.specific">({{ req.specific }})</span> </p>
+        <img  @click="openEditMenu(index)" class="gear--disabled" src="../assets/images/gear.svg" />
```
willespencer, line 47: Can you remove the gear svg from reqs entirely? Talked to designers and we aren't implementing it, so no point in including the button at all.

src/components/Requirements.vue
```diff
@@ -1,149 +1,199 @@
 <template v-if="semesters">
-  <div class="requirements">
-    <div class="fixed">
-    <h1 class="title">School Requirements</h1>
+  <div class= "section">
+  <div v-if="isEditing" class= "editing">
```
willespencer, line 3: Could you leave a comment at the top of this section just saying that this is the editing section if we were to implement it later, but right now this section never displays?

src/components/Modals/Onboarding.vue
```diff
@@ -288,8 +287,40 @@ export default {
...
+      this.displayOptions.major.forEach(major => {
+        if (Array.isArray(major.acronym)) {
+          major.acronym.flat(Infinity);
+          for (let i = 0; i < major.acronym.length; i += 1) {
+            const newMajor = {
+              shown: false,
+              stopClose: false,
+              boxBorder: '',
+              arrowColor: '',
+              placeholderColor: '',
```
willespencer, line 304: One problem I noticed on onboarding is that after you select a major, close onboarding, and re-open it, any majors selected have gray text instead of black text. Could there be a problem here or somewhere else in your added code that does not set the color properly?

willespencer:
One more bug I just noticed ! On courseplan.io we can select a college but not a major and the reqs menu still works, but it does not on this branch. Can you allow people to not select a major?

src/components/Modals/Onboarding.vue
ayesha4698: opps

src/components/Requirements.vue
```diff
@@ -1,149 +1,163 @@
 <template v-if="semesters">
+  <div class= "section">
   <div class="requirements">
-    <div class="fixed">
+    <div class="fixed" :class="{ bottomPreview: isBottomPreview && !isBottomBar, bottomBar: isBottomBar }">
```
willespencer, line 4: There seems to be an error where isBottomPreview and isBottomBar is not defined. Any reason these classes were added?

src/components/Requirements.vue
```diff
@@ -1,149 +1,163 @@
...
             </div>
-            <div class="col-7" @click="toggleDescription(index, 'ongoing', id)">
-              <p class="sup-req pointer">{{subReq.requirement.name}}</p>
+            <div class="col p-0 ">
+              <button class="btn req-name" :style="{ 'color': `#${req.color}` }" @click="toggleDetails(index)">{{ (req.displayDetails) ? "HIDE" : "VIEW" }} ALL {{ req.group }} REQUIREMENTS</button>
             </div>
-            <div class="col">
-              <p class="sup-req-progress text-right">
-                ( {{
-                  (subReq.requirement.fulfilledBy !== 'self-check')
```
willespencer, line 81: Was this and a bunch of other things in this file meant to be changed? This deletion seems to be causing every requirement to be a "self-check," and if I switch the code back to this from what it was changed to, it fixes it. Do you know if that is the only change that needs to be reverted, or is a lot of the code in here overwriting things that should not be?

src/components/Requirements.vue
ayesha4698: This was here before Theresa changes so I wasn't sure what to do with it lol but it doesn't break if deleted so I just did.

src/components/Requirements.vue
ayesha4698: Ahh yeah I had deleted this before I realized the Reqs Json had changed quite a bit-- should be fixed now.

PR #76, opened 2020-03-09
Fully implement requirement checking with the new pre-compute design

SamChou19815:
### Summary 

Make requirement checking to use pre-computed satisfiable course list.

To help review, I will walk through the higher-level steps I created the diff:

1. split the current `req-functions.js` into two parts: `requirement-checker.ts` and `reqs-functions.ts`
  - `requirement-checker.ts` is responsible for checking whether a course (with full information) satisfies a requirement. This file goes into `requirement-generator`
  - `reqs-functions.ts` is responsible for checking whether a class satisfies a requirement by checking whether the course code is included in the precomputed `courses` array, as well as computing a requirement-fulfillment array for other parts of the frontend.
2. Connect `requirement-checker.ts` to `requirement-satisfiable-course-generator.ts`. Now the generator completely works.
3. Connect slightly tweaked `reqs-functions.ts` to `Requirement.vue`. Tweaks:
  - Now it requires the vue component to pass in more information, like the number of credits for a class. Fortunately, the information is already available.
  - The university requirement (academic credit) can let almost all classes fulfill it. Putting the names of all those classes into the pre-computed array is a waste of space. Therefore, I just hardcoded the logic of testing university requirement satisfiability.
4. Profit!

### Test Plan 

Everything still runs, but the `getReqs` function is not longer `async`. It runs faster!

Screen Shot 2020-03-09 at 03 35 48

### Notes 

I decide to check in the generated `decorated-requirements.json` into the repo because:

- It's expensive to generate (most of the cost is on generating courses)
- Once the algorithm is stable, this file doesn't tend to change very often

### Future Plans

There are some immediate future work plans:

1. Purge unused information in the generated `decorated-requirements.json`
2. Remove unused fields in the type definition of requirements.
3. Optimize the format of `decorated-requirements.json` to reduce its size.

`decorated-requirements.json` now is over 2MB. The primary reason is that a weird `BE Focus Area Elective` has `"search": ["all-eligible"]`, so its courses array contains all courses across all semesters. We might need to special case for those requirements that tend to have a lot of satisfiable courses.


requirement-generator/fetcher.ts
```diff
@@ -85,13 +85,15 @@ const getAllCoursesInSemester = async <T extends keyof Course>(
...
 
+type AllCourses<T extends keyof Course> = { [semester: string]: readonly Pick<Course, T>[] };
+
 const getAllCourses = async <T extends keyof Course>(
   courseFieldFilter: CourseFieldFilter<T>,
   coolingTimeMs: number = 50,
   doPrintDebuggingInfo: boolean = true
-): Promise<readonly Pick<Course, T>[]> => {
+): Promise<AllCourses<T>> => {
   const startTime = new Date().getTime();
```
hanywang2: Is this ever used?

requirement-generator/fetcher.ts
```diff
@@ -116,7 +118,21 @@ const getAllCourses = async <T extends keyof Course>(
   return courses;
 };
 
-const courseFieldFilter = getCourseFieldFilter(['subject', 'catalogNbr']);
+const courseFieldFilter = getCourseFieldFilter([
```
hanywang2: What is the process for retrieving course credits for requirements calculations again?

hanywang2: Is it directly from user data?

requirement-generator/requirement-satisfiable-course-generator.ts
```diff
@@ -1,75 +1,57 @@
...
+    requirement.search,
+    requirement.includes,
+    requirement.excludes || []
+  );
+}
+
+const produceSatisfiableCourseIdForCollegeOrMajorRequirement = (
+  collegeRequirement: CollegeOrMajorRequirement
 ): readonly string[] => {
   const set = new Set<string>();
```
hanywang2: Could we change set to another variable name with more clarity?

requirement-generator/requirement-sources-jsons.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Course, RequirementsJson } from './types';
+
+/** All courses fetch from Cornell Class API, with only the fields we need. */
+export const filteredAllCourses: { readonly [semester: string]: readonly Course[] } = JSON.parse(
+  readFileSync('requirement-generator/filtered-all-courses.json').toString()
+);
+
+/** The requirement json that guides our algorithm to decide whether a class satisfies a requirement. */
+export const requirementJson: RequirementsJson = JSON.parse(
+  readFileSync('src/requirements/reqs.json').toString()
```
hanywang2: Is readFileSync and toString() the best option compared to `require('src/requirements/reqs.json')`

requirement-generator/requirement-sources-jsons.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * The module that re-exposes the generated jsons with strict static types.
+ */
+
+import { readFileSync } from 'fs';
+import { Course, RequirementsJson } from './types';
+
+/** All courses fetch from Cornell Class API, with only the fields we need. */
+export const filteredAllCourses: { readonly [semester: string]: readonly Course[] } = JSON.parse(
+  readFileSync('requirement-generator/filtered-all-courses.json').toString()
```
hanywang2: Same case here

requirement-generator/types.ts
```diff
@@ -1,19 +1,76 @@
 export type Course = {
```
hanywang2: Amazing stuff with setting up types

requirement-generator/fetcher.ts
SamChou19815: This is used to generate the courses json

requirement-generator/fetcher.ts
SamChou19815: Yes, from user data.

requirement-generator/requirement-sources-jsons.ts
SamChou19815: A little technical background: this `requirement-generator` folder is currently an isolated project from the `src/` folder. Using `require` will break typescript import. In the future, I plan to merge them together when things are stabilized.

requirement-generator/requirement-sources-jsons.ts
SamChou19815: Haha, I did this for a different reason.
This file is *generated* by fetcher.ts, so a direct import will break linter and TS if you don't have the generated file in place. This `filtered-all-courses.json` file is now 40MB now. I'm debating and weighting whether we should check it into the repo.

requirement-generator/requirement-sources-jsons.ts
hanywang2: That's fair thank you

requirement-generator/fetcher.ts
hanywang2: Apologies, I meant to specify `startTime`

hanywang2: But nvm, I see that it's used for console logging

requirement-generator/requirement-sources-jsons.ts
hanywang2: That makes sense got it

PR #77, opened 2020-03-09
Simple requirement code cleanup

SamChou19815:
### Summary 

- [x] Removed some unused fields from type definition
- [x] Optimize space efficiency for generated requirement json

See the commit comments for more details

### Test Plan 

Everything still works.

### Notes

Do not merge #76 is merged. I will change the base branch once #76 is merged.


requirement-generator/types.ts
```diff
@@ -45,7 +45,7 @@ export interface CollegeOrMajorRequirement extends BaseRequirement {
 }
 
 export interface DecoratedCollegeOrMajorRequirement extends BaseRequirement {
-  readonly courses: readonly string[];
+  readonly courses: 'all-eligible' | readonly string[];
```
hanywang2: Is there a reason that courses could either be a string or list? What if it's ['all-eligible']?

src/components/Modals/Onboarding.vue
```diff
@@ -192,13 +192,13 @@
 </template>
 
 <script>
-import reqsData from '@/requirements/reqs.json';
```
hanywang2: Just for clarification, this won't change our workflow for data entry because it reads from reqs.json at requirement-sources right?

requirement-generator/types.ts
SamChou19815: It won't be. The case of string list is literally a list of all courses, like
```
["CS2112", "CS2800"]
```

src/components/Modals/Onboarding.vue
SamChou19815: It will. Now you will need to change it in `reqs.json`, and run the generator in `requirement-generator` folder.

hanywang2: I see. So just for adding data, this won't change anything

requirement-generator/types.ts
hanywang2: Right, but it could also be 'all-eligible'?

src/components/Modals/Onboarding.vue
SamChou19815: Yes

hanywang2: Awesome. All good now

requirement-generator/types.ts
SamChou19815: The `getSatisfiableCourses` is the function that generates the array. It will never put `all-eligible` in that array,

PR #78, opened 2020-03-09
Compress decorated requirement json

SamChou19815:
### Summary

A large source of the big size of the json is caused by repetition of names. For example, if there is a requirement that all CS courses are eligible, then the word `CS` will be repeated for n * m times, where n = number of semesters, m =  number of CS courses.

To optimize this, we can store the classes in a tree structure rather than a flattened giant list. You can see the tree structure type definition in `src/requirements/types.ts`'s `EligibleCourses` type. With this structure, a lot of information are deduplicated! It also allows a much more efficient search, since we can avoid an O(n) search for each requirement.

The result of this optimization is amazing.

### Test Plan 

The json size reduced from 1.4M to less than 700K, and everything still works.

### Notes

Do not merge #77 is merged. I will change the base branch once #77 is merged.


requirement-generator/requirement-satisfiable-course-generator.ts
```diff
@@ -1,5 +1,12 @@
 import { writeFileSync } from 'fs';
-import { Course, CollegeOrMajorRequirement, DecoratedRequirementsJson } from './types';
+import {
+  Course,
+  CollegeOrMajorRequirement,
+  DecoratedRequirementsJson,
+  UniversityRequirements,
+  DecoratedCollegeOrMajorRequirement,
+  EligibleCourses
+} from './types';
```
hanywang2: Amazing stuff with Course types now that we have a more defined data structure

requirement-generator/requirement-satisfiable-course-generator.ts
```diff
@@ -17,25 +24,50 @@ function courseCanSatisfyCollegeOrMajorRequirement(
...
+): EligibleCourses => {
+  const allMap: { [semester: string]: { [subject: string]: string[] } } = {};
   Object.entries(filteredAllCourses).forEach(([semester, courses]) => {
+    const semesterMap: { [subject: string]: string[] } = {};
     courses
       .filter(course => courseCanSatisfyCollegeOrMajorRequirement(course, collegeRequirement))
-      .forEach(it => set.add(`${semester}: ${it.subject} ${it.catalogNbr}`));
+      .forEach(course => {
+        let subjectSet = semesterMap[course.subject];
+        if (subjectSet == null) {
```
hanywang2: would it be null or undefined?

requirement-generator/requirement-satisfiable-course-generator.ts
```diff
@@ -17,25 +24,50 @@ function courseCanSatisfyCollegeOrMajorRequirement(
```
hanywang2: Once the function is better flushed, we would want documentation for maintainability

requirement-generator/requirement-satisfiable-course-generator.ts
```diff
@@ -17,25 +24,50 @@ function courseCanSatisfyCollegeOrMajorRequirement(
 
 const produceSatisfiableCourseIdForCollegeOrMajorRequirement = (
   collegeRequirement: CollegeOrMajorRequirement
-): readonly string[] => {
-  const set = new Set<string>();
+): EligibleCourses => {
+  const allMap: { [semester: string]: { [subject: string]: string[] } } = {};
```
hanywang2: Can this variable be named a little bit better? I wasn't sure what this is or why it is returned at first glance. `allElidgibleCoursesSemestersMap` would be much better

requirement-generator/requirement-satisfiable-course-generator.ts
SamChou19815: `bla == null` is equivalent to `bla === null || bla === undefined`

PR #80, opened 2020-03-10
Compute requirements using TypeScript

SamChou19815:
### Summary

This diff starts the real work of making requirements computation more maintainable and extensible in the future.

#### Rationale for the change

For background, the current requirement computation setup has several flaws:

- It's intimidating to look at and work in a 2000+ line json
- Some of the field's meaning is not clear unless you read the code elsewhere
- Enforce type checking on the json is hard because it can have so many variants
- Over the time, there might be junk left in the json
- Adding support for complex queries and checks usually requires a complete refactoring of the entire json, rather than only affecting a few requirements that need special care.

In my opinion, to create a truly extensible system for requirements checking, you either:

1. build a domain specific programming language for it
2. directly use TypeScript/JavaScript to compute requirements

Approach 1 is only necessary if you somehow want ordinary users to contribute to requirement from the website UI, not via pull requests. At the moment, I don't think the course plan team plan to do anything like that. Therefore, approach 2 is more reasonable and practical.

With approach 2, some of the difficult problem with json goes away automatically:

- You can break down all the checking code into different modules: no more 2000+ line files!
- You directly write code for computing requirement. There is no more separation of configuration and logic
- You get good type checking for free when you are directly using TypeScript
- Junks can be discovered by linter and TypeScript. We can easily spot them and purge them.
- Adding support for complex-query is very easy: you just write slightly more code for *one* requirement. TypeScript is a Turing complete language, so you can do anything you like for a requirement checking.

The only down side for this approach is that an outsider can no longer add a bunch of stuff in json and a new major support will appear. He/she will need to do the following:

1. Learn TypeScript
2. Compile new TypeScript code by `npm run build:req-gen`
3. Generate new json by `npm run req-gen`
4. Profit

However, this overhead is way smaller than maintaining the giant json.

#### Implementation

I manually rewrite all the requirement checking logic in TypeScript inside the `requirement-generator/checker` folder. All checkers functions are then assembled together into a object like

```typescript
{
  req1Checker: (course: Course): boolean => { /* ... */ },
  req2Checker: (course: Course): boolean => { /* ... */ },
  req3Checker: (course: Course): boolean => { /* ... */ },
  // ...
}
```

Then I added the names of the checker to the requirement json. You can see the diff that every requirement now has a new field like `"checkerName": "csCore"`. During requirement checking time, the checker reads the `checkerName` field, find the checker function in the object mentioned above, and run the checker to see whether course `c` satisfies requirement `r`.

When this refactoring is done, the `requirement-generator/requirement-checker.ts` is no longer needed, because we no longer interpret the json to check satisfiability of requirements.

#### Future Plan

The only goal of this diff is refactoring. To get absolute confidence that the refactoring doesn't break anything, I try to ensure that the generated json has **zero change**. Since all requirement checking on the frontend depends this generated json, no change in that json means no change in frontend behavior.

That being said, I did discover some potential bugs with the current requirement checking configuration. For example, some requirements' search seem to refer to some nonexisting fields, and some of the fields in the requirement json appears to be never used. They will be fixed in future diffs.

In future diffs, I also plan to remove all the `search`, `include`, `exclude` fields in `reqs.json` since their jobs are completely replaced by the TypeScript code newly written in this diff. Again, they are not removed because it will cause the script to generate a different json.

### Test Plan 

I can assert this diff breaks nothing, because the script generates **exactly the same** `decorated-requirements.json`, comparing against the last commit.


requirement-generator/checkers/colleges/checkers-college-ag.ts
```diff
@@ -0,0 +1,122 @@
...
+    'PLSCI 1300',
+    'PLSCI 1420',
+    'PLSCI 4190',
+    'PLSCS 1120',
+    'STS 2871',
+    'VIEN 2204'
+  ]
+);
+
+const calsChemistryOrPhysics = (course: Course): boolean => ['CHEM', 'CHEME', 'PHYS'].includes(course.subject);
```
SamChou19815: @hanywang2 This is the part I think the original behavior is wrong.
You probably only want 'CHEM' but 'CHEME' is also included, because the original code does something like `subject.includes(...)` rather than `subject === ...` To preserve this behavior, I also added 'CHEME' here.

requirement-generator/checkers/colleges/checkers-college-ag.ts
```diff
@@ -0,0 +1,122 @@
...
+  '(D-',
+  '(FL-',
+  '(HA-',
+  '(KCM-',
+  '(LA-',
+  '(SBA-'
+  // REQ_TODO: has uniqueIncludes
+  // REQ_TODO: the original reqs.json is probably wrong!
+  // @ts-ignore
+].some(code => course.code && course.code.includes(code));
```
SamChou19815: Not sure what's the original intention

requirement-generator/checkers/majors/cs.ts
```diff
@@ -0,0 +1,33 @@
+import { Course } from '../../types';
+import { courseMatchesCode, courseMatchesCodeOptions } from '../checkers-common';
+
+const csIntroProgramming = (course: Course): boolean => (
+  courseMatchesCodeOptions(course, ['CS 1110', 'CS 1112', 'CS 1114', 'CS 1115'])
+  || courseMatchesCodeOptions(course, ['CS 2110', 'CS 2112'])
```
hanywang2: What's the plan to not allow for cases of cross-over requirements like CS 1110 and CS 1112 satisfying the entire csIntroProgramming requirement?

requirement-generator/checkers/checkers-common.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param code : code to check courseName (can contain * to denote any value)
+ * @returns if a code matches the course name (CS 2110 and CS 2*** returns true, AEM 3110 and AEM 32** returns false)
+ */
+export function ifCodeMatch(courseName: string, code: string): boolean {
+  for (let i = 0; i < courseName.length; i += 1) {
+    if (code[i] !== '*' && courseName[i] !== code[i]) return false;
+  }
+  return true;
+}
+
```
hanywang2: Because these are super common methods, I think there should be specifications

requirement-generator/checkers/colleges/ag.ts
```diff
@@ -0,0 +1,122 @@
+import { Course } from '../../types';
+import { courseMatchesCodeOptions } from '../checkers-common';
+
```
hanywang2: Not gonna review every single requirement for accuracy as I believe they follow the same logic from reqs.json. Will iron out once the process is set up.

requirement-generator/checkers/index.ts
```diff
@@ -0,0 +1,44 @@
+import { Checkers } from './types';
```
hanywang2: Can we rename this file to `requirementsList.ts` or whatever is more suitable? Reason being is that we want to make the adding more requirements data process as clear and seamless as possible

requirement-generator/types.ts
```diff
@@ -20,17 +18,13 @@ export interface BaseRequirement {
   readonly name: string;
   readonly description: string;
   readonly source: string;
-  readonly search?: readonly RequirementSearch[];
-  readonly includes?: readonly string[][];
-  readonly excludes?: readonly string[][];
```
hanywang2: Amazing approach to improving the requirements functionality. It's super clear to me now that having a JSON structure was the best way to approach requirements at all

requirement-generator/checkers/colleges/checkers-college-ag.ts
hanywang2: You are completely right. That was a mistake. Not sure where I read that 'CHEME' is also considered as chemistry. Removing 'CHEME' is the move here

requirement-generator/checkers/majors/cs.ts
SamChou19815: It's not achievable with the current setup. The current setup simply exposes the problem more clearly. However, I do keep the group structure so that we do not completely lose the original json structure.

requirement-generator/checkers/colleges/checkers-college-ag.ts
hanywang2: In course data. Distribution requirements are represented as "(CA-AG)" or "(CA-AS)" or "(D-EN)" or "(D-AG)". In this case, the rationale was to use the includes feature (scans for substrings) to look for those distribution requirements without considering the suffix (or in this the college the distribution is affiliated to).

requirement-generator/checkers/majors/cs.ts
hanywang2: That makes sense

PR #81, opened 2020-03-10
Misc requirements fix

SamChou19815:
### Summary

- (Optimization) purge `include`, `exclude`, `search` fields that are no longer used.
- (Optimization) self check courses will now have `{}` as its `courses` field.
- (Optimization) if a requirement has no courses in semester `s` that satisfies it, then `s` will not appear in the `courses` field in that requirement
- (Bug Fix) Fill in correct BE focus area elective classes. See https://beadvised.bee.cornell.edu/full-list-of-focus-area-courses/
- (Cleanup) With the bug fix above, no non-university requirements have `all-eligible` courses. We can remove the special case to improve type readability.

### Test Plan

JSON generation script can still run. Generated json can work with the rest of the system.


requirement-generator/checkers/majors/be.ts
```diff
@@ -29,7 +29,17 @@ const bioEngineeringEngineeringStatistics = (course: Course): boolean => (
   || courseMatchesCodeOptions(course, ['CEE 3040', 'ENGRD 2700'])
 );
 
-const bioEngineeringFocusAreaElective = 'all-eligible' as const;
+const bioEngineeringFocusAreaElective = (course: Course): boolean => courseMatchesCodeOptions(
```
hanywang2: wowo that's a lot of ctrl c + v

src/requirements/reqs.json
```diff
@@ -48,13 +48,6 @@
                     "name": "CALS Credits",
                     "description": "55 CALS credits are required for graduation. CALS credits include all courses from departments within CALS and courses offered in Applied Economics and Management, Biological Sciences, Biology & Society, Earth and Atmospheric Sciences, Information Science, Nutritional Science, and The Department of Statistics and Data Science.",
                     "source": "https://cals.cornell.edu/undergraduate-students/student-services/degree-requirements/graduation-requirements",
-                    "search": ["acadGroup"],
```
hanywang2: out of curiosity, did you write a script to delete all of search and includes attributes or did u do it manually?

SamChou19815: Manually. The json still has something that I don't fully understand, so I manually deleted everything useless that I know for sure.

PR #82, opened 2020-03-10
Support sub-requirements

SamChou19815:
#### Rationale

Without sub-requirements, the nested include check will interfere with each other. For example, if a requirement requires `(A || B) && (C || D)`, the old infrastructure will lose the `and` and essentially only checks `(A || B) || (C || D)`.

#### Implementation

I updated all the checkers to either be a single function `(course: Course) => boolean` or an array of `(course: Course) => boolean`. An array of checker function represents the case that an requirement has sub-requirements, and each checker in the array correspond to one sub-requirement. As a result, the generated json now contains an array of `EligibleCourses` objects, where each object in array represents all courses that can satisfy one sub-requirement.

On the client side, the requirement checker is updated to use an array of courses. It groups satisfiable courses according to sub-requirement satisfiability. Then for each sub-requirement, as long there is one satisfiable course, we declare it done.

### Test Plan

Tested the original failing case of ['CS 2800', 'CS 2802']. Now it works!

Screen Shot 2020-03-10 at 01 43 05


src/requirements/reqs-functions.ts
```diff
@@ -206,46 +207,57 @@ function iterateThroughCollegeOrMajorRequirements(
...
-        // add the course to the list of courses used to fulfill that one requirement
-        coursesThatFulilledRequirement.push(code);
-
+      coursesThatFulfilledSubRequirement.forEach(({ code }) => {
         // Add course to dictionary with name
         if (code in satisfiedRequirementMap) satisfiedRequirementMap[code].push(requirementName);
         else satisfiedRequirementMap[code] = [requirementName];
-      }
+        courseCodesThatFulfilledRequirement.add(code);
+      });
```
SamChou19815: @hanywang2 @willespencer 
For each sub-requirement, we can potentially put multiple courses into the requirement fulfillment map.
This is not a problem with my implementation: the old implementation also implicitly does this (After all, the old one can't even distinguish between sub-requirements 😞).
I'm not sure how the `satisfiedRequirementMap` is used elsewhere, so I will leave the question to you.

hanywang2: Can you give an example? I'm having trouble understanding

SamChou19815: Consider the CS core requirement:
```json
[
  ["CS 2800", "CS 2802"],
  ["CS 3110"],
  ["CS 3410", "CS 3420"],
  ["CS 4410"],
  ["CS 4820"]
]
```

Suppose I have `["CS 2800", "CS 2802"]`, then the `satisfiedRequirementMap`  will contain
```json
{
  "CS2800": ["CS Core"],
  "CS2802": ["CS Core"],
}
```

I'm not exactly sure about the purpose of `satisfiedRequirementMap`, so I'm not sure whether the setup will cause problems.

hanywang2: The `satisfiedRequirementMap` is used for the information icon on the course card. It's not on display right now, but there should be a small information icon that when you hover over it on the course, it will display the requirement that it satisfies. Most ideally, we would want it to display like this if you have `["CS 2800", "CS 2802", "CS 3110", "CS 3410", "CS 3420"]`
```
{
   "CS 2800": [{type: "info": requirement: "CS Core"}]
   "CS 2802": [{type: "duplicate": requirement: "CS Core"}]
   "CS 3110": [{type: "info": requirement: "CS Core"}]
   "CS 3410": [{type: "info": requirement: "CS Core"}]
   "CS 3420": [{type: "duplicate": requirement: "CS Core"}]
}
```
...or something along these lines. We haven't implemented the icons yet so they could be subject to change, but anything that is informative to the user on which requirement all courses satisfy (and ideally sub-requirements tho the previous implementation doesn't address that)

hanywang2: We would want an implementation for `satisfiedRequirementMap` for this branch because we will be needing it very soon

PR #84, opened 2020-03-11
Strictify Requirements.vue

SamChou19815:
### Summary

To make further progress on making complex requirement map post-processing, it's important to strictify things a little bit so that we know the static types of those giant maps being passing around and manipulated. Therefore, I decide to convert `Requirements.vue` to be in TypeScript.

The conversion is mostly straightforward. If you read the diff, you will see that all I have added is type annotation. The only exception is that I changed `req.required` to `req.required || 0` to make TS happy, unless it will yell at us that `req.required` can be undefined.

### Test Plan

Played with the app. No behavior change.

### Notes

Waiting on #83 to be merged before merging this.

### DIFF
--- a/src/components/Requirements.vue
+++ b/src/components/Requirements.vue
@@ -141,13 +141,16 @@
   </div>
 </template>
 
-<script>
+<script lang="ts">
 import { Vue } from 'vue-property-decorator';
+// @ts-ignore
 import VueCollapse from 'vue2-collapse';
-import Course from '@/components/Course';
-import Modal from '@/components/Modals/Modal';
-/** @typedef { import('../requirements/types').StrictFulfilledByType } StrictFulfilledByType */
-/** @typedef { import('../requirements/types').BaseRequirement<StrictFulfilledByType> } Requirement */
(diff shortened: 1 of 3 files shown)


src/components/Requirements.vue
```diff
@@ -264,25 +274,17 @@ export default {
         // }
       ],
       requirementsMap: {
-        // CS 1110: 'MQR-AS'
+        // CS 1110: ['MQR-AS']
```
hanywang2: great catch

src/requirements/reqs-functions.ts
```diff
@@ -38,11 +32,11 @@ function createRequirementJSON(
     required: requirement.minCount,
     description: requirement.description,
     source: requirement.source,
-    fulfilled: null,
+    fulfilled: undefined,
```
hanywang2: What's the reason we are setting fulfilled as undefined rather than null?

SamChou19815: So we have one less case to worry about. Usually the convention is to either completely ban `undefined` or completely ban `null`

PR #87, opened 2020-03-12
Migrate more files to TypeScript and remove broken user data fetch

SamChou19815:
### Summary

Trying to migrate some easy files to TypeScript. The goal is to increase some type coverage without increasing merge conflicts for other devs, so it's focused on files that have not been touched recently.

During the migration, I discovered the source of `missing or insufficient firebase permission error`. In `src/store.ts`, we are fetching a collection (`user`) that no longer exists. To prevent future confusion, I removed the code as well as reference to the user collection.

### Test Plan

Nothing breaks.


src/vueDragulaConfig.ts
```diff
@@ -1,6 +1,7 @@
 import Vue from 'vue';
 import 'dragula/dist/dragula.css';
 
+// @ts-ignore
 import { Vue2Dragula } from 'vue2-dragula';
```
SamChou19815: This dependency does not have typescript type definition 😢

hanywang2: That is a big oof

src/store.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  actions: {
+    fetchUserProfile({ commit, state }) {}
+  },
+  mutations: {
+    setCurrentUser(state: State, val: User) {
+      state.currentUser = val;
+    }
+  }
+});
```
hanywang2: What's the purpose of using Vuex.Store in this case?

src/components/Dashboard.vue
```diff
@@ -50,8 +50,8 @@ import BottomBar from '@/components/BottomBar';
 import NavBar from '@/components/NavBar';
 import Onboarding from '@/components/Modals/Onboarding';
 
-
-import '@/vueDragulaConfig';
+import '@/vueDragulaConfig.ts';
```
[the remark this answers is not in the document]
SamChou19815: Good question, now I updated the linter config to disallow .ts in imports, to be consistent with the rest of the code.

src/components/Login.vue
```diff
@@ -134,15 +134,23 @@
...
 </template>
 
-<script>
-import firebase from 'firebase/app';
+<script lang="ts">
+import Vue from 'vue';
+import firebase, { User } from 'firebase/app';
 
-const fb = require('../firebaseConfig.js');
+import * as fb from '@/firebaseConfig.ts';
```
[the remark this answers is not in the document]
SamChou19815: Now we must remove `.js` because it's no longer a js file.

src/store.ts
hanywang2: @SamChou19815 do you mind adding some comments on store.ts to explain what it does? I don't have much experience with it and I think it could help the devs as well

SamChou19815: See the deleted `src/store.js`. The original file also uses Vuex.Store, and I'm not sure why.

src/components/Login.vue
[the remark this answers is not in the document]
hanywang2: O I mean the difference between `import * as fb` and `const fb =`

src/store.ts
SamChou19815: I guess it's like redux in the React world. However, I think currently it's being misused. I think using it storing all the courses taken would be a better idea for using the store.

hanywang2: Interesting... I'll make a note to ask Will and the devs about it

src/components/Login.vue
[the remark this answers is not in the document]
SamChou19815: Usually on the web people use the `import` syntax. The `require` is only necessary on node, because node.js invented a module system before the ESModule (which uses `import`) becomes standard.
You might wonder for those typescript files, even if some of them runs on node, still use `import`  syntax. That's because typescript is a compiler, and it can transform it to use `require` after compilation.
In general, use `import` if you can.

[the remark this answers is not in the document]
hanywang2: Gotcha. Thanks for clearing that up for me

PR #88, opened 2020-03-12
Refactor types of requirement fulfillment

SamChou19815:
### Summary

It turns out that the previous few refactors are still not enough to implement double-counting detections and other potential post-processing functionality in a clean way.

This diff intends to address one of a few remaining problems. Currently, `RequirementFulfillment` object still has a bunch of field that's frontend related. The computation of these fields should not be entangled together with requirement satisfiability code.

This diff purges all uses of these bad practices for `MutableRequirementFulfillment`. Notable changes include:

- Doing capitalization of the first letter on the `SingleMenuRequirement` side instead of `RequirementFulfillment` side
- Refactor the `displayDescription` field into another object type, so that `RequirementFulfillment` no longer contains any UI-only fields
- After that `RequirementFulfillment` can be made immutable! Thus, it's now immutable

### Test Plan

Everything still works.

### DIFF
--- a/src/requirements/types.ts
+++ b/src/requirements/types.ts
@@ -80,28 +80,32 @@ export type DecoratedRequirementsJson = {
   readonly major: MajorRequirements<DecoratedCollegeOrMajorRequirement>;
 };
 
-export type MutableRequirementFulfillment = {
-  name: string;
-  type: string;
-  courses: readonly string[];
-  required?: number;
-  description: string;
-  source: string;
-  fulfilled?: number;
-  progressBar: boolean;
-  displayDescription: boolean;
+export type RequirementFulfillment = {
+  /** The original requirement object. */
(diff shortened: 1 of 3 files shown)


src/requirements/types.ts
```diff
@@ -80,28 +80,32 @@ export type DecoratedRequirementsJson = {
...
+  readonly requirement: BaseRequirement;
+  /** A list of course codes that satisfy this requirement. */
+  readonly courses: readonly string[];
+  /**
+   * Current fulfillment progress.
+   * When it's a number, it's either number of courses or number of credits.
+   * When it's undefined, it means that the requirement is self-check.
+   */
+  readonly fulfilled?: number;
 };
```
hanywang2: Great idea making Requirementsfulfillment immutable

PR #90, opened 2020-03-12
Break requirement computation into small phases

SamChou19815:
### Summary

Now we can finally make our first concrete push towards requirement computation post-processing. This diff setups the infra for post-processing via monads (:octocat::bangbang:).

I first refactor the type of `RequirementFulfillment` to be parametric over metadata M. In other words, each requirement fulfillment can have base data `{ requirement, courses }` and potentially other metadata `M`. The other metadata `M` can be:

- nothing: `{ }`
- fulfillment statistics: `{ fulfillment: 42 }`
- some alert information like `{ alert: ['satisfies by two courses', 'has requirement that might be double counted', ...] }`
- any combo of above

Then I created `postProcessRequirementsFulfillments` that computes over metadata. You can see the JSDoc about it's usage. (This pattern is known as monad in functional programming languages.)

The above you setup finally allows us to break `iterateThroughCollegeOrMajorRequirements` into three clearly defined phases:

1. Compute all requirement fulfillment locally for each requirement.
2. Compute fulfillment statistics for each requirement.
3. Build requirement map

This clear separation of phases makes some problems very easy to spot. Namely, the code of building the requirement map doesn't really need to be inside `iterateThroughCollegeOrMajorRequirements`. However, refactoring it elsewhere will be the focus of future diff to keep this one small.

The rest of the diff is mostly moving code around.

### Test Plan

Everything still works.

### DIFF
--- a/src/requirements/reqs-functions.ts
+++ b/src/requirements/reqs-functions.ts
@@ -4,42 +4,13 @@ import {
   BaseRequirement,
   DecoratedCollegeOrMajorRequirement,
   RequirementFulfillment,
+  RequirementFulfillmentStatistics,
   GroupedRequirementFulfillmentReport
 } from './types';
 
 type RequirementMap = { readonly [code: string]: readonly string[] };
 type MutableRequirementMap = { [code: string]: readonly string[] };
 type MutableRequirementMapWithMutableChildren = { [code: string]: string[] };
-
-/**
- * @param requirement : the requirement information as object
- * @param totalRequirementCredits : total credits of courses that satisfied requirement
- * @param totalRequirementCount : total number of courses that satisfied requirement
- * @param coursesThatFulilledRequirement : courses that satisfied requirement
- * @returns {RequirementFulfillment}
- */
-function createRequirementJSON(
-  requirement: BaseRequirement,
-  totalRequirementCredits: number,
-  totalRequirementCount: number,
-  coursesThatFulilledRequirement: readonly CourseTaken[][]
-): RequirementFulfillment {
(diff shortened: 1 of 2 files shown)


src/requirements/reqs-functions.ts
```diff
@@ -4,43 +4,14 @@ import {
...
-      break;
-    case 'self-check':
-      fulfilled = undefined;
-      break;
-    default:
-      throw new Error('Fulfillment type unknown.');
-  }
-  return { requirement, courses: coursesThatFulilledRequirement, fulfilled };
-}
-
```
hanywang2: Definitely makes the code much cleaner. Requirements JSON like this was definitely not necessary in the equation

src/requirements/reqs-functions.ts
```diff
@@ -128,35 +99,128 @@ function iterateThroughUniversityRequirements(
...
   mergeRequirementsMap(requirementsMap, satisfiedRequirementMap);
 
   return requirementJSONs;
 }
 
+/**
+ * @param coursesTaken a list of all taken courses.
+ * @param requirement the requirement to compute course fulfillment.
+ * @returns a naively computed list of courses that fulfill the requirement, partitioned into sub-requirement filfillment.
+ */
```
hanywang2: Thank you for the documentation

SamChou19815:
@hanywang2 The `fulfilled` is already used here: https://github.com/cornell-dti/course-plan/blob/42c7147222d3d2a867622ef6d77ae7db7bcc80bb/src/components/Requirements.vue#L14-L24

PR #91, opened 2020-03-12
Decouple requirement filfillment map and group computation

SamChou19815:
### Summary

With the infra setup in #90, we can clearly see how easy it is to decouple requirement fulfillment computation and `requirementMap` computation. Now instead of letting them entangle each other in `reqs-functions.ts`, we should make map computation depends on fulfillment list computation.

I moved all map computation code into `computeRequirementMap`, which takes in an already computed grouped requirement fulfillment. A lot of code is simplified:

- It turns out that `mergeRequirementsMap` is completely unnecessary when we build the map globally, so it's purged.
- `iterateThroughXxxRequirements` now become pure functions, so I renamed it to be `computeXxxRequirementFulfillments` to reflect this fact.

The biggest achievement of this diff is that now all functions in `reqs-functions.ts` are pure.

### Test Plan

Everything still works.

### DIFF
--- a/src/requirements/reqs-functions.ts
+++ b/src/requirements/reqs-functions.ts
@@ -1,7 +1,6 @@
 import requirementJson from './typed-requirement-json';
 import {
   CourseTaken,
-  BaseRequirement,
   DecoratedCollegeOrMajorRequirement,
   RequirementFulfillment,
   RequirementFulfillmentStatistics,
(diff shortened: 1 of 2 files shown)


src/requirements/reqs-functions.ts
```diff
@@ -94,32 +73,27 @@ function iterateThroughUniversityRequirements(
     minCount: 0,
     applies: 'all'
   } as const;
```
hanywang2: What does `as const` do in this case?

SamChou19815: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions
It's necessary to make TypeScript type `fulfilledBy` as an enum rather than a regular string.

PR #97, opened 2020-04-02
Simple Mobile Version

tcho6319:
### Summary 
This pull request is the simple mobile version of the CoursePlan web app. All features are still retained in the mobile version, though some may be styled a bit differently to better accommodate vertical scrolling. The biggest difference between mobile and web is that the requirements menu is hidden and can be toggled in the mobile version. Five breakpoints are used for screen sizes: 1347px, 976px, 878px, 600px, and 440px.

General features/changes at each breakpoint:
- 1347px: Bottom Bar Tab max change from max 4 tabs to max 2.
- 976px: Change width of requirements bar and bottom bar. Bottom bar course content stacked.
- 878px: isTablet. Hide requirements bar by default. Requirement bar is toggleable.
- 600px: Adjusting for smaller mobile size, see more tab width, modal widths
- 440px: isMobile. Changing width of bottom bar tabs (2 lines), decrease bottom bar height

- [x] Made widths of bottom bar, modals, and onboarding responsive.
- [x] Requirements bar is toggleable in mobile ver using the info icon as a placeholder and takes up the entire screen.
- [x] Nav bar in mobile view is horizontal and fixed on top of screen.
- [x] Re-consider which features, if any, should be removed from mobile ver.
- [x] Re-consider icon to use for toggling requirements menu.

### Test Plan 
Responsive design was checked by physically resizing window and by using responsive/iphone widths in Developer Tools. More testing should be done to ensure proper display of the app on actual phones.

### Notes
- [] Further testing on different screen sizes.


hanywang2:
Change recommendations introduced on Slack

src/components/Dashboard.vue
```diff
@@ -91,12 +92,20 @@ export default {
       bottomBar: { isPreview: false, isExpanded: false },
       requirementsKey: 0,
       isOnboarding: false,
-      isEditingProfile: false
+      isEditingProfile: false,
+      isOpeningRequirements: false,
+      isMobile: window.innerWidth <= 1347
```
willespencer, line 97: Didn't know you could do this, smart!

hanywang2: @tcho6319 is big brains

src/components/BottomBarCourse.vue
```diff
@@ -98,18 +98,130 @@
             <p>{{ courseObj.description }}</p>
           </div>
         </div>
+        <div class="bottombarcourse-bar-details-noOverflow" v-if="isSmallerWidth">
```
willespencer, line 101: Definitely a way to do this without duplicating all of the code. Want to get this merged in though, so could you create a Monday.com item and assign you and me to it, and one of us can cover it later?

src/components/BottomBarCourse.vue
```diff
@@ -303,4 +418,72 @@ export default {
...
+      padding: 0.5rem;
+    }
+
+    &-bar {
+      &-info{
+        &-noOverflow {
+          width: 100%;
+          height: auto;
+          float: none;
+          overflow: none !important;
```
willespencer, line 440: !important is usually bad practice. There is some good information here if you want to read some more about why and other options: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity . See if you can fix it here and everywhere else it is in this code in a few minutes, and if you can't, make sure it does not break anything and make an item 😄

src/components/BottomBarTitle.vue
```diff
@@ -38,18 +38,37 @@ export default {
...
 
-@media only screen and (max-width: 1347px) {
+@media only screen and (max-width: 976px) {
+  .bottombartitle {
+    left: 25.5rem;
+    width: calc(100vw - 25.5rem);
+  }
+}
+
+@media only screen and (max-width: 878px) {
```
willespencer, line 61: Should this be 976 px like the other ones?

src/components/BottomBarTabView.vue
```diff
@@ -234,7 +235,7 @@ export default {
     }
   }
 }
-@media only screen and (max-width: 878px) {
+@media only screen and (max-width: 600px) {
```
willespencer, line 238: I'm seeing 976 and 440 breakpoints elsewhere. Is having a third one necessary?

src/components/Dashboard.vue
```diff
@@ -135,8 +139,12 @@ export default {
         });
     },
 
-    isMobileEventHandler(e) {
-      this.isMobile = window.innerWidth <= 1347;
+    resizeEventHandler(e) {
+      this.isMobile = window.innerWidth <= 440;
+      this.isTablet = window.innerWidth <= 878;
```
willespencer, line 144: Should this be the 976 number?

src/components/Dashboard.vue
```diff
@@ -157,6 +165,13 @@ export default {
       return semesters;
     },
 
+    updateSemesterView() {
+      if (this.isMobile) {
+        // Make sure semesterView is not compact by default on mobile
```
willespencer, line 170: Nice 👍

src/components/Dashboard.vue
```diff
@@ -568,7 +606,7 @@ export default {
 }
 
 
-@media only screen and (max-width: 1347px) {
+@media only screen and (max-width: 878px) {
```
willespencer, line 609: Same comment as above

src/components/Modals/Modal.vue
```diff
@@ -231,7 +231,7 @@ export default {
   width: 15.5rem;
 }
 
-@media only screen and (max-width: 878px) {
+@media only screen and (max-width: 600px) {
```
willespencer, line 234: Same here.

src/components/BottomBarCourse.vue
tcho6319: Monday item made!

src/components/BottomBarTitle.vue
tcho6319: Here it should be 878px because the bottom bar title width changes drastically from the 976px breakpoint.

At the 979px breakpoint: To ensure that we can have the requirements bar show for as small of a screenwidth as possible, the requirements bar width is decreased and so the bottom bar title has to move left by the decreased width amount. 

At the 878px breakpoint: The requirements bar is hidden by default and becomes toggleable so the bottom bar title width has to be 100%.

src/components/BottomBarTabView.vue
tcho6319: At the 600px breakpoint: Adjusting for smaller mobile size, see more tab width. 

I updated the PR description up top to include the reasoning for additional breakpoints if you want me to expand more on them :)

src/components/Dashboard.vue
tcho6319: Here we would want it to be 878 and not 976 because at the 878px breakpoint, the requirements bar would be hidden by default and toggleable which is what this.isTablet controls

src/components/Dashboard.vue
tcho6319: Similar reasoning as to why this.isTablet = window.innerwidth <= 878 and not 976 from up above

src/components/Modals/Modal.vue
tcho6319: At 600px, we're starting to control for more mobile sized screens so we want the modal to have width 100% in that case.

src/components/BottomBarCourse.vue
tcho6319: Removed !important from the file!

hanywang2:
@tcho6319 realize that if u merge it this will be the first time in I believe a century that there are less than 3 PRs for Will to review. Now how can we be ok with that?

willespencer:
I just want to mention that all of the other PRs have *already* been reviewed. I'm feeling pretty good rn 😄

PR #100, opened 2020-04-07
Update requirement documentation

SamChou19815:
### Summary 

Update requirements docs to match the new setup.

### Test Plan 

Reviewer could  👀  https://github.com/cornell-dti/course-plan/blob/requirement-docs/src/requirements/README.md directly.

### Notes 

Every dev who plans to work on requirements should take a look at the doc and point out places that you are confused.

### DIFF
--- a/src/requirements/README.md
+++ b/src/requirements/README.md
@@ -1,44 +1,126 @@
 # Requirements Data
 
-The requirements data and its interpreting algorithm is uniquely developed by and for CoursePlan. The data for requirements data can be accessed at `requirements/reqs.json`. The structure is designed to be able to accurately express almost all requirements at Cornell while making it as easy as possible to read and maintain. Here's an example of the data for the College of Engineering.
+The requirements data and its interpreting algorithm is uniquely developed by and for CoursePlan.
+This document aims to give you a high level overview of how it works.
 
-```json
-"EN": {
-	"name": "Engineering",
-	"requirements": [
-		{
-			"name": "Mathematics",
-			"description": "MATH 1910, 1920, 2930 or 2940, and a mathematics course chosen by the Major.",
-			"source": "https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements",
-			"search": ["code"],
-			"includes": [["MATH 1910"], ["MATH 1920"], ["MATH 2930", "MATH 2940"]],
-			"excludes": [["CS 1810"]],
-			"fulfilledBy": "credits",
-			"minCount": 14
-		},
-		{
-			"name": "Freshman Writing Seminars",
-			"description": "All students are required to take two first-year writing seminars",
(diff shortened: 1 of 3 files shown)


src/requirements/README.md
```diff
@@ -1,44 +1,103 @@
...
+
+Requirement computation is split into two phases for efficiency concerns:
+
+1. Pre-computation phase.
+   In this phase, for each requirement (except a few that almost all classes that fulfill it), we
+   enumerate all courses that can fulfill it. We store the enumeration into a "decorated" requirement
+   json.
+2. Client-side computation phase.
+   In this phase, the user's course list is compared against requirement data.
+   For requirements that can be satisfied by almost any course, we run a simple regex check to see
```
willespencer, line 16: Can you explain (maybe with an example?) what these regexs are like in the "University Requirements" section? I get why these reqs are computed differently than the other requirements from this doc, but not specifically how they are computed.

src/requirements/README.md
```diff
@@ -1,44 +1,103 @@
...
+   should not be. (NOTE: THIS HAS NOT BEEN IMPLEMENTED YET.)
+3. Finally, We sum up the total number of sub-requirements/credits already fulfilled for each
+   requirement, to decide whether a requirement as a whole as been satisfied.
+
+## Contributing
+
+If you want to change a requirement checker, edit relevant files in `checkers` folder.
+
+If you want to add a support for new major or college, add relevant files in `checkers` folder.
+Then, import and link your new checkers in
```
willespencer, line 99: Could you add a little bit more information here, specifically on what "import and link your new checkers" entails? Since we might have some people less experienced with requirements adding data, it would be useful to expand on this.

src/requirements/README.md
```diff
@@ -1,44 +1,103 @@
...
+pre-computed to reduce the data sent from server to client.
+
+It's computed on the client side directly in `computeUniversityRequirementFulfillments` function
+inside [`reqs-functions.ts`](./reqs-functions.ts).
+
+#### College or Major Requirements
+
+1. We first naively compute whether a sub-requirement is fulfilled by checking whether any course in
+   user's list is in the pre-computed satisfying course list.
+2. Then, we run a deduplication pass to remove all courses that have been double-counted that
```
einc, line 89: Is this where we would check for cross-listed classes fulfilling a single requirement? (like ENGRD 2110 and CS 2110)? Or should all versions of the same class be added in the checker?

SamChou19815: Currently there is no way. We can add another pass to the pre-computation phase that adds in all cross-listed course. We also need to add another deduplication phase that merges cross listed courses together on client side.

src/requirements/README.md
SamChou19815: Added link to the relevant code

src/requirements/README.md
SamChou19815: Added examples

PR #104, opened 2020-04-19
Provide helper higher order functions to make writing requirements more intuitive

SamChou19815:
### Summary

During course plan meeting, someone expressed concern that the new requirement setup is too hard to understand. This is a valid concern, since the new setup uses a lot of idea of functional programming and might not be friendly to beginners.

Reverting to the old setup is not an option since we don't want to lose the customization power. Therefore, we need to strike a balance between flexibility and readability.

The solution is actually not too hard: we can provide some higher order functions that abstract away the functional programming detail, to make you feel like you are writing the old nested array `includes`. Beginners can easily follow some examples without the need to really understand wha't happening under the hood.

### Test Plan

I use the new helper functions to rewrite almost all requirements. It should be much easier to read now.

You can run `npm run req-gen` to see this change doesn't change the requirement computation result.


src/requirements/checkers/majors/be.ts
```diff
@@ -1,44 +1,40 @@
...
-];
+const bioEngineeringBioChemistry = includesWithSubRequirements(
+  ['BIOMG 3300', 'BIOMG 3330', 'BIOMG 3350', 'BIOMG 3310'],
+  ['BIOMG 3320']
+);
 
 const bioEngineeringEngineeringStatistics = (course: Course): boolean => (
   courseMatchesCodeOptions(course, ['BEE 2220', 'ENGRD 2210', 'CHEME 3130', 'MSE 3030'])
   || courseMatchesCodeOptions(course, ['CEE 3040', 'ENGRD 2700'])
 );
```
SamChou19815, line 28: This requirement computation is wrong. The fix is delayed to the next diff to ensure zero change in generated requirement json.

SamChou19815: Fixed in #105

src/requirements/checkers/checkers-common.ts
```diff
@@ -43,3 +43,13 @@ export const courseIsFWS = (course: Course): boolean => (
...
+
+export const includesWithSingleRequirement = (...includes: readonly string[]) => (course: Course): boolean => (
+  courseMatchesCodeOptions(course, includes)
+);
+
+export const includesWithSubRequirements = (...includes: readonly string[][]): readonly ((course: Course) => boolean)[] => (
+  includes.map(
+    subRequirementInclude => (course: Course) => courseMatchesCodeOptions(course, subRequirementInclude)
+  )
+);
```
hanywang2, line 55: All functions in this file will be referenced to a lot. For the sake of clarity to the requirement adders, can we get specifications for these two new functions too?

src/requirements/checkers/majors/be.ts
hanywang2: approved!

PR #118, opened 2020-05-03
small requirement fixes from cp grill

einc:
### Summary 

This pull request are some small requirement fixes from cp grill
- [x] Fixed CALS Credits (55 required for graduation) 
- [x] Removed duplicate FL req for A&S
- [x] Added Chris's major :)
- [x] Added some non-Engineering minors (business, psych, cog sci)
- [x] Temporarily commented out showing the user's college when displaying minor because they aren't related and showing it is confusing

### Test Plan 

Tested CALS Credits by adding CALS school and other department classes.
Tested minors by adding the required classes. 
CALS Credit Fixed
BU


src/requirements/data/index.ts
```diff
@@ -130,30 +140,45 @@ const json: RequirementsJson = {
     }
   },
   minor: {
-    DBME: {
-      name: 'Dyson Business Minor for Engineers',
```
hanywang2, line 134: Are we sure that the business minor is the same for everyone outside of engineering?

src/components/Requirements.vue
```diff
@@ -21,7 +21,7 @@
           <div :style="{'border-bottom': minor.display ? `2px solid #${reqGroupColorMap[req.group][0]}` : ''}"
             @click="activateMinor(id)" class="major-title" v-for="(minor, id) in minors" :key="minor.id">
             <p :style="{'font-weight': minor.display ? '500' : '', 'color' : minor.display ? `#${reqGroupColorMap[req.group][0]}` : ''}"  class="minor-title-top">{{minor.minorFN}}</p>
-            <p :style="{'color': minor.display ? `#${reqGroupColorMap[req.group][0]}` : ''}" class="minor-title-bottom">({{user.collegeFN}})</p> <!-- Change for multiple colleges -->
+            <!-- <p :style="{'color': minor.display ? `#${reqGroupColorMap[req.group][0]}` : ''}" class="minor-title-bottom">({{user.collegeFN}})</p> Change for multiple colleges -->
```
hanywang2, line 24: Super nice. Tho are there minors that you can't take if you're in certain colleges?

src/requirements/data/index.ts
einc: I think there's three diff business minors (Business (which i also added), this one, and one for Life Sciences Majors) that are all different. So I just went with the name that the Johnson school gave each one

src/components/Requirements.vue
einc: Yea and according to certain majors too (but we don't have any restrictions yet). i just commented this out bc it was confusing to show the user's school bc it doesn't have any relation with the minor

willespencer: Making an item on Monday for us to add an excludes for minors based on majors 😄

src/requirements/data/colleges/ag.ts
```diff
@@ -8,7 +8,8 @@ const calsCreditsRequirement: CollegeOrMajorRequirement = {
     + 'Biological Sciences, Biology & Society, Earth and Atmospheric Sciences, Information Science, Nutritional Science, '
     + 'and The Department of Statistics and Data Science.',
   source: 'https://cals.cornell.edu/undergraduate-students/student-services/degree-requirements/graduation-requirements',
-  checker: (course: Course): boolean => ['AG', 'BU'].includes(course.acadGroup),
+  checker: (course: Course): boolean => ['AG'].includes(course.acadGroup)
+  || ['AEM', 'BIOEE', 'BIOMG', 'BIOMI', 'BIONB', 'BSOC', 'EAS', 'INFO', 'NS', 'STSCI'].includes(course.subject),
```
willespencer, line 12: Should we be doing this the same way as we do A&S now?

einc: i don't think so. for arts i think we just check any distribution that ends in a '-as'. but for this req, it says it has to be in CALS or the listed certain departments.
so like 'into to china' would fulfill the a&s req (because it's a -as class), but it shouldn't for this one.

einc: hm i tried doing it the same way using '-AG' but certain departments like info don't get counted, even tho info sci is offered in CALS. the catalogDistr for info sci is '-AS'

willespencer: Oops you're totally doing it right, my bad.

PR #120, opened 2020-05-07
Walkthrough

rmz38:
### Summary
This pull request is for the new walk through designed by Chris. It uses IntroJs to implement the tour. So far, it implements Chris' walk through that procs after onboarding is finished. If the dashboard initializes with a semester, this will work as intended. However, it does not, so the walk through stops at the requirements bar. 

- [x] implemented the initial walk through assuming one semester exists
- [x] figured out creating walk throughs when first using a feature using firebase
- [x] styled tooltips
- [ ] implemented the walk through when first using a feature
- [x] found a workaround for the issue of no semesters existing

### Test Plan 
I used the site before and after onboarding to test:
image
image
image


src/assets/scss/introjs.scss
```diff
@@ -0,0 +1,103 @@
...
+    border-radius: 9px;
+}
+.introjs-helperNumberLayer{
+    display: none;
+}
+.introjs-helperLayer{
+    border-radius: 9px;
+}
+.introjs-donebutton{
+    color: white !important;
```
willespencer, line 12: Let's talk about the !important's! There is usually a way to get around using them, especially by being more specific. There is some good information here if you want to read some more about why to avoid it and some other options: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity. I'm not sure how possible this is because you're editing css provided by the library, but it is worth an attempt! Let me know if you can't figure out how and I can see if it is possible.

src/assets/scss/introjs.scss
```diff
@@ -0,0 +1,103 @@
...
+    background-image: none !important;
+    background-color: #508197 !important;
+    text-shadow: none;
+    border: none;
+    border-radius: 3px;
+    float: right !important;
+    font-family: "ProximaNova";
```
willespencer, line 19: You should be able to apply this font-family to a parent css class so you don't have to apply it in all of the below classes as well

src/assets/scss/introjs.scss
```diff
@@ -0,0 +1,103 @@
...
+}
+.introjs-skipbutton:active{
+    background-image: none;
+    background-color:transparent;
+    font-family:"ProximaNova"; 
+    padding-left: 0px;
+    color: #508197;
+    background-color: white;
+}
+.introjs-skipbutton:focus{
```
willespencer, line 49: Since this :focus is identical to the :active, you can combine them! There are some great examples of the best way to do this in SCSS in Navbar.vue

src/assets/scss/introjs.scss
```diff
@@ -0,0 +1,103 @@
...
+    font-family:"ProximaNova"; 
+    padding-left: 0px;
+    color: #508197;
+    background-color: white;
+}
+.introjs-skipbutton:hover{
+    box-shadow:none;
+}
+.introjs-nextbutton{
+    // display: none;
```
willespencer, line 61: Remove commented out css

src/assets/scss/introjs.scss
```diff
@@ -0,0 +1,103 @@
...
+    color: white !important;
+    background-image: none;
+    background-color: #508197;
+    text-shadow: none;
+    border: none;
+    border-radius: 3px;
+    font-family: "ProximaNova"
+}
+.introjs-nextbutton:hover{
+    // display: none;
```
willespencer, line 71: Same here and the class below it.

src/components/Dashboard.vue
```diff
@@ -33,6 +35,24 @@
         @updateRequirementsMenu="updateRequirementsMenu"
       />
     </div>
+    <tourwindow
+      :title="welcome"
+      :text="welcomeBodytext"
+      :exit="welcomeExit"
+      :buttonText="welcomeButtonText"
+      v-on:hide = "welcomeHidden = false; if(startTour==false) startTour = true"
```
willespencer, line 43: Couple things
1. There is a shorthand for 'v-on' - you can just use '@' (see https://vuejs.org/v2/guide/syntax.html#v-on-Shorthand)
2. Might be a bit nicer to move everything in this v-on to a function to be a bit more readable and easier to change later.

src/components/Dashboard.vue
```diff
@@ -100,7 +129,18 @@ export default {
...
+      maxBottomBarTabs: window.innerWidth <= 1347 ? 2 : 4,
+      welcome: 'Welcome Cornellian!',
+      welcomeBodytext: 'View your college requirements, plan your semesters and courses, and more.',
+      welcomeExit: 'No, I want to skip this',
+      welcomeButtonText: 'Start Tutorial',
+      welcomeHidden: false,
+      startTour: false,
+      showTourEndWindow: false,
+      congrats: 'Congratulations! That’s a wrap',
+      congratsBodytext: 'Other than this, there is more you can explore, so feel free to surf through CoursePlan 🏄',
```
willespencer, line 141: Instead of putting this emoji in the text, can you save it as an image and use that instead? Let me know if you need help.

src/components/Dashboard.vue
```diff
@@ -400,7 +439,10 @@ export default {
       // Get map of requirements
       this.buildRequirementsAlert(requirementsMap);
     },
-
+    showTourEnd() {
+      this.showTourEndWindow = true;
+      console.log('asdfasdf');
```
willespencer, line 444: Console log 😄

src/components/Modals/TourWindow.vue
```diff
@@ -0,0 +1,139 @@
...
+          <img src = "@/assets/images/Person_planning.svg" alt = "person planning">
+        </div>
+        <div class = "content">
+          <div class = "title">
+            {{ title }}
+          </div>
+          <div class = "body">
+            {{ text }}
+          </div>
+          <button v-on:click = "$emit('hide'); $emit('startTour')">
```
willespencer, line 15: Consider using the shorthand here and elsewhere in the file. Also consider making a function to call (but can go either way)

src/components/Modals/TourWindow.vue
```diff
@@ -0,0 +1,139 @@
...
+    };
+  },
+  directives: {
+    'click-outside': clickOutside
+  },
+  mounted() {
+  },
+  methods: {
+    startWalkthrough() {
+      // add stuff for starting tour
```
willespencer, line 61: Add a "TODO" to the front of this comment so it is easier to find later, and make a Monday item in the walkthrough section so we do not forget.

src/components/Requirements.vue
```diff
@@ -1,6 +1,13 @@
 <template v-if="semesters">
   <div class="requirements">
-    <div class="fixed">
+    <div class="fixed"
+      data-intro='<b>This is your Requirements Bar 📋</b><br>
```
willespencer, line 4: Same thing with this emoji

src/components/Semester.vue
```diff
@@ -29,8 +29,19 @@
...
+    v-if="isFirstSem" class="semester-addSemesterButton" @click="openSemesterModal"
+    data-intro='<b>Add your past and future Semester Cards</b><br>
+      <div class = "introjs-bodytext">Once you’re done setting up your current semester,
+      feel free to add both past and future semesters. Try to utilize your requirements bar</div>'
+    data-step ='4'
+    data-disable-interaction = '1'
+    >+ New Semester</button>
+    <div class="semester-content"
+      data-step = '2'
+      data-intro = '<b>This is a Semester Card of your current semester! 🍂❄️🌸☀️</b><br><div class = "introjs-bodytext">
```
willespencer, line 42: These emojis already have images in the assets folder!

src/components/Semester.vue
```diff
@@ -108,6 +125,9 @@ const spring = require('../assets/images/springEmoji.svg');
 const winter = require('../assets/images/winterEmoji.svg');
 const summer = require('../assets/images/summerEmoji.svg');
 
+// testing walkthrough using init
```
willespencer, line 128: What is this for?

src/components/Login.vue
```diff
@@ -118,6 +118,8 @@
 <script lang="ts">
 import Vue from 'vue';
 import firebase, { User } from 'firebase/app';
+// @ts-ignore
+import introJs from 'intro.js';
```
SamChou19815, line 122: intro.js has typescript type definitions: https://www.npmjs.com/package/@types/intro.js
Install it and remove the @ts-ignore

src/assets/scss/introjs.scss
rmz38: Unfortunately other than using IDs it looks like using !important is the best way to override it.
image

src/components/Requirements.vue
rmz38: Already mentioned the issue on slack, but I'm not sure how to add the built filepath to the data-intro string

rmz38:
Still need to add:
- [ ] Showing walkthrough on desktop even if first used on mobile
- [ ] Replace clipboard emoji with svg
- [ ] Tooltips that show up the first time using a feature if design requires it

src/components/Modals/TourWindow.vue
```diff
@@ -0,0 +1,131 @@
...
+export default {
+  props: {
+    title: String,
+    text: String,
+    exit: String,
+    buttonText: String
+  },
+  data() {
+    return {
+      hideOnClick: ''
```
hanywang2, line 50: I'm not exactly understanding why hideOnClick is a string? Could this be variable named a little better or am I not understanding something here?

src/components/Requirements.vue
```diff
@@ -372,6 +399,13 @@ export default Vue.extend({
       });
       return (id < 2 || id === currentDisplay);
     },
+    showReqTooltip() {
+      /* Revisit this when implementing tooltips appearing for first time
+      if (this.req_tooltip_seen === false) {
+        tour.start();
+        // add update req tooltip seen
+      } */
+    },
```
hanywang2, line 409: Let's not comment out code but leave the function available. Either comment it out completely or most ideally remove the entire section

src/components/Semester.vue
```diff
@@ -180,10 +205,9 @@ export default {
       this.isShadow = false;
       this.isDraggedFrom = false;
     });
-
+    init = false;
```
hanywang2, line 208: What is happening here?

hanywang2 (changes requested):
Let's try to keep the PR sizes smaller in the future. We could work out the walkthrough in steps and just have some way to gate out the incompleted features. But great job overall!

src/assets/scss/introjs.scss
```diff
@@ -0,0 +1,85 @@
...
+}
+.introjs-donebutton{
+    color: white !important;
+    background-image: none !important;
+    background-color: #508197 !important;
+    text-shadow: none;
+    border: none;
+    border-radius: 3px;
+    float: right !important;
+    padding: .5em .7em .5em !important;
```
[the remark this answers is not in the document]
rmz38: yes :(
image

src/components/Dashboard.vue
```diff
@@ -66,6 +91,16 @@ Vue.component('requirements', Requirements);
...
+Vue.component('tourwindow', TourWindow);
+
+const tour = introJs();
+tour.setOption('exitOnEsc', 'false');
+tour.setOption('doneLabel', 'Finish');
+tour.setOption('skipLabel', 'Skip This Tutorial');
+tour.setOption('nextLabel', 'Next');
+tour.setOption('exitOnOverlayClick', 'false');
+
+const surfing = require('../assets/images/surfing.svg');
```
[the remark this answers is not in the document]
rmz38: How come the season emojis use require?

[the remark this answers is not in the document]
willespencer: That's probably just what was intuitive to whoever wrote it and we didn't catch it before. I personally think leaving it as require would be better for now since it is consistent with the other emojis like you mentioned - but I can make a Monday item to replace all requires with imports later.

I just tested it by the way, and a line like this works fine. 
`import surfing from '@/assets/images/surfing.svg';`

src/components/Dashboard.vue
```diff
@@ -133,7 +182,11 @@ export default {
             this.subjectColors = doc.data().subjectColors;
             this.uniqueIncrementer = doc.data().uniqueIncrementer;
             this.loaded = true;
+            this.requirements_tooltip_seen = doc.data().requirements_tooltip_seen;
           } else {
+            this.semesters.push(this.createSemester([], this.getCurrentSeason(), this.getCurrentYear()));
```
willespencer, line 187: This creates an empty semester by default right? Nice

src/components/Dashboard.vue
```diff
@@ -167,14 +220,30 @@ export default {
...
+        currentSeason = 'Summer';
+      } else {
+        currentSeason = 'Fall';
+      }
+      return currentSeason;
+    },
+    getCurrentYear() {
+      const currentYear = new Date().getFullYear();
+      return this.yearText || this.year || currentYear;
+    },
```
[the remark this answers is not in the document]
willespencer: Same with the getCurrentSeason method! Especially since it is also used in NewSemester.vue, we could avoid code duplication by doing this.

src/components/Modals/TourWindow.vue
willespencer: Bringing this old comment back :) `v-on:click` can just be `@click` here and throughout this file

src/components/Requirements.vue
```diff
@@ -1,6 +1,13 @@
 <template v-if="semesters">
   <div class="requirements">
-    <div class="fixed">
+    <div id="req-tooltip" class="fixed"
```
willespencer, line 3: Let's not keep this div if it is not going to show up at all on CoursePlan right now? Add a comment here and comment out if you think the code is worth preserving, otherwise delete it.

willespencer (changes requested):
This is great! Just left a few comments about some small things to resolve first before merging, and replied to a few of Han's. 

Also, just wanted to mention that the actual walkthrough modals are a bit off from the designs. The font, the padding, and the button styling are all a bit off, and the step counter ('1/4', '2/4', etc) is missing from each modal. No need to worry about that until this pull request is merged in since the functionality is all there, but let's talk about when & how to resolve those 😄.

src/components/Dashboard.vue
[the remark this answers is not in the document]
rmz38: Han said he will take care of code unification :))

willespencer (approved):
Awesome! Just fix the merge conflict and then it can be merged in!

hanywang2 (approved):
LGTM!

PR #122, opened 2020-05-09
Allow users to switch between two options to fulfill a requirement

einc:
### Summary 

This pull request is the first step towards implementing a feature to allow users to switch between options to fulfill a requirement. For example, Bio majors are required to fulfill the General Chemistry Requirement by taking either CHEM 2070 and CHEM 2080 or just CHEM 2150. Another example is the A&S FL requirement. 

- [x] added 2 examples of where this would be needed (bio and FL requirements)
- [x] added two optional fields to the JSON to pair options for a requirement and select a default to show
- [x] can switch between two options for the requirement! 
- [ ] didn't implement frontend (no dropdown). i put a button placeholder with the name of the paired requirement for now. 
- [ ] selected option is **not** saved to firebase (so the requirement resets to default when requirement bar is updated)

### Test Plan 
- [ ] check that total fulfilled and total requirements are accurate and that options don't count as extra requirements
- [ ] check that show completed doesn't show when there are completed 'hidden' requirements
Tested switching between options for the Bio requirement. Tested switching when both were in on-going requirements, both in completed, and one in on-going and one in completed.
ongoing
filled


SamChou19815:
I think everything in this PR should already be handled by my recent toggleable requirements PRs merged into master, so I'm going to close this. After I added support for saving data into firestore, I will close another one.
@einc correct me if I'm wrong.

PR #125, opened 2020-05-09
Add communications requirements major data

hanywang2:
### Summary 

Add requirements for the communications major with info from the communications department requirements page https://cals.cornell.edu/education/degrees-programs/communication-major-minor

### Test Plan 

Test that all 7 requirements satisfies the logic provided by the page source.

Screen Shot 2020-05-09 at 5 03 35 PM


src/requirements/data/majors/comm.ts
```diff
@@ -0,0 +1,156 @@
...
+    operator: 'or',
+    fulfilledBy: 'credits',
+    minCount: 6
+  },
+  {
+    name: 'Focus Area Upper Level',
+    description: 'Students must take six credits (two courses) of coursework within their declared Focus Area at the 3100+ level.',
+    source: 'https://communication.cals.cornell.edu/undergraduate-program/major-requirements/',
+    checker: includesWithSingleRequirement(
+      'COMM 31**',
```
einc, line 39: i think ur missing COMM 2179, MUSIC/PMA/COML/ENG 2703

einc: wait actually doing the *** includes classes that shouldn't be. for example, COMM 4350 counts now when it shouldn't.

src/requirements/data/majors/comm.ts
```diff
@@ -0,0 +1,157 @@
...
+    source: 'https://communication.cals.cornell.edu/undergraduate-program/major-requirements/',
+    checker: includesWithSingleRequirement(
+      'COMM 30**',
+    ),
+    operator: 'or',
+    fulfilledBy: 'credits',
+    minCount: 3
+  },
+  {
+    name: 'First-Year Writing Workshop (FWS)',
```
willespencer, line 122: Isn't this covered by college requirements? Or is it an additional one

hanywang2: That's actually a really interesting point. I believe that in COMM, you only need 1 FWS unlike all other majors in CALS. Yes, it technically is, but it's not accurate for the COMM major. This is one of the few examples where college requirements depend on the major. However, we don't support that yet

willespencer: Then we should get rid of this right? Since the req is already covered in college requirements (just at a more extreme level then it should be). And then make a Monday item in "Requirements Handling" to keep track of that discrepancy?

hanywang2: Right but then we would give the wrong impression that you need two FWS courses as a COMM major

PR #136, opened 2020-09-12
FetchCourses firebase function

tcho6319:
### Summary 
This pull request is the implementation of a Firebase function, `FetchCourses`, to fetch course data from the local `filtered-all-courses.json` file, as opposed to hitting the Cornell course roster API.

`FetchCourses` works as follows: 
- [x] Given an input of a list of course codes (i.e. ["CS 1110", "INFO 2300"]), `FetchCourses` outputs the corresponding course objects from the most recent roster or semester. 
- [x] Handles lowercase course codes case (i.e. ["cs 1110", "info 2300"]). 
- [x] Handles duplicate course codes case (i.e. for the input ["CS 1110", "CS 1110"],  `FetchCourses` would only return one course object for "CS 1110").
- [x] Logs console messages for unfetchable course codes (i.e. ["hi"]).
 

To be able to call the function from a .vue file (e.g. the Requirements component):
1) Insert these statements in the .vue file
     `import firebase from 'firebase/app';`
     `require('firebase/functions');`
     `const functions = firebase.functions();`
2) See the `triggerFetchCourses()` function below for an example of calling `FetchCourses`:
```
    // Example of triggering FetchCourses Firebase function
    triggerFetchCourses() {
       const FetchCourses = firebase.functions().httpsCallable('FetchCourses');
       FetchCourses({ courseCodes: ['CS 1110', 'INFO 2300'] }).then(result => {
         // Read result of the Cloud Function.
         console.log(result.data);
       });
     }
```

### Test Plan
1) run `npm install` in the functions folder
2) run `firebase deploy -P staging --only functions:FetchCourses`
3) Test input when testing function on Google Cloud Platform (https://console.cloud.google.com/functions/details/us-central1/FetchCourses?project=cornelldti-courseplan-dev&tab=testing): `{"data": {"courseCodes": [<string list of course codes>]} }`

### Notes
Before making PR, points to discuss:
- [x] Function is currently called `returnText`. Must rename to `FetchCourses` and ask Will for access again.
- [x] Confirm logging is right approach for unfetchable courses/inputs.
- [x] Confirm/add to Test Plan.
- [x] Currently all access has been granted for development to avoid CORS issue. Discuss if this is secure.*

*Since all access can be given per Firebase function, we have tentatively agreed that all-access may be alright for now. If concerns are raised about security, we will definitely revisit this note.

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -97,7 +97,7 @@ function logUnfetchedCourseCode(courseCode) {
   console.log("Unable to fetch course data for course code: ", courseCode);
 }
 
-exports.returnText = functions.https.onCall(data => {
+exports.FetchCourses = functions.https.onCall(data => {
   let courseCodes = data.courseCodes.map(a => a.toUpperCase());
   let courses = [];
 
--- a/functions/index.js
+++ b/functions/index.js
@@ -93,12 +93,38 @@ exports.TrackUsers = functions.https.onRequest(async (req, res) => {
       
 });
 
+function logUnfetchedCourseCode(courseCode) {
+  console.log("Unable to fetch course data for course code: ", courseCode);
+}
+
 exports.returnText = functions.https.onCall(data => {
+  let courseCodes = data.courseCodes.map(a => a.toUpperCase());
+  let courses = [];
+
+  // Assume rosters is sorted from least recent to most recent semester
+  let rosters = Object.keys(filteredAllCourses);
+  // Iterate over rosters starting from most recent semester
+  for (let i = rosters.length - 1; i >= 0 && courseCodes.length > 0; i -= 1) {
+    let rosterCourses = filteredAllCourses[rosters[i]];
+    // Fetch course objects whose code is in courseCodes
+    let fetchedCourses = rosterCourses.filter(rosterCourse =>
+      courseCodes.indexOf(rosterCourse.subject.concat(' ', rosterCourse.catalogNbr))!= -1);
+    
+    // Delete course codes of fetchedCourses from courseCodes
+    // Update courses with fetchedCourses
+    fetchedCourses.forEach((fetchedCourse) => {
+      let fetchedCourseCode = fetchedCourse.subject.concat(' ', fetchedCourse.catalogNbr);
+      courseCodes = courseCodes.filter(courseCode => courseCode != fetchedCourseCode);
+
+      courses.push(fetchedCourse);
+    });
+
+  }
+  // Log courses that could not be fetched
+  courseCodes.map(a => logUnfetchedCourseCode(a));
+  
   return {
-    returnText: data.text,
-    courses: filteredAllCourses['FA14'][0]
+    courses: courses
   };
 });
-
-
 
--- a/functions/index.js
+++ b/functions/index.js
@@ -139,6 +139,12 @@ function sortByMostRecentRosters(rosters) {
   return rosters.sort(compareRosters);
 }
 
+/** FetchCourses fetches the most recent course objects for the list of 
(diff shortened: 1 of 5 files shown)


functions/index.js
```diff
@@ -91,3 +93,38 @@ exports.TrackUsers = functions.https.onRequest(async (req, res) => {
...
+
+  }
+  // Log courses that could not be fetched
+  courseCodes.map(a => logUnfetchedCourseCode(a));
+  
+  return {
+    courses: courses
+  };
+});
+
```
hanywang2, line 130: What's the reason that this is happening on Firebase Functions? An onCall function is just an API, but it also doesn't look like you're using admin privileges to the db. Can't this happen on the frontend?

src/components/Requirements.vue
```diff
@@ -185,6 +187,10 @@ import Modal from '@/components/Modals/Modal.vue';
 import { BaseRequirement as Requirement, CourseTaken, SingleMenuRequirement } from '@/requirements/types';
 import { computeRequirements, computeRequirementMap } from '@/requirements/reqs-functions';
 
+require('firebase/functions');
```
hanywang2, line 190: I think you can `import` it rather than using `require`

functions/index.js
tcho6319: My reasoning was that importing the filtered-all-courses.json would be too heavy a load for the frontend to handle. Also, when a Firebase function is deployed, Firebase seems to package the `functions` folder (that I added filtered-all-courses.json to) to upload once, which I thought would save on time.

src/components/Requirements.vue
```diff
@@ -420,6 +424,14 @@ export default Vue.extend({
...
       this.minors = minors;
     }
+    // // Example of triggering FetchCourses Firebase function
+    // triggerFetchCourses() {
+    //   const FetchCourses = firebase.functions().httpsCallable('FetchCourses');
+    //   FetchCourses({ courseCodes: ['CS 1110', 'INFO 2300'] }).then(result => {
+    //     // Read result of the Cloud Function.
+    //     console.log(result.data);
+    //   });
+    // }
```
SamChou19815, line 434: 🤔

SamChou19815 (commented):
Please also update fetcher.ts to output the same file to `functions/filtered-all-courses.json`

https://github.com/cornell-dti/course-plan/blob/b715db58a5c857fa38aa5fe6a5ac70fa9d75ff88/src/requirements/fetcher.ts#L135-L137

functions/index.js
```diff
@@ -91,3 +93,79 @@ exports.TrackUsers = functions.https.onRequest(async (req, res) => {
...
+      return 1;
+    case 'SU':
+      return 2;
+    case 'FA':
+      return 3;
+    default:
+  }
+}
+
+function compareRosters(roster1, roster2) {
```
willespencer, line 114: Good implementation of what we talked about!

functions/index.js
```diff
@@ -91,3 +93,79 @@ exports.TrackUsers = functions.https.onRequest(async (req, res) => {
...
+    return 0;
+  }
+}
+
+function sortByMostRecentRosters(rosters) {
+  // Sorts from most recent roster
+  return rosters.sort(compareRosters);
+}
+
+exports.FetchCourses = functions.https.onCall(data => {
```
willespencer, line 142: Looks good!

willespencer (approved):
Looks good! Just make the change Sam mentioned and let me figure out the security issue, then you can merge!

willespencer (approved):

hanywang2:
A little late to the party but great job @tcho6319!

Issue #139, opened 2020-09-17
[RFC] New Architecture for Requirement Computation


SamChou19815:
### Current setup and issues

Currently, we compute requirement fulfillment in a quite naive way. We first go through a list of all requirements, attach courses taken by the user to them, and then utilize these **local** information to decide the fulfillment progress.

The current setup can solve some simple cases, but it's not scalable when we consider these complications:

- Some classes cannot be double-counted. If we use a local `requirement -> class[]` map, it would be hard to filter them out.
- Some requirements have multiple fulfillment options.
- We want to have the ability for user to choose between several different fulfillment options.

#### Insights

The main issue with the current implementation is that we are looking at the whole picture only from the perspective of requirements. In reality, the "viewpoint" from a class is also important. (e.g. filter away double counting) Therefore, we natually found an abstraction for a two-way view: GRAPH.

#### Mathematical definition

The graph is not a general-purposed graph. Instead, we add a few restricions to the graph. Formally, let's define `CoursePlanRequirementFulfillmentGraph` using the standard notation: `(R \union C, E)`, where `R` means a set of requirements to satisfy, `C` means a set of courses that is taken by user, and `(r, c) \in E` is a tuple denoting the edge between a requirement node `r` and a course node `c`. This graph can also be thought as a mathematical relation between `R` and `C`. Thus, the problem of computing requirements can be thought as building and refining this graph.

#### Graph Interface

Programmatically, we can define an interface like:

```ts
interface CoursePlanRequirementFulfillmentGraph {
  // These are the minimal required methods.
  getAllRequirements(): readonly Requirement[];
  getAllCourses(): readonly Course[];
  getAllEdges(): readonly (readonly [Requirement, Course])[];
  addEdge(requirement: Requirement, course: Course): void;
  removeEdge(requirement: Requirement, course: Course): void;
  
  // Some helper methods can be added for efficiency or usability, like
  getConnectedCoursesFromRequirement(requirement: Requirement): readonly Course[];
  getConnectedRequirementsFromCourse(course: Course): readonly Requirement[];
  existsEdge(requirement: Requirement, course: Course): boolean;
  // ...
}
```

#### Algorithm

The algorithm has a simple interface:

```ts
const computeCoursePlanRequirementGraph = (
  requirementsNeedToSatisfy: readonly Requirement[],
  coursesTakenByUser: readonly Course[],
  userChoicesOfRequirementFulfillmentOptions: readonly (readonly [Course, Decision])[])
[...]

PR #140, opened 2020-09-24
Setup requirement graph data structure

SamChou19815:
### Summary 

This PR starts to implement proposal in #139 by setup the `RequirementFulfillmentGraph` data structure.

I made a generic `RequirementFulfillmentGraph` that doesn't depend on a specific representation of requirement or course object, so that the structure itself can be developed independently to avoid potential changes from requirement or course representation. It contains all the methods I described in #139. All the methods are tested in Jest.

Inside my implementation of graph, we need to lookup all courses by a requirement and all requirements by a course, so I setup two hash maps for O(1) lookup. `addEdge` and `removeEdge` ensures that they are in sync. 

Unlike Java where every object has `equals` and `hashCode`, we don't have the same luxury in JS. Therefore, I also implemented `HashMap` and `HashSet` that takes in a function `getUniqueHash` for the key type to simulate the `hashCode` method in Java. The basic idea is that we call this function to get a string that can uniquely identify a requirement/course, so that we can use that unique identifier as a key for quick lookup using JavaScript's `Map`. (samlang uses the same approach: 
https://github.com/SamChou19815/samlang/blob/86ffed2edcb58573b34dae94bbea83e80c591d98/samlang-core-utils/index.ts#L54-L134)

### Test Plan 

Added unit tests for `HashMap`, `HashSet` and `RequirementFulfillmentGraph`
Added a step for `npm run test` in CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "tsc": "tsc",
     "build:staging": "vue-cli-service build --mode staging",
     "format": "prettier --write '**/*.{js,vue,scss,css,html}'",
-    "format:check": "prettier --check '**/*.{js,vue,scss,css,html}'"
+    "format:check": "prettier --check '**/*.{js,vue,scss,css,html}'",
+    "test": "jest"
   },
   "dependencies": {
     "@firebase/analytics": "^0.2.13",
@@ -35,6 +36,10 @@
     "vuex": "^3.1.2"
   },
   "devDependencies": {
+    "@babel/core": "^7.11.6",
+    "@babel/preset-env": "^7.11.5",
+    "@babel/preset-typescript": "^7.10.4",
+    "@types/jest": "^26.0.14",
     "@types/node-fetch": "^2.5.5",
     "@typescript-eslint/eslint-plugin": "^2.21.0",
     "@typescript-eslint/parser": "^2.21.0",
@@ -44,8 +49,10 @@
     "@vue/eslint-config-airbnb": "^4.0.1",
     "@vue/eslint-config-typescript": "^4.0.0",
     "babel-eslint": "^10.0.1",
+    "babel-jest": "^26.3.0",
     "eslint": "^5.16.0",
     "eslint-plugin-vue": "^5.0.0",
+    "jest": "^26.4.2",
     "node-sass": "^4.13.1",
     "sass-loader": "^8.0.2",
     "typescript": "~3.7.5",
--- a/src/requirements/__test__/requirement-graph.test.ts
+++ b/src/requirements/__test__/requirement-graph.test.ts
@@ -1,0 +1,63 @@
+import RequirementFulfillmentGraph from '../requirement-graph';
+
+it('RequirementFulfillmentGraph works.', () => {
+  // We mock a requirement graph built using plain strings to represents requirement and courses.
+  const graph = new RequirementFulfillmentGraph<string, string>(
+    s => s,
+    s => s
+  );
+
+  graph.addEdge('CS3410/CS3420', 'CS 3410');
+  graph.addEdge('CS3410/CS3420', 'CS 3420');
+  graph.addEdge('Probability', 'MATH 4710');
+  graph.addEdge('Elective', 'MATH 4710');
+
+  expect(graph.getAllRequirements()).toEqual(['CS3410/CS3420', 'Probability', 'Elective']);
+  expect(graph.getAllCourses()).toEqual(['CS 3410', 'CS 3420', 'MATH 4710']);
+  expect(graph.getAllEdges()).toEqual([
+    ['CS3410/CS3420', 'CS 3410'],
+    ['CS3410/CS3420', 'CS 3420'],
+    ['Probability', 'MATH 4710'],
+    ['Elective', 'MATH 4710']
+  ]);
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -13,7 +13,8 @@
     "tsc": "tsc",
     "build:staging": "vue-cli-service build --mode staging",
     "format": "prettier --write '**/*.{js,vue,scss,css,html}'",
-    "format:check": "prettier --check '**/*.{js,vue,scss,css,html}'"
+    "format:check": "prettier --check '**/*.{js,vue,scss,css,html}'",
+    "test": "jest"
```
hanywang2, line 17: amazing

src/requirements/requirement-graph.ts
```diff
@@ -0,0 +1,91 @@
...
+    const edges: (readonly [Requirement, Course])[] = [];
+    this.requirementToCoursesMap.forEach((courses, requirement) => {
+      courses.forEach(course => edges.push([requirement, course]));
+    });
+    return edges;
+  }
+
+  public addEdge(requirement: Requirement, course: Course): void {
+    let existingCoursesLinkedToRequirement = this.requirementToCoursesMap.get(requirement);
+    if (existingCoursesLinkedToRequirement == null) {
```
hanywang2, line 47: '== null' is a great practice that I realized is the industry standard for JS this summer. Should better standardize that in CP

src/requirements/requirement-graph.ts
```diff
@@ -0,0 +1,91 @@
+import { HashMap, HashSet } from './util/collections';
+
+/**
+ * A mutable graph data structure that represents the mathematically relation between requirement
+ * and course.
+ * An edge between requirement `r` and course `c` implies that `c` may help satisfy `r`. It's up to
+ * the client of the data structure to gradually refine the graph so that an edge `r-c` implies that
+ * `c` definitely helps satisfies `r`.
```
hanywang2, line 8: so would it continue after 'r-c' to 'r-c-r'? How does the relationship between 'c-r' work and what does that mean?

src/requirements/requirement-graph.ts
```diff
@@ -0,0 +1,91 @@
...
+    this.requirementToCoursesMap = new HashMap(getRequirementUniqueID);
+    this.courseToRequirementsMap = new HashMap(getCourseUniqueID);
+  }
+
+  public getAllRequirements(): readonly Requirement[] {
+    return this.requirementToCoursesMap.keys();
+  }
+
+  public getAllCourses(): readonly Course[] {
+    return this.courseToRequirementsMap.keys();
```
hanywang2, line 34: Are the keys here the course codes? ie CS 2110, INFO 2300, etc..?

hanywang2 (commented):
Just have some thoughts from the initial read of the PR. Gonna let @willespencer approve or RC this. requirement-graph.ts seems to be the meat of the change and an important one to understand so I would only comfortably approve this after either Will or I run it locally and log it through to understand how it works

src/requirements/requirement-graph.ts
SamChou19815: It might be the course id from course roster so that cross-listed courses are merged into one

src/requirements/requirement-graph.ts
SamChou19815: No, the graph is undirected. So `r-c` automatically implies `c-r`

src/requirements/util/collections.ts
```diff
@@ -0,0 +1,87 @@
+export class HashMap<K, V> {
```
willespencer, line 1: I don't think I've ever had to use a HashMap in Javascript before - are you just using the Map() but setting the keys as hash values using the getUniqueHash function?

willespencer (approved):
Looks great! Mostly just left some clarifying questions for my own information. I also ran `npm run test` and  verified that everything you wrote passes!

src/requirements/util/collections.ts
SamChou19815: Yes. I create a class for it to encapsulate the implementation detail so that the graph itself doesn't need to worry about it.

src/requirements/__test__/requirement-graph.test.ts
```diff
@@ -0,0 +1,63 @@
+import RequirementFulfillmentGraph from '../requirement-graph';
+
+it('RequirementFulfillmentGraph works.', () => {
```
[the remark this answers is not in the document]
SamChou19815: I think these tests would be sufficient on its own. It always covers some common setup and tricky cases (e.g. removing on edge causing a requirement to have no linked courses). Additional unit tests should be done on business logic itself instead of this abstract graph. e.g. testing whether all double counting has been correctly eliminated in a phase.

src/requirements/util/collections.ts
willespencer: Yup that makes sense!

src/requirements/__test__/requirement-graph.test.ts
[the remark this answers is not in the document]
willespencer: Yeah that's more what I was thinking about, but you're right the graph itself should be good to go

PR #142, opened 2020-10-01
Implement phase 1, 2-1, 3-1 for requirement graph builder

SamChou19815:
### Summary 

This PR implements part of the 3-phase algorithm defined in this issue comment: https://github.com/cornell-dti/course-plan/issues/139#issuecomment-694641551. I omitted phase 2-2 and phase 3-2 in this PR to keep it relatively small. Phase 2-2 and 3-2 are tricky enough to deserve their own PRs.

Similar to #140, I keep the type of requirement, course and user choice type generic, so that the algorithm itself is flexible enough to allow different types of requirement, course and user choice representation. Information is extracted from those objects via functions like `getCorrespondingRequirementAndAllRelevantCoursesUnderFulfillmentStrategy`. In this way, we can easily write unit tests that can have a very different representation of `Requirement`, `Course`, etc compared to prod code. As a result, we can write tests using raw strings as requirement/courses to get easily-understable test assertions. It also allows us to keep this function unchanged while we are changing concrete requirement and course type for frontend UI
purposes.

### Test Plan 

`npm run test`

The added test no only tests the function, but also serves as a tutorial on how to use my `buildRequirementFulfillmentGraph` function.

### DIFF
--- a/src/requirements/__test__/requirement-graph-builder.test.ts
+++ b/src/requirements/__test__/requirement-graph-builder.test.ts
@@ -1,0 +1,151 @@
+import buildRequirementFulfillmentGraph from '../requirement-graph-builder';
+
+const requirements = ['CS3410/CS3420', 'Probability', 'Elective'];
+const getUniqueID = (it: string) => it;
+const getAllCoursesThatCanPotentiallySatisfyRequirement = (
+  requirement: string
+): readonly string[] => {
+  switch (requirement) {
+    case 'CS3410/CS3420':
+      return ['CS3410', 'CS3420'];
+    case 'Probability':
+      return ['MATH4710'];
+    case 'Elective':
+      return ['CS3410', 'CS3420', 'MATH4710'];
+    default:
+      throw new Error('Should not get here!');
+  }
+};
+// If choice is 1, pick 3410 fulfillment strategy, else, choose 3420
+const getCorrespondingRequirementAndAllRelevantCoursesUnderFulfillmentStrategy = (
+  choice: 1 | 2
+) => ({
+  correspondingRequirement: 'CS3410/CS3420',
+  coursesOfChosenFulfillmentStrategy: [choice === 1 ? 'CS3410' : 'CS3420']
+});
+
+it('buildRequirementFulfillmentGraph phase 1 test 1', () => {
+  const graph = buildRequirementFulfillmentGraph({
+    requirements,
+    userCourses: ['CS3410', 'CS3420', 'MATH4710'],
+    userChoiceOnFulfillmentStrategy: [],
+    userChoiceOnDoubleCountingElimiation: [],
+    getRequirementUniqueID: getUniqueID,
+    getCourseUniqueID: getUniqueID,
+    getAllCoursesThatCanPotentiallySatisfyRequirement,
+    getCorrespondingRequirementAndAllRelevantCoursesUnderFulfillmentStrategy
+  });
+
+  expect(graph.getConnectedCoursesFromRequirement('CS3410/CS3420')).toEqual(['CS3410', 'CS3420']);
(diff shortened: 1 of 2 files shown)


src/requirements/requirement-graph-builder.ts
```diff
@@ -0,0 +1,122 @@
...
+   * requirements do not allow double counting, then a tuple (r1, c) means that we should keep the
+   * (r1, c) edge in the graph and drop the (r2, c) edge.
+   */
+  readonly userChoiceOnDoubleCountingElimiation: readonly (readonly [Requirement, Course])[];
+  /** See RequirementFulfillmentGraph for its usage. */
+  readonly getRequirementUniqueID: (r: Requirement) => string;
+  /** See RequirementFulfillmentGraph for its usage. */
+  readonly getCourseUniqueID: (c: Course) => string;
+  /**
+   * Aaively give a list of courses that can satisfy a requirement. Most of the time this function
```
willespencer, line 36: Small typo "Aaively"

src/requirements/requirement-graph-builder.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    graph.getConnectedCoursesFromRequirement(correspondingRequirement).forEach(connectedCourse => {
+      if (!coursesToKeepSet.has(connectedCourse)) {
+        graph.removeEdge(correspondingRequirement, connectedCourse);
+      }
+    });
+  });
+
+  // Phase 2-2:
+  //   Auto-select fulfillment for requirements with multiple fulfillment strategies
```
willespencer, line 99: Based on the current WIP design, I think users will have to select a fulfillment before adding a course on the add course modal. This modal does have default fulfillments pre-selected in the design, so would we just want default generating code there instead of auto-selecting code here?  Let me know if that makes sense or if I'm off about anything.

src/requirements/requirement-graph-builder.ts
```diff
@@ -0,0 +1,122 @@
...
+        graph.removeEdge(connectedRequirement, course);
+      }
+    });
+  });
+
+  // Phase 3-2: Auto-choose a requirement-course pair for double counted courses without user choice.
+  // TODO: not implemented.
+  // Important note for this phase: We need to support requirements that allow double counting.
+
+  // Phase MAX_INT: PROFIT!
```
willespencer, line 118: Best phase 😄

src/requirements/__test__/requirement-graph-builder.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    userCourses: ['CS3410', 'CS3420', 'MATH4710'],
+    userChoiceOnFulfillmentStrategy: [2],
+    userChoiceOnDoubleCountingElimiation: [],
+    getRequirementUniqueID: getUniqueID,
+    getCourseUniqueID: getUniqueID,
+    getAllCoursesThatCanPotentiallySatisfyRequirement,
+    getCorrespondingRequirementAndAllRelevantCoursesUnderFulfillmentStrategy
+  });
+
+  // In this case, 3410 is removed since user chooses strategy 1.
```
willespencer, line 102: Strategy 2 right?

willespencer (approved):
Left a few small comments, just a couple typos and one clarifying question! Looks really good, and thanks for making both the tests and the implementation easy to follow!

src/requirements/__test__/requirement-graph-builder.test.ts
SamChou19815: ah yes, bad copy pasta

src/requirements/requirement-graph-builder.ts
SamChou19815: if we want  default generating code then it will not be part of this graph builder phase. we can do it elsewhere.

PR #146, opened 2020-10-07
Connect requirement graph algorithm to req-functions

SamChou19815:
### Summary 

This pull request connects my graph algorithm previously developed in isolation to the req-functions that are directly interacting with frontend. This is designed to be a 1-1 migration, so it doesn't introduce the ability of eliminating double-counting and fulfillment strategy choice yet.

Some code in the graph algorithm is tweaked a little bit to help resolve some bugs, which I will explain in individual comments. The change might look scary, but it's mostly replacing this block of code

https://github.com/cornell-dti/course-plan/blob/2af5e71a01dc534b7aeb977ea6006b46eff06808/src/requirements/reqs-functions.ts#L127-L175

with my graph algorithm.

### Test Plan 

CI, and the requirement checks still work fine for me:

Screen Shot 2020-10-07 at 16 38 40

### DIFF
--- a/src/requirements/reqs-functions.ts
+++ b/src/requirements/reqs-functions.ts
@@ -214,15 +214,15 @@ export function computeRequirements(
     reqs: computeUniversityRequirementFulfillments(coursesTaken)
   });
 
-  // PART 2: check college & major requirements
+  // PART 2: check college & major & minor requirements
   if (!(college in requirementJson.college)) throw new Error('College not found.');
   const collegeReqs = requirementJson.college[college];
 
-  type DecoratedCollegeOrMajorRequirementWithSourceType = DecoratedCollegeOrMajorRequirement & {
+  type RequirementWithSourceType = DecoratedCollegeOrMajorRequirement & {
     readonly sourceType: 'College' | 'Major' | 'Minor';
     readonly sourceSpecificName: string;
   };
-  const requirementsToBeConsideredInGraph: readonly DecoratedCollegeOrMajorRequirementWithSourceType[] = [
+  const requirementsToBeConsideredInGraph: readonly RequirementWithSourceType[] = [
     ...collegeReqs.requirements.map(
       it => ({ ...it, sourceType: 'College', sourceSpecificName: college } as const)
     ),
@@ -247,7 +247,7 @@ export function computeRequirements(
   ];
 
   const requirementFulfillmentGraph = buildRequirementFulfillmentGraph<
-    DecoratedCollegeOrMajorRequirementWithSourceType,
+    RequirementWithSourceType,
     CourseTaken,
     undefined
   >({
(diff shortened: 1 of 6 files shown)


src/requirements/util/collections.ts
```diff
@@ -19,7 +19,8 @@ export class HashMap<K, V> {
   }
 
   get(key: K): V | undefined {
-    return this.backingMap.get(this.getUniqueHash(key))?.[1];
+    const entry = this.backingMap.get(this.getUniqueHash(key));
+    return entry == null ? undefined : entry[1];
```
SamChou19815, line 23: Used to overcome the Vue webpack doesn't support optional chaining yet.

src/requirements/requirement-graph-builder.ts
```diff
@@ -67,15 +67,17 @@ const buildRequirementFulfillmentGraph = <Requirement, Course, UserChoiceOnFulfi
...
-  const userCourseSet = new HashMap<Course, null>(getCourseUniqueID);
-  userCourses.forEach(course => userCourseSet.set(course, null));
+  const userCourseSet = new HashMap<Course, Course>(getCourseUniqueID);
+  userCourses.forEach(course => userCourseSet.set(course, course));
 
   // Phase 1:
   //   Building a rough graph by naively connecting requirements and courses based on
   //   `getAllCoursesThatCanPotentiallySatisfyRequirement`.
   requirements.forEach(requirement => {
+    graph.addRequirementNode(requirement);
```
SamChou19815, line 77: Without this, requirement without matching courses won't show up in the final array at all, which is bad.

src/requirements/reqs-functions.ts
```diff
@@ -285,46 +214,135 @@ export function computeRequirements(
     reqs: computeUniversityRequirementFulfillments(coursesTaken)
   });
 
-  // PART 2: check college requirements
+  // PART 2: check college & major requirements
```
willespencer, line 217: It looks like this code also checks minor requirements, could you add that to this comment

src/requirements/reqs-functions.ts
```diff
@@ -285,46 +214,135 @@ export function computeRequirements(
...
   if (!(college in requirementJson.college)) throw new Error('College not found.');
   const collegeReqs = requirementJson.college[college];
 
-  groups.push({
-    groupName: 'College',
-    specific: college,
-    reqs: computeCollegeOrMajorRequirementFulfillments(coursesTaken, collegeReqs.requirements)
+  type DecoratedCollegeOrMajorRequirementWithSourceType = DecoratedCollegeOrMajorRequirement & {
```
willespencer, line 221: Do we want to rename this to DecoratedRequirement or DecoratedCollegeMajorMinor Requirement, since it seems to handle every case now?

willespencer (approved):
Looks great! Confirmed that the test files passed and requirements on the deployment still work as expected. Just left a couple minor comments!

PR #149, opened 2020-10-13
Report illegally double-counted courses during requirement computation

SamChou19815:
### Summary 

According to our meeting in the Saturday, we decided to give up on enforcing that we have a graph free of double-counted courses all the time. Instead, we give warnings. This diff implements the decision.

Now in additional to the graph, we also return a list of **illegally** double-counted courses. The word *illegal* is important here, since we know that they are requirements that explicitly allow double-counting (e.g. probability requirement in engineering). To account for the requirement that can be double-counted, we use the return value from the new function `allowDoubleCounting: (requirement: Requirement) => boolean` passed to the graph computation algorithm.

The frontend UI can then look up the graph for each illegally double-counted course to decide what warning message to display. e.g. `MATH 4710` is in the list. After looking up the graph from `MATH 4710`, we found that it's connected to `CHEM requirement` and `external specialization` requirements that both don't allow double-counting, then we can show warning messages like `MATH 4710 is used to satisfy both CHEM requirement and external specialization, but both requirements don't allow double counting`.

### Test Plan 

`yarn test`

Modified the old unit tests to add `allowDoubleCounting` parameter. Added new unit tests on the list of the illegally double-counted courses.

### DIFF
--- a/src/requirements/reqs-functions.ts
+++ b/src/requirements/reqs-functions.ts
@@ -246,7 +246,7 @@ export function computeRequirements(
       .flat()
   ];
 
-  const requirementFulfillmentGraph = buildRequirementFulfillmentGraph<
+  const { requirementFulfillmentGraph } = buildRequirementFulfillmentGraph<
     RequirementWithSourceType,
     CourseTaken,
     undefined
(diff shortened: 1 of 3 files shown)


src/requirements/reqs-functions.ts
```diff
@@ -284,7 +284,10 @@ export function computeRequirements(
       // userChoiceOnFulfillmentStrategy
       correspondingRequirement: requirementsToBeConsideredInGraph[0],
       coursesOfChosenFulfillmentStrategy: []
-    })
+    }),
+    // TODO: Replace this dummy implementation once we decided how to determine if a requirement is
+    // double-countable. Meanwhile, make it always return true to match the old behavior.
+    allowDoubleCounting: () => true
```
willespencer, line 290: If we decided false was the default, would this function just return true if we specifically say a requirement can be double counted in our checkers or if it was a minor requirement?
And do you want to discuss how we want to handle this over Slack with maybe Ein and/or Han?

willespencer (approved):
Looks good and tests pass! Just left a clarifying question 😄

src/requirements/reqs-functions.ts
SamChou19815: Yes

PR #150, opened 2020-10-15
Created Initial Components for Requirements Bar and Factored Out RequirementView.vue and SubRequirement.vue

tcho6319:
### Summary
This pull request is the first step towards modularizing the Requirements Bar, specifically `Requirements.vue` and its child components.

- [x] Created components for `Requirements.vue` refactoring
- [x] Factored out `RequirementView.vue` from `Requirements.vue` for each Requirement (i.e. University, College, Major)
- [x] Factored out `RequirementHeading.vue` from `RequirementView.vue`
- [x] Factored out `SubRequirement.vue` from `RequirementView.vue`
- [x] Removed undefined var init in `Semester.vue`

The created components for `Requirements.vue` include: `RequirementView.vue`, `RequirementHeader.vue`, `SubRequirement.vue`, `IncompleteSubReqCourse.vue`, `IncompleteSelfCheckSubReqCourse.vue`, `CompletedSubReqCourse.vue`, and `SeeAllSubReqCourses.vue`.

- `Requirements.vue`
    - `RequirementView.vue`
         - `RequirementHeader.vue`
         - `SubRequirement.vue`
              - `IncompleteSubReqCourse.vue`
              - `IncompleteSelfCheckSubReqCourse.vue`
              - `CompletedSubReqCourse.vue`
    - `SeeAllSubReqCourses.vue`

A diagram to show the component-feature relationship:
Screen Shot 2020-10-16 at 12 09 58 PM

**Note: Although `IncompleteSelfCheckSubReqCourse.vue` is not shown in diagram, it will probably be a child component of `SubRequirement.vue`. **

The requirements bar should be replaced when a user clicks "See All":
Screen Shot 2020-10-15 at 5 55 22 PM

### Test Plan
Since this PR is a simple refactoring of `Requirements.vue` so that `RequirementView.vue` and `RequirementHeading.vue` components are for every requirement, and `SubRequirement.vue` component is used for every sub-requirement, there should be no changes to functionality and user flow. User should be able to view requirements & sub-requirements and expand/collapse description for specific sub-requirement.

### DIFF
--- a/src/components/RequirementHeader.vue
+++ b/src/components/RequirementHeader.vue
@@ -1,0 +1,237 @@
+<template>
+  <div class="requirementheader">
+    <!-- TODO change for multiple colleges -->
+    <div v-if="index<=2 || index == 2 + majors.length" class="row top">
+      <p class="name col p-0">{{ req.name }}</p>
+    </div>
+      <!-- TODO change for multiple colleges -->
+      <div v-if="index==2" class="major">
+        <div :style="{'border-bottom': major.display ? `2px solid #${reqGroupColorMap[req.group][0]}` : ''}"
+          @click="activateMajor(id)" class="major-title" v-for="(major, id) in majors" :key="major.id">
+          <p :style="{'font-weight': major.display ? '500' : '', 'color' : major.display ? `#${reqGroupColorMap[req.group][0]}` : ''}"  class="major-title-top">{{major.majorFN}}</p>
+          <p :style="{'color': major.display ? `#${reqGroupColorMap[req.group][0]}` : ''}" class="major-title-bottom">({{user.collegeFN}})</p>
+        </div>
+      </div>
+      <div v-if="index==2+majors.length" class="minor">
+        <div :style="{'border-bottom': minor.display ? `2px solid #${reqGroupColorMap[req.group][0]}` : ''}"
+          @click="activateMinor(id)" class="major-title" v-for="(minor, id) in minors" :key="minor.id">
+          <p :style="{'font-weight': minor.display ? '500' : '', 'color' : minor.display ? `#${reqGroupColorMap[req.group][0]}` : ''}"  class="minor-title-top">{{minor.minorFN}}</p>
+          <!-- <p :style="{'color': minor.display ? `#${reqGroupColorMap[req.group][0]}` : ''}" class="minor-title-bottom">({{user.collegeFN}})</p> Change for multiple colleges -->
+        </div>
+      </div>
+
+      <!-- progress bar settings -->
+      <div v-if="showMajorOrMinorRequirements" >
+        <div class="progress">
+          <div
+            class="progress-bar"
+            :style="{ 'background-color': `#${reqGroupColorMap[req.group][0]}`, width: `${(req.fulfilled/req.required)*100}%`}"
+            role="progressbar"
+          ></div>
+        </div>
+
+        <p class="progress-text">
+          <span class="progress-text-credits">{{ req.fulfilled }}/{{ req.required }}</span>
+          <span class="progress-text-text"> Total {{ req.type }} Inputted on Schedule</span>
+        </p>
+
+        <!--View more college requirements -->
+        <div class="row top">
+          <div class="col-1 p-0" >
+            <button :style="{ 'color': `#${reqGroupColorMap[req.group][0]}` }" class="btn" @click="toggleDetails(index)">
+              <!-- svg for dropdown icon -->
+              <img
+                v-if="req.displayDetails"
+                class="arrow arrow-up"
+                :src="require(`@/assets/images/dropup-${reqGroupColorMap[req.group][1]}.svg`)"
+                alt="dropup"
+              />
+              <img
+                v-else
+                class="arrow arrow-down"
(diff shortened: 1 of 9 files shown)


src/components/SubRequirement.vue
```diff
@@ -0,0 +1,186 @@
...
+          class="arrow arrow-down"
+          src="@/assets/images/dropdown.svg"
+          alt="dropdown"
+        />
+        <img
+          v-if="!subReq.displayDescription && isCompleted"
+          class="arrow arrow-down"
+          src="@/assets/images/dropdown-lightgray.svg"
+          alt="dropdown"
+        />
```
tcho6319, line 30: I was trying to find a way to conditionally change src value like how I did for class but I didn't have any luck. If anyone has suggestions to improve this, please lmk!

src/components/Semester.vue
```diff
@@ -205,7 +205,7 @@ export default {
       this.isShadow = false;
       this.isDraggedFrom = false;
     });
-    init = false;
+    const init = false;
```
SamChou19815, line 208: this init is probably useless?

src/components/SubRequirement.vue
```diff
@@ -0,0 +1,186 @@
...
+    line-height: 14px;
+  }
+}
+.separator {
+  height: 1px;
+  width: 100%;
+  background-color: #d7d7d7;
+}
+
+</style>
```
SamChou19815, line 186: the css seems to be copied over from Requirements.vue, maybe you should also delete those copied css in Requirements.vue

src/components/SubRequirement.vue
willespencer: The best thing I can think of right now is calling a function in src that returns the correct string. So defining something like dropImage() that returns the up/down/gray variants depending on those two bools. If you do that you would probably only need one  tag so it might be better? Not sure though.

willespencer (changes requested):
This looks great! Responded to your comment and also have a couple more questions...

1. Do we also want a component for the top of each section? Like the "College requirements" and progress bar?
2. When the new self-check system gets implemented, how do you envision them fitting in to this structure? When completed, the requirement looks essentially the same, so using that same component makes sense, but what do you think about when they are incomplete?

src/components/Semester.vue
tcho6319: Yes I don't really know the purpose of this init in Semester.vue. I will try removing it and see if anything breaks.

src/components/SubRequirement.vue
tcho6319: @willespencer Yeah I tried to define a function that did this but for some reason the value was a str and not the actual relative path if that makes sense. So the string itself would correctly be "@/assets/images/dropdown-lightgray.svg" but it was as if I didn't have a relative path at all. It would as if I typed in any other string like "hi". Would it be ok to keep it like this for now or should I try to find another way to make the code cleaner?

tcho6319:
> This looks great! Responded to your comment and also have a couple more questions...
> 
> 1. Do we also want a component for the top of each section? Like the "College requirements" and progress bar?
> 2. When the new self-check system gets implemented, how do you envision them fitting in to this structure? When completed, the requirement looks essentially the same, so using that same component makes sense, but what do you think about when they are incomplete?

@willespencer 
1. We can create a component for the top of each section. We would probably have to handle the multiple majors and minors tab, but I think that would be doable. Should we then create a `Requirement.vue` component that would include the top section component and `SubRequirement.vue`?

2. Yes, I can create an `IncompleteSelfCheckSubReqCourse.vue` component. I would imagine that this component would be at the same level as the `IncompleteSubReqCourse.vue` and `CompletedSubReqCourse.vue`, under `SubRequirement.vue`.

src/components/RequirementHeader.vue
```diff
@@ -0,0 +1,237 @@
...
+          </div>
+        </div>
+      </div>
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    index: Number,
```
SamChou19815, line 73: we should give this a more description name

tcho6319: Sure I can do `reqIndex` since it is the index of the element in the reqs array.

src/components/SubRequirement.vue
willespencer: Resolved over slack!

willespencer:
> > This looks great! Responded to your comment and also have a couple more questions...
> > 
> > 1. Do we also want a component for the top of each section? Like the "College requirements" and progress bar?
> > 2. When the new self-check system gets implemented, how do you envision them fitting in to this structure? When completed, the requirement looks essentially the same, so using that same component makes sense, but what do you think about when they are incomplete?
> 
> @willespencer
> 
> 1. We can create a component for the top of each section. We would probably have to handle the multiple majors and minors tab, but I think that would be doable. Should we then create a `Requirement.vue` component that would include the top section component and `SubRequirement.vue`?
> 2. Yes, I can create an `IncompleteSelfCheckSubReqCourse.vue` component. I would imagine that this component would be at the same level as the `IncompleteSubReqCourse.vue` and `CompletedSubReqCourse.vue`, under `SubRequirement.vue`.

Yes, both of these solutions make sense to me! Thanks Theresa 😄

willespencer (approved):
Looks great! Merge it in when you're ready 😄

PR #152, opened 2020-10-17
Make all Buttons/Dropdowns have Pointers

hahnbeelee:
### Summary 
Went through the website and added `cursor: pointer;` to places where they were needed.

### Test Plan
Check all dropdowns and buttons and make sure that when you hover over them, the cursor becomes a pointer


src/components/Modals/CourseMenu.vue
```diff
@@ -191,7 +191,7 @@ export default {
     justify-content: space-between;
     padding: 0.5rem 1rem;
     position: relative;
-
+    cursor: pointer;
```
hanywang2, line 194: @willespencer so typically when a style is repeated, I have the opinion that it's better to just put that into its own class like "cursor-pointer" than to put it into every class. Wdyt?

hanywang2 (commented):
Left some thoughts but generally LGTM

src/components/Modals/CourseMenu.vue
hahnbeelee: yeah, i was thinking about doing that initially and then i realized that i was adding the class pointer everywhere on the template. so, i kinda looked at the pre-existing classes, and concluded that adding the 1-liner `cursor: pointer;` to the css seemed more concise.

src/components/RequirementHeader.vue
```diff
@@ -6,14 +6,14 @@
     </div>
       <!-- TODO change for multiple colleges -->
       <div v-if="reqIndex==2" class="major">
-        <div :style="{'border-bottom': major.display ? `2px solid #${reqGroupColorMap[req.group][0]}` : ''}"
+        <div :style="{'border-bottom': major.display ? `2px solid #${reqGroupColorMap[req.group][0]}` : '', 'cursor': majors.length > 1 ? 'pointer' : '' }"
```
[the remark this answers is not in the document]
hahnbeelee: addressed!

src/components/Modals/CourseMenu.vue
```diff
@@ -42,8 +42,8 @@
           <span class="courseMenu-text">Edit Credits</span>
         </div>
         <img v-if="!isLeft" class="courseMenu-arrow" src="@/assets/images/sidearrow.svg" />
-         <div v-if="displayEditCourseCredits" class="courseMenu-content courseMenu-editCredits courseMenu-centerCredits"
-         :class="{'courseMenu-editCredits--left': isLeft }">
+          <div v-if="displayEditCourseCredits" class="courseMenu-content courseMenu-editCredits courseMenu-centerCredits"
```
willespencer, line 45: Ok this is super nitpicky - but was this improperly indented before or did you accidently change it?

src/components/Modals/CourseMenu.vue
willespencer: Yeah what Han said is def good practice, but I also agree with Hahnbee that this is more concise in this specific case, plus we don't really do that anywhere in our codebase as of now 😕. So leaning towards leaving it as is.

src/components/RequirementHeader.vue
```diff
@@ -133,8 +140,8 @@ export default {
   }
 }
 .btn:focus,.btn:active {
-   outline: none !important;
-   box-shadow: none;
+  outline: none !important;
```
willespencer, line 143: Same question here as above.

willespencer (approved):
Looks great! Everything seems to now have the correct pointers, except the Onboarding modal but that will need to be updated later anyway 😄. Just left a small comment, but once you respond and I look at it you can merge this in!

src/components/Modals/CourseMenu.vue
hahnbeelee: i think every indentation change i made was because my extension "indent rainbow" shows me it's incorrect - this is how it looked before i changed it.
Screen Shot 2020-10-18 at 9 24 53 PM

PR #153, opened 2020-10-21
Make crosslisted courses count towards fulfillment even if subject doesn't match

SamChou19815:
### Summary 

This PR implements my strategy to use `crseId` from course roster as the unique identifier for each class. `crseId` has a nice property that two courses which are crosslisted of each other will have the same `crseId`. So using `crseId` as unique ID can make crosslisted courses requirement fulfillment automatically work, since they really appear the same to the graph algorithm. This PR also provides a solid foundation for AP/IB/Transfer credits since now everything has a `crseId` now.

To make it work, `courseTaken` must have `courseId`, which we currently don't store in the database. Therefore, I also updated the `createCourse` code to store the new field.

### Test Plan 

Before:
before
After:
after

You can see that everything is still the same, except that engineering distribution is now `4/6`, which means that `cs 2112` is now correctly counted!

### Notes  

I really don't like the horribly untyped course that does magic on the courses. Hopefully they will all be in TS in one day.

### Breaking Changes  

**THIS IS A BREAKING CHANGE!**
**THIS IS A BREAKING CHANGE!**
**THIS IS A BREAKING CHANGE!**

You need to delete and readd all of your courses, since it needs to record `crseId` in db now.


src/components/Dashboard.vue
```diff
@@ -337,6 +337,7 @@ export default {
       const alerts = { requirement: null, caution: null };
 
       const newCourse = {
+        crseId: course.crseId,
```
willespencer, line 340: Could we add a check here to set crseId to something if course.crseId is not defined? Not really sure what that would be though...

src/requirements/reqs-functions.ts
```diff
@@ -257,23 +257,22 @@ export function computeRequirements(
...
-                });
-              });
-            });
+          Object.entries(eligibleCourses).forEach(([roster, courseIds]) => {
+            courseIds.forEach(courseId => courses.push({
+              roster,
+              courseId,
+              // Only roster and courseId are used for equality comparison,
+              // so other dummy values doesn't matter.
+              code: 'DUMMY',
```
willespencer, line 271: Is there a reason you are setting these to 'DUMMY' instead of just removing the parameters altogether from the typescript definition?

willespencer (changes requested):
Left a couple comments! My main concern right now is that this is a breaking change. I think we've been operating under the assumption for a while that beta users would be able to keep their current data before our launch, but merging this in would prevent that right? 

I guess I'm wondering if there are any solutions that would let us preserve old data? Something like adding a check if a course does not have the id and assigning it when this occurs, adding some sort of dummy id, or making a script to update existing data. But also if you think that deleting user data before launch is inevitable then that would be good to know as well.

src/components/Dashboard.vue
SamChou19815: A default value doesn't make sense since it will mess up with the computation

src/requirements/reqs-functions.ts
SamChou19815: It's still used for university req computation

https://github.com/cornell-dti/course-plan/blob/b87611aad3ab68d93136bb551900a8678e644c5f/src/requirements/reqs-functions.ts#L83-L84

SamChou19815:
> I guess I'm wondering if there are any solutions that would let us preserve old data? Something like adding a check if a course does not have the id and assigning it when this occurs, adding some sort of dummy id, or making a script to update existing data. But also if you think that deleting user data before launch is inevitable then that would be good to know as well.

Using script to update data is the one I'm thinking about. I can add the script later.

SamChou19815:
@willespencer I added a script that can be run by `cd functions && node update-user-data.js`. I already used it to update all dev db data.

willespencer (approved):
Looks good! Thanks for writing the script 😄 

Do we want to run this on the prod db now as well? Or hold off for a bit?

SamChou19815:
> Looks good! Thanks for writing the script 😄 
> 
> 
> 
> Do we want to run this on the prod db now as well? Or hold off for a bit?

It's not very meaningful to run it now, since we still need to run it again before another prod release.

Issue #138, opened 2020-09-16
Existing requirement computation issues


SamChou19815:
Some of my opinions on how we can make the existing requirement computation more maintainable. None of the issues below should block merging a PR.

### Types of `operator` (`'and' | 'or' | null`) is very confusing.

They don't seem to relate to double counting. A better name would be `requiredCourseCheckerStrategy: 'all-of' | 'any-of'`.

### Bloated `BaseRequirement`

Instead of combining `credits` | `courses` | `self-check` requirements into a single type. We should use TypeScript's disjointed union capability to split them into 3 different types, and do pattern matching on them. In this way, we can get rid of a bunch of `null` and `undefined` possibilibity that are just added for `self-check` types of courses.

### Requirement with several fulfillment strategies

Related to the above issue, we probably should create a new type for requirement like

```typescript
type RequirementWithMultipleFulfillmentStrategy = {
  readonly type: 'RequirementWithMultipleFulfillmentStrategy'
  readonly strategies: readonly Requirement[]
}

type Requirement = 
  | RequirementToBeFulfilledByCourses
  | RequirementToBeFulfilledByCredits
  | RequirementForSelfCheck
  | RequirementWithMultipleFulfillmentStrategy
```

willespencer:
These all look like good improvements! I'll add them to Monday

PR #161, opened 2020-10-31
Specialize type of self-check requirements in requirements data

SamChou19815:
### Summary 

In the past, we are not very careful about distinguishing between types of requirements with different fulfillment strategy. The worst example is self-check requirement. Since self-check requirement isn't treated as a real requirement (as far as requirement computation is concerned), a lot of fields for other requirements do not make sense for self-check requirements. e.g. `eligibleCourses, totalCount, checker` etc.

In particular, we try to use one single type to describe all three kinds of requirements fulfillment strategies: courses, credits, self-check. As a result, we end up with some ugly types like:

https://github.com/cornell-dti/course-plan/blob/0948b89260c29ec29dd166d3e8311afadbc83b88/src/requirements/types.ts#L39-L53

https://github.com/cornell-dti/course-plan/blob/0948b89260c29ec29dd166d3e8311afadbc83b88/src/requirements/types.ts#L64-L66

, where the null case only makes sense for self-check courses. This kind of types make reasoning correctness of the code much harder, since we now have to consider whether the null case is for self-check or not, instead of deal with self-check requirements in the first line of code and have much fewer cases to worry about.

**This PR is the first step to reverse that BAD DESIGN DECISION.**. It implements the proposal in #138 and uses typescript's [disjointed union](https://basarat.gitbook.io/typescript/type-system/discriminated-unions) feature to enforce compile-time exhaustiveness check. After this PR, we can remove a bunch of null values for self-check requirements.

The value of this PR goes beyond simply writing fewer lines of code for self-check requirements. **This PR paves the way for implementing toggle-able requirements in a principled way**. With the infra setup by this PR, we can later support toggle-able requirements by adding another variant to `RequirementFulfillmentInformation`, making it something like:

```typescript
/**
 * @param A additional information only attached to credits and courses type.
 * @param B additional information only attached to credits and courses type.
 */
type RequirementFulfillmentInformation<A = {}, B = {}> =
  | {
      readonly fulfilledBy: 'self-check';
      // Currently unused.
      readonly minCount?: number;
    }
  | ({
      /** Defines whether courses are 'double counted': and for no double counting and or for double counting */
      readonly operator: 'and' | 'or';
      readonly fulfilledBy: 'credits' | 'courses';
      /**
       * The minimum count required to fulfill this requirement.
       *
       * - When fulfilledBy === 'credits', this field stores the min number of credits.
       * - When fulfilledBy === 'courses', this field stores the min number of courses.
       */
      readonly minCount: number;
      /**
       * Some requirements have sub-requirements.
       *
       * - `minCount` specifies how many types of sub-requirements needs to be satisfied.
       * - `totalCount` specifies how many courses/credits need to be earned in total.
       */
      readonly totalCount?: number;
    } & A)
  | ({
        readonly fulfilledBy: 'toggleable';
    });


export type CollegeOrMajorRequirement = RequirementCommon &
  RequirementFulfillmentInformation<
    { readonly checker: Checker | readonly Checker[] },
    { 
      readonly multipleOptionsChecker: {
        readonly [optionName: string]: { 
          readonly minCount: number; 
          readonly checkers: readonly Checker[]
        }
      } 
    }
  >;
```

### Test Plan 

Preview site still works.

### Breaking Changes  

None


src/requirements/fetcher.ts
```diff
@@ -138,8 +138,8 @@ const generateSemesterJSONs = async <T extends keyof Course>(
     );
     const courses: AllCourses<T> = {};
     courses[semester] = semesterCourses;
-    const fileRequirementsSrc = 'src/requirements/filtered-' + semester + '-courses.json';
-    const fileFunctionsSrc = 'functions/filtered-' + semester + '-courses.json';
+    const fileRequirementsSrc = `src/requirements/filtered-${semester}-courses.json`;
```
willespencer, line 141: This is fine, but either you or Theresa will probably have a fun merge conflict with this haha.

willespencer (approved):
Good first step 😄! Looks good to merge.

PR #162, opened 2020-11-02
Enforce prettier check for all js files

SamChou19815:
### Summary 

Everyone in course plan seems to code in different style and it causes merge conflicts a lot of times merely due to style change. This PR is the first diff in the series that tries to gradually enforce Prettier style for all files.

Running `npm run format` to fix everything is too disruptive, especially for a lot of requirement code and Vue code. This diff starts from the js files, which only consists of a small minority of the codebase and it's touched often recently. I will decide which ones to format next depending on the dev meeting to see what everyone else is doing.

### Test Plan 

CI

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,18 +2,26 @@
 
 module.exports = {
   root: true,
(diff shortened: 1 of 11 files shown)


willespencer (approved):
Looks good! This is definitely something we should be working towards, so yeah let's talk about it at the dev meeting today 😄

PR #163, opened 2020-11-03
Setup type definition and checker code for toggleable requirements

SamChou19815:
### Summary 

This PR setups the foundation for integrating toggleable requirements into the system. To build confidence of this relatively complex migration, I decided not to add any toggleable requirements into the system in this diff. Instead I will only change the type definition to accommodate for toggleable requirements and change relevant code to make everything type check again.
 
In the next diff, I'm planning to add the first toggleable requirement in #131 and implement relevant frontend changes.

### Test Plan 

`npm run req-gen` creates the same json, which proves that all the non-toggleable requirement checker logic is still the same.

The frontend checking should also be unimpacted:

Screen Shot 2020-11-03 at 11 28 01

### DIFF
--- a/src/requirements/reqs-functions.ts
+++ b/src/requirements/reqs-functions.ts
@@ -2,10 +2,12 @@ import buildRequirementFulfillmentGraph from './requirement-graph-builder';
 import requirementJson from './typed-requirement-json';
 import {
   CourseTaken,
+  EligibleCourses,
   DecoratedCollegeOrMajorRequirement,
   RequirementFulfillment,
   RequirementFulfillmentStatistics,
-  GroupedRequirementFulfillmentReport
+  GroupedRequirementFulfillmentReport,
+  Course
 } from './types';
 
 type RequirementMap = { readonly [code: string]: readonly string[] };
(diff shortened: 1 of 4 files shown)


src/requirements/reqs-functions.ts
```diff
@@ -171,7 +158,74 @@ function computeFulfillmentStatistics<T extends {}>({ requirement, courses: cour
...
+      return computeFulfillmentStatisticsFromCourses(
+        filterAndPartitionCoursesThatFulfillRequirement(coursesTaken, requirement.courses),
+        requirement.fulfilledBy,
+        requirement.operator,
+        requirement.minCount,
+        requirement.totalCount
+      );
+    case 'toggleable': {
+      // Choose the max fulfillment progress as the statistics
+      // TODO: take user choices into account when we are going to integrating that.
```
willespencer, line 210: For this specific part, are you planning on just copying over a lot of Ein's dropdown code?

willespencer (approved):
Looks good to merge! Just left a small question

src/requirements/reqs-functions.ts
SamChou19815: This will be the eventual goal, but in the next PR, I'm just doing to add the toggleable requirement without doing this. Integrating Ein's change also implies it needs to be integrated with the graph, which can be more involved.

PR #137, opened 2020-09-14
Toggleable Requirements 

einc:
Draft so Will can take a first look at it. 
https://github.com/cornell-dti/course-plan/pull/122

### Summary 

This pull request is the first step towards implementing feature Foo

- [] implemented X
- [ ] fixed Y

### Breaking Changes  

- Database schema change (anything that changes Firestore collection structure)
- Other change that could cause problems (Detailed in notes)

PR #164, opened 2020-11-04
Change the bio major chem requirement into toggleable one and fix display

SamChou19815:
### Summary 

This pull request adds the chem toggleable requirement from #137 into the main code. We will use the max-percentage wins rule from #163. Changing it to use user choices will be the focus in another PR.

While integrating the toggleable requirement to it, I also found some display bug that causes it to render `1/undefined toggleable`, so I fixed them in this diff.

### Test Plan 

Using option 1 as expected:
option1

Using option 2 as expected:
option2


willespencer (approved):
Looks good! Good call on renaming the requirement too.

PR #171, opened 2020-11-06
Refactor most of the Dashboard firestore user data converter to user-data.ts

SamChou19815:
### Summary 

There are way too many things happening in Dashboard.vue right now, which makes it hard to reason about what's the data flow. This diff factors out most of the data converters into user-data.ts, so it can results in a shorter Dashboard.vue.

### Test Plan 

Data still loads and displays correctly:

Screen Shot 2020-11-06 at 10 55 19

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -101,6 +101,9 @@ import {
   AppUser,
   AppCourse,
   AppSemester,
+  firestoreCourseToAppCourse,
+  firestoreSemesterToAppSemester,
+  createAppUser,
 } from '@/user-data';
 import { RequirementMap } from '@/requirements/reqs-functions';
 
@@ -205,9 +208,13 @@ export default Vue.extend({
             this.uniqueIncrementer = firestoreUserData.uniqueIncrementer;
             this.loaded = true;
           } else {
-            this.semesters.push(this.createSemester([], this.getCurrentSeason(), this.getCurrentYear()));
-            this.firebaseSems.push({
+            this.semesters.push({
               id: this.currSemID,
+              type: this.getCurrentSeason(),
+              year: this.getCurrentYear(),
+              courses: [],
+            });
+            this.firebaseSems.push({
               type: this.getCurrentSeason(),
               year: this.getCurrentYear(),
               courses: [],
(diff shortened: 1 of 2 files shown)


src/components/Dashboard.vue
```diff
@@ -273,130 +281,6 @@ export default Vue.extend({
...
-      // Get last semester of available course. TODO: Remove when no longer firebase data dependant
-      const lastRoster = course.lastRoster || course.roster;
-
-      // Create course from saved color. Otherwise, create course from subject color group
-      const color = course.color || this.addColor(subject);
-
-      const alerts = { requirement: null, caution: null };
-
-      // Update requirements menu
-      this.updateRequirementsMenu();
```
SamChou19815, line 377: Previously we call this function to update the key for every class. I think it is unnecessary, and makes the refactoring harder. Therefore, I decide to call this function at the end of all updates. e.g. instead of updating requirement whenever we processed one class, we update them after processing all classes.

src/components/Dashboard.vue
```diff
@@ -683,65 +567,26 @@ export default Vue.extend({
...
+      user.transferCourse.forEach(course => {
+        const courseInfo = firestoreCourseToAppCourse(
+          course.course,
+          () => this.incrementID(),
+          (subject) => this.addColor(subject)
+        );
+        transferClasses.push(courseInfo);
+        // ; // TODO for user to pick which req a class goes for
+      });
+      this.currentClasses = transferClasses;
```
SamChou19815, line 587: I decide to leave transfer course processing here for now. I'm not sure how it is used at this point, but just kept to preserve old behavior.

src/user-data.ts
```diff
@@ -131,3 +130,196 @@ export type AppSemester = {
...
+    return [Math.min(...courseCreditRange), Math.max(...courseCreditRange)];
+  }
+  // @ts-ignore
+  return [course.credits, course.credits];
+};
+
+export const firestoreCourseToAppCourse = (
+  course: FirestoreSemesterCourse,
+  incrementID: () => number,
+  addColor: (subject: string) => string
```
SamChou19815, line 157: This is very sad because the old conversion code is not pure functions, but with some tricky side effects tightly coupled with the component, so I think the only way is to make the caller explicitly pass in those side effect functions.

src/user-data.ts
```diff
@@ -131,3 +130,196 @@ export type AppSemester = {
...
+    exam: [],
+    transferCourse: [],
+    tookSwim: data.tookSwim,
+  };
+  if ('exam' in data && data.exam.length > 0) {
+    user.exam = [...data.exam];
+  }
+  if ('class' in data && data.class.length > 0) {
+    user.transferCourse = [...data.class];
+  }
```
SamChou19815, line 301: In general, I really don't like this kind of field existence conditional check. We should ideally write scripts to migrate all the old data, instead of using this kind of undocumented hack.

willespencer (changes requested):
Definitely a good call on factoring those functions out! I am noticing though that there is a console error preventing new courses from being added because createCourse was moved... could you please fix that? 

Otherwise it looks good, just make sure to mention this change to all the devs during standup!

SamChou19815:
> Definitely a good call on factoring those functions out! I am noticing though that there is a console error preventing new courses from being added because createCourse was moved... could you please fix that?
> 
> Otherwise it looks good, just make sure to mention this change to all the devs during standup!

Fixed by temporarily add it back. It should eventually be changed to use emit, but I'm not in the mood of fixing it in this diff. I will probably create a another diff in the future that fixes all of this kind of bad encapsulation code.

willespencer (approved):
Sorry about that, I fixed most if not all of those parent references in my draft, will finish it up ASAP. Looks good now!

PR #172, opened 2020-11-06
[easy] Change some simple Vue components to TS

SamChou19815:
### Summary 

Slightly improve TS percentage, also slightly make future TS migration easier.

### Test Plan 

👀


willespencer (approved):
The TypeScript is taking over 👀

PR #177, opened 2020-11-07
Migrate subrequirement to typescript

SamChou19815:
### Summary 

While waiting on design for toggleable requirements, let me just migrate the file into TS first to reduce the potential PR size for toggleable requirements.

### Test Plan 

It's a type-only diff, and everything type checks.

### DIFF
--- a/src/components/SubRequirement.vue
+++ b/src/components/SubRequirement.vue
@@ -40,24 +40,27 @@
 </template>
 
 
-<script>
-import Vue from 'vue';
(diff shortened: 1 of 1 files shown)


hanywang2 (approved):
Thank you @SamChou19815. How many more files are still not in TS?

SamChou19815:
> Thank you @SamChou19815. How many more files are still not in TS?

probably still at least one third. even for files in TS, there are a lot of `any`s.

PR #180, opened 2020-11-08
Implement toggleable requirement selector UI

SamChou19815:
### Summary 

Implement toggleable requirement UI according to [this design](https://www.figma.com/file/zHChNQWjT9RJ7K20BS28Rk/FINAL-FINAL-FINAL-SPEC?node-id=1468%3A4533). This PR only handles the UI part and stores the choice in a local state. Storing data in DB will be handled in another diff.

### Test Plan 

Test with biological science major in A&S, which is the only current requirement with toggleable requirement.
Screen Shot 2020-11-07 at 19 56 23

### DIFF
--- a/src/components/SubRequirement.vue
+++ b/src/components/SubRequirement.vue
@@ -268,6 +268,7 @@ button.view {
       align-items: center;
       color: $darkPlaceholderGray;
       background: transparent;
+      cursor: pointer;
     }
     &-wrapper {
       display: flex;
@@ -304,6 +305,11 @@ button.view {
         align-items: center;
         color: $lightPlaceholderGray;
         padding-left: 10px;
+        cursor: pointer;
+
+        &:first-child {
+          border-radius: 7px 7px 0px 0px;
+        }
 
         &:last-child {
           border-radius: 0px 0px 7px 7px;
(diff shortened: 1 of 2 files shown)


src/components/SubRequirement.vue
```diff
@@ -82,6 +124,15 @@ export default Vue.extend({
       const type = isCompleted ? 'completed' : 'ongoing';
       this.$emit('toggleDescription', reqIndex, type, subReqIndex);
     },
+    closeMenuIfOpen() {
+      if (this.showFulfillmentOptionsDropdown) {
+        this.showFulfillmentOptionsDropdown = false;
+      }
+    },
```
benjamin-shen, line 131: Isn't this just equivalent to
```
closeMenu() {
  this.showFulfillmentOptionsDropdown = false;
}
```
That may be cleaner?

benjamin-shen (commented):
Looks great, left a minor comment on the code.  
Also on the UI:
Can you make the top borders square? Also can the hover highlight not overhang the bottom of the dropdown (i.e. add border radius to the last child)? The CSS for the Add Semester dropdowns is pretty similar, if you want to reference it.

src/components/SubRequirement.vue
SamChou19815: oops, right

benjamin-shen (commented):
The bottom hover overhang looks fixed now. Can you remove the top border-radius too?
image
Also add cursor: pointer to everything in the dropdown

benjamin-shen (approved):

PR #182, opened 2020-11-08
Migrate more Vue components to typescript

SamChou19815:
### Summary 

Migrate more Vue components to TS.

powertothetypes

### Test Plan 

It's a type only change, and everything type checks.


willespencer (approved):
Power to the types!

PR #183, opened 2020-11-09
Unentangle major and minor name and display

SamChou19815:
### Summary

Right now our requirement related component is very messy. Since Vue allows you to mutate things directly, we are currently mixing mutable data along with immutable data together in a giant array. This makes reasoning about the correctness and data flow very hard.

This PR starts to unentangle mutable and immutable data, starting with major and minor. We have names for major and minor, which should be immutable. We also want to track which major/minor to display. This should be a local state tracked in `data` instead of being together in a mutable array.

To unentangle them, I make Requirement.vue track the index of the major/minor to display and pass this down, instead of tracking display status for each major minor object. I also did other related cleanup to ensure that all the mutable data only appears in `data`, and all immutable data only appears in `props` and `computed`.

### Test Plan

toggle

### DIFF
--- a/src/components/RequirementView.vue
+++ b/src/components/RequirementView.vue
@@ -4,6 +4,8 @@
       :reqIndex="reqIndex"
       :majors="majors"
       :minors="minors"
+      :displayedMajorIndex="displayedMajorIndex"
+      :displayedMinorIndex="displayedMinorIndex"
       :req="req"
       :reqGroupColorMap="reqGroupColorMap"
       :user="user"
@@ -20,7 +22,7 @@
         <div class="separator"></div>
         <div
           v-for="(subReq, id) in req.ongoing"
-          :key="subReq.id">
+          :key="id">
           <subrequirement
             :subReqIndex="id"
             :subReq="subReq"
@@ -47,7 +49,7 @@
 
       <!-- Completed requirements -->
         <div v-if="req.displayCompleted">
-          <div v-for="(subReq, id) in req.completed" :key="subReq.id">
+          <div v-for="(subReq, id) in req.completed" :key="id">
             <div class="separator" v-if="reqIndex < reqs.length - 1 || req.displayDetails"></div>
             <subrequirement
               :subReqIndex="id"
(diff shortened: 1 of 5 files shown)


src/components/RequirementView.vue
```diff
@@ -20,7 +22,7 @@
         <div class="separator"></div>
         <div
           v-for="(subReq, id) in req.ongoing"
-          :key="subReq.id">
+          :key="id">
```
SamChou19815, line 25: The old code is clearly wrong, probably due to some copy-pasta. subReq never contains an id. I also made some similar fixes elsewhere.

src/requirements/types.ts
```diff
@@ -154,7 +154,7 @@ export type SingleMenuRequirement = {
   readonly ongoing: DisplayableRequirementFulfillment[];
   readonly completed: DisplayableRequirementFulfillment[];
   readonly name: string;
-  readonly group: string;
+  readonly group: 'UNIVERSITY' | 'COLLEGE' | 'MAJOR' | 'MINOR';
```
tcho6319, line 157: Do we still need to include 'UNIVERSITY' if we merged university and college requirements?

src/components/RequirementView.vue
```diff
@@ -74,43 +76,37 @@ import RequirementHeader from '@/components/RequirementHeader.vue';
...
-  readonly majorFN: string;
-}
-export type Minor = {
-  display: boolean;
-  readonly minor: string;
-  readonly minorFN: string;
-}
+// reqGroupColorMap maps reqGroup to an array [<hex color for progress bar>, <color for arrow image>]
+const reqGroupColorMap = {
+  UNIVERSITY: ['508197', 'grayblue'],
```
tcho6319, line 87: Could we remove UNIVERSITY from the reqGroupColorMap since we won't show the UNIVERSITY group separately?

tcho6319 (commented):
Looks good, Sam! Just left some comments about whether or not we need to include the UNIVERSITY group.

src/components/RequirementView.vue
SamChou19815: removed it

tcho6319 (approved):
Looks good to me!

PR #184, opened 2020-11-09
Make convertSemesters as pure function

SamChou19815:
### Summary 

The `firestoreCourseToAppCourse` is still doing too many things. After talking with @hanywang2, looks like the predecessor of this function `convertCourse` is doing two things at once: converting data from firestore and converting data from course roster. This is very bad because it makes it very hard to reason about what kind of data we are dealing with, and it caused a tight coupling between data conversion and component logic.

This PR starts to unentangle it. It's obvious from context that `convertSemesters` is only dealing with firebase data, so we can see how we can simplify it. We know in the firestore data that every course has a unique id and a color, so there is no need to call some code to create color and id for it. I believe those code is written for course roster data.

To avoid doing too many things at once in this PR, I still keep the ugly callback structure in `firestoreCourseToAppCourse` (the `incrementID` and `addColor` callback), but make the callback passed into `firestoreCourseToAppCourse` from `convertSemesters` no longer having side effect. After this, `convertCourse` will finally become a pure function!

### Test Plan 

I added some assertions that this is safe. You can load and look at the console that none of the error newly added is thrown.

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -202,12 +202,18 @@ export default Vue.extend({
         .then(doc => {
           if (doc.exists) {
             const firestoreUserData = doc.data() as FirestoreUserData;
-            this.semesters = this.convertSemesters(firestoreUserData.semesters);
+            this.semesters = this.convertSemesters(
+              firestoreUserData.semesters,
+              firestoreUserData.subjectColors,
+            );
+            this.currSemID += this.semesters.length;
+
             this.firebaseSems = firestoreUserData.semesters as FirestoreSemester[];
             this.user = this.parseUserData(firestoreUserData.userData, firestoreUserData.name);
             this.subjectColors = firestoreUserData.subjectColors;
             this.uniqueIncrementer = firestoreUserData.uniqueIncrementer;
             this.loaded = true;
+            this.updateRequirementsMenu();
           } else {
             this.semesters.push({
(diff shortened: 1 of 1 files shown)


src/components/Dashboard.vue
```diff
@@ -243,19 +249,28 @@ export default Vue.extend({
       this.isOpeningRequirements = !this.isOpeningRequirements;
     },
 
-    convertSemesters(firebaseSems: readonly FirestoreSemester[]) {
-      const semesters = firebaseSems.map(firebaseSem => {
-        const appSemester = firestoreSemesterToAppSemester(
+    convertSemesters(
+      firebaseSems: readonly FirestoreSemester[],
+      subjectColors: { readonly [subject: string]: string },
```
hanywang2, line 254: What's the purpose of the refactor where subjectColors is applied to convertSemesters? How i this better than the former solution

hanywang2 (commented):
Question

src/components/Dashboard.vue
SamChou19815: Then convertSemesters is no longer entangled with `addColor`, which has some side effect (doing some firestore operations) that's hard to reason about.

SamChou19815: @hanywang2

SamChou19815:
subsumed in #185.

PR #187, opened 2020-11-09
Eliminate all clickOutside copy-pasta

SamChou19815:
### Summary 

🤦 

### Test Plan 

👀

### DIFF
--- a/src/components/Course.vue
+++ b/src/components/Course.vue
@@ -57,22 +57,9 @@
 <script>
 import Vue from 'vue';
 import CourseMenu from '@/components/Modals/CourseMenu';
+import { clickOutside } from '@/utilities';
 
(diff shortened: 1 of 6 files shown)


benjamin-shen (approved):
lol... nice. lgtm

While you're in `utilities`, do you also want to refactor `getCurrentSeason()` (in `Dashboard` and `NewSemester`)? That function should probably not be `export default`. Or it can be refactored in another PR.

SamChou19815:
> lol... nice. lgtm
> 
> While you're in `utilities`, do you also want to refactor `getCurrentSeason()` (in `Dashboard` and `NewSemester`)? That function should probably not be `export default`. Or it can be refactored in another PR.

I would probably do that in another diff.

PR #189, opened 2020-11-10
Give a unique ID for each requirement

SamChou19815:
### Summary 

By giving each requirement a unique ID, we can test whether two requirement objects are the same much easily. It also gives us a way to avoid using array index as a way to find requirement in the frontend. It will also give me a space-efficient way to store user's choice on toggleable requirements, since I only need to use the requirement ID to tell which requirement is the choice associated with.

### Test Plan 

- Check that every requirement in the generated json has a unique id
- Requirement checking still works in frontend.


SamChou19815:
The PR needs more thoughts. The current implementation might corrupt db during requirement updates

willespencer (approved):

SamChou19815:
Now generating the id from names, category and code instead. It does now starts to require that requirement names are stable.

PR #193, opened 2020-11-11
Convert all semester related components to TypeScript

SamChou19815:
### Summary 

Another type-only change to convert all semester related component to typescript

### Test Plan 

Everything type checks.

### DIFF
--- a/src/components/Modals/NewSemester.vue
+++ b/src/components/Modals/NewSemester.vue
@@ -105,7 +105,8 @@
   </div>
 </template>
 
-<script>
+<script lang="ts">
+import Vue, { PropType } from 'vue';
 import { clickOutside } from '@/utilities';
 
 import fall from '@/assets/images/fallEmoji.svg';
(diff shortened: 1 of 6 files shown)


src/components/Modals/NewSemester.vue
```diff
@@ -105,19 +105,25 @@
...
+<script lang="ts">
+import Vue, { PropType } from 'vue';
 import { clickOutside } from '@/utilities';
 
 import fall from '@/assets/images/fallEmoji.svg';
 import spring from '@/assets/images/springEmoji.svg';
 import winter from '@/assets/images/winterEmoji.svg';
 import summer from '@/assets/images/summerEmoji.svg';
 import {
+  // @ts-ignore
```
willespencer, line 117: Why do these need to be ts-ignored?

willespencer (approved):
Looks good! Just left a question out of curiosity

src/components/Modals/NewSemester.vue
SamChou19815: Tried to add stub types for scss but this destructuring syntax just doesn't work :(

PR #194, opened 2020-11-11
Respect user's choice on toggleable requirement when computing it

SamChou19815:
### Summary 

This PR integrates toggleable user choices with the requirement graph computation. Now the algorithm will respect user choices! The core of the algorithm is implemented a while ago, so this PR is mostly connecting stuff.

Storing user choices in DB will be done in a future PR. In this PR, I only lifted the state of user choices to the Requirement.vue, so that the graph algorithm can use this data.

### Test Plan 

toggleable

### Notes

You can see in the screenshot that all menus will be collapsed when the user changes the choice. This is a problem that I plan to deal it in another diff. 

Cause:
Right now we have to recompute requirement fulfillment completely, when some of the input changes. Unfortunately, the state of whether to display each menu is entangled the fulfillment data, so when nuking the requirement fulfillment during recomputation those display states are also gone.

Plan:
In a future PR, I plan to separate the display data with fulfillment data to fully solve this problem.

### DIFF
--- a/src/components/RequirementView.vue
+++ b/src/components/RequirementView.vue
@@ -28,7 +28,6 @@
             :subReq="subReq"
             :reqIndex="reqIndex"
             :toggleableRequirementChoice="toggleableRequirementChoices[subReq.id]"
-            :changeToggleableRequirementChoice="changeToggleableRequirementChoice"
             :color="reqGroupColorMap[req.group][0]"
             :isCompleted="false"
             @changeToggleableRequirementChoice="changeToggleableRequirementChoice"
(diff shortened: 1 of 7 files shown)


src/components/RequirementView.vue
```diff
@@ -27,8 +27,11 @@
             :subReqIndex="id"
             :subReq="subReq"
             :reqIndex="reqIndex"
+            :toggleableRequirementChoice="toggleableRequirementChoices[subReq.id]"
+            :changeToggleableRequirementChoice="changeToggleableRequirementChoice"
             :color="reqGroupColorMap[req.group][0]"
             :isCompleted="false"
+            @changeToggleableRequirementChoice="changeToggleableRequirementChoice"
```
einc, line 34: Would we also have this for the completed requirements or is it only for the ongoing ones? 
I tried toggling the option when Option 1 is fulfilled to Option 2, but it doesn't update. But this is expected because it's not saved to Firebase yet right?
toggleableReqFilled

einc (approved):

src/components/RequirementView.vue
SamChou19815: It should be fixed now. Forget there are two subrequirement component in this component.

PR #195, opened 2020-11-12
Store whether to display requirement group detail/completed ones using state

SamChou19815:
### Summary 
 
In this notes of #194, I mentioned that storing whether to displaying details/completed requirement inside the array can cause the app to forget about its state during a full requirement recomputation.

This diff removes those fields from `SingleMenuRequirement` type and instead store it inside two separate fields `displayCompleted`/`displayDetails` that are independent of the requirement array. After this diff, recomputing the requirement won't also close all expanded details and completed requirement sections!

### Test Plan 

- Toggle detail/completed still works.
- Now try to change requirement fulfillment option in toggleable requirement will not auto close expanded menus on the left!

### Notes

There are still some work to be done about dealing with toggle details within a subrequirement.

### DIFF
--- a/src/components/RequirementHeader.vue
+++ b/src/components/RequirementHeader.vue
@@ -55,10 +55,10 @@
         <!--View more college requirements -->
         <div class="row top">
           <div class="col-1 p-0" >
-            <button :style="{ 'color': `#${reqGroupColorMap[req.group][0]}` }" class="btn" @click="toggleDetails(reqIndex)">
+            <button :style="{ 'color': `#${reqGroupColorMap[req.group][0]}` }" class="btn" @click="toggleDetails(req.name)">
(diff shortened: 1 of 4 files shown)


willespencer (commented):
Looks good! My only question is if it would be possible/easy for not only the College/Major/Minor menus to stay open when a req is toggled, but also the descriptions under each sub requirement? If not, this is good and still a significant improvement!

SamChou19815:
> Looks good! My only question is if it would be possible/easy for not only the College/Major/Minor menus to stay open when a req is toggled, but also the descriptions under each sub requirement? If not, this is good and still a significant improvement!

That one will require some type refactoring, so I would prefer to do it in another diff.

willespencer (approved):

PR #197, opened 2020-11-13
Only apply completed-ptext to text to unbreak toggleable option UI

SamChou19815:
### Summary 

The old css code incorrectly makes the `opacity: 0.8` css cascading down to the dropdown container, causing a bad transparent UI (see first screenshot in test plan)

This diff fixes it by only applying it to text.

### Test Plan 

Previous:
wrong

Fixed (filled state):
correct-filled

Fixed (unfilled state):
correct-unfilled


willespencer (approved):
Thanks for fixing this!

PR #198, opened 2020-11-14
Store whether to display details/completed in requirement bar as local state

SamChou19815:
### Summary 

This PR makes each requirement view store the display state instead of making them managed at a central place.

### Test Plan 

- Toggling detail and completed section still works.
- Now changing toggleable requirements no longer auto collapse subrequirement details!


willespencer (approved):
Big improvement! Looks good, thank you 😄

PR #199, opened 2020-11-14
Generate course equivalents from AP/transfer credit data

benjamin-shen:
### Summary 

This pull request is the first step towards integrating AP/transfer credit on the requirements side.

- [x] designed new schema, modified data, and wrote new types. moved some existing types from `types.ts` to `ExamCredit.ts` for encapsulation purposes
- [x] wrote a function to translate user-specific exam data to their equivalent courses, using the static exam data

### Test Plan 

To test, you can add the following code to `App.vue` (or another frontend Vue component) and play around with the exam names, scores, college, and major:
```
import getCourseEquivalents from './requirements/data/exams/ExamCredit';

const sampleExams = {
  AP: [
    {
      name: 'Computer Science A',
      score: 0,
    },
    {
      name: 'Chemistry',
      score: 5,
    },
  ],
  IB: [
    {
      name: 'Chemical And Physical Systems',
      score: 5,
    },
  ],
};

console.log(getCourseEquivalents('EN', 'CHEM', sampleExams));
```

Example 1 (cs score is too low, chem exams don't count towards chem major):
image
image

Example 2 (cs score is greater than minimum score):
image
image

Example 3 (chem exams count towards non-chem major):
image
image
Note that there are two courses with the same course id. This double counting is accounted for in Sam's requirement algorithm, not this computation.

Example 4 (AS college has different equivalent courses than EN college):
image
image
Note that the chem course ids are different than in the previous example.

### Notes 

TODOs: 
- Add unit tests
- Convert old schema data to new schema data
- Change the hardcoded 'FA20' to use a utils function currently in Theresa's branch
- Integrate this function in the requirements algorithm

### DIFF
--- a/src/requirements/data/exams/ExamCredit.ts
+++ b/src/requirements/data/exams/ExamCredit.ts
@@ -1,21 +1,6 @@
 /* eslint-disable max-len */
 import { CourseTaken } from '../../types';
 
-// const sampleExams: ExamsTaken = {
-//   AP: [
-//     {
-//       name: 'Computer Science A',
-//       score: 5,
-//     },
-//   ],
-//   IB: [
-//     {
-//       name: 'Chemical And Physical Systems',
-//       score: 5,
-//     },
-//   ],
-// };
-
 export type ExamRequirements = {
   readonly name: string;
   readonly fulfillment: {
@@ -44,16 +29,16 @@ function userDataToCourses(
   const courses: CourseTaken[] = [];
   userExams.forEach(userExam => {
     // match exam to user-taken exam
-    let exam: ExamRequirements | undefined;
-    exams.forEach(e => {
+    const exam = exams.reduce((prev: ExamRequirements | undefined, curr: ExamRequirements) => {
       // check if exam name matches and score is high enough
-      if (e.name.includes(userExam.name) && userExam.score >= e.fulfillment.minimumScore) {
+      if (curr.name.includes(userExam.name) && userExam.score >= curr.fulfillment.minimumScore) {
         // update exam variable if this exam has a higher minimum score
-        if (!exam || exam.fulfillment.minimumScore < e.fulfillment.minimumScore) {
-          exam = e;
+        if (!prev || prev.fulfillment.minimumScore < curr.fulfillment.minimumScore) {
+          return curr;
         }
       }
-    });
+      return prev;
+    }, undefined);
     // generate the equivalent course
     if (exam) {
       const roster = 'FA20'; // TODO this is hardcoded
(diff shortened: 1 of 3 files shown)


src/requirements/data/exams/ExamCredit.ts
```diff
@@ -1,316 +1,443 @@
...
+    let exam: ExamRequirements | undefined;
+    exams.forEach(e => {
+      // check if exam name matches and score is high enough
+      if (e.name.includes(userExam.name) && userExam.score >= e.fulfillment.minimumScore) {
+        // update exam variable if this exam has a higher minimum score
+        if (!exam || exam.fulfillment.minimumScore < e.fulfillment.minimumScore) {
+          exam = e;
+        }
+      }
+    });
```
SamChou19815, line 56: nit: you can do it functionally using `.find`

src/requirements/data/exams/ExamCredit.ts
```diff
@@ -1,316 +1,443 @@
...
+//             collegesApplied: ["EN", "AS"],
+//             majorExcluded: [],
+//             mininmumScore: 6,
+//             courseEquivalents: [['CS 1110']],
+//             requirementEquivalents: []
+//             // varies
+//         },
+//         */
+//   ],
+// };
```
SamChou19815, line 443: what's happening with these commented code

src/requirements/data/exams/ExamCredit.ts
benjamin-shen: How do I use `.find` with the second `if` guard? Like "find exam with max minimumScore compared to other exams"?
I had trouble with `.reduce` earlier but I got it to work now, would that resolve the nit? I totally agree with you btw, I was trying for a while to make it functional.

src/requirements/data/exams/ExamCredit.ts
benjamin-shen: It's the old code (with the old schema). I was planning on adding the correct data in another PR.

src/requirements/data/exams/ExamCredit.ts
```diff
@@ -1,316 +1,443 @@
...
+//       score: 5,
+//     },
+//   ],
+//   IB: [
+//     {
+//       name: 'Chemical And Physical Systems',
+//       score: 5,
+//     },
+//   ],
+// };
```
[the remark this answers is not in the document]
benjamin-shen: Oh yeah, I'll remove it

src/requirements/data/exams/ExamCredit.ts
SamChou19815: oh didn't see the comparison stuff. then the code makes sense to me. not push blocking

src/requirements/data/exams/ExamCredit.ts
willespencer: I'm cool leaving it commented out like this as long as Sam is

willespencer (approved):
Looks good to me! Let me know if/when you want to talk about good next steps!

willespencer (approved):

PR #202, opened 2020-11-18
Factor out alerts from AppCourse object

SamChou19815:
### Summary 

We want to eventually make the data from firestore as the single source of truth. Therefore, we should aim to reduce the different between `FirestoreSemesterCourse` and `AppCourse`. In particular, we should avoid making `AppCourse` contain information that is not in the database (i.e. the stuff that can be computed on demand on the frontend).

The alerts information is one such example. In additional to introducing differences between frontend data type and db data type, this field inside `AppCourse` is also mutated in many different places. It's one of the reason that forces us to use the inefficient deep watch on semester array. Eliminating this mutable field from the object can also help make the flow cleaner.

This refactor diff computes a duplicate course id array instead. Then each course component can use this array to decide whether to display the duplicate warning. While coding for this diff, I also found some unused code, which I will annotate in the PR comments.

### Test Plan 

Try to add two courses that are the same. The warnings still exist.

### DIFF
--- a/src/components/Course.vue
+++ b/src/components/Course.vue
@@ -23,16 +23,7 @@
           <span v-if="!compact && semesterString" class="course-semesters">{{
             semesterString
           }}</span>
-          <!-- <div v-if="!compact && alerts.requirement" class="course-outerWrapper course-tooltip">
-            <div class="course-iconWrapper course-iconWrapper--info">
-              <img class="course-icon course-icon--info" src="../assets/images/info.svg" />
-            </div>
-            <div
-              class="course-tooltiptext course-tooltiptext--info"
-              v-html="requirementString"
-            ></div>
-          </div> -->
-          <div v-if="alerts.caution" class="course-outerWrapper course-tooltip">
+          <div v-if="cautionString" class="course-outerWrapper course-tooltip">
             <div v-if="!compact" class="course-iconWrapper course-iconWrapper--caution">
               <img class="course-icon course-icon--caution" src="../assets/images/caution.svg" />
             </div>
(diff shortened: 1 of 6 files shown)


src/components/Course.vue
```diff
@@ -23,16 +23,7 @@
...
           }}</span>
-          <!-- <div v-if="!compact && alerts.requirement" class="course-outerWrapper course-tooltip">
-            <div class="course-iconWrapper course-iconWrapper--info">
-              <img class="course-icon course-icon--info" src="../assets/images/info.svg" />
-            </div>
-            <div
-              class="course-tooltiptext course-tooltiptext--info"
-              v-html="requirementString"
-            ></div>
-          </div> -->
```
SamChou19815, line 34: This segment of code and the related `loadRequirementsMap` has been dead for quite a while since https://github.com/cornell-dti/course-plan/commit/44c0a27ead2297dd8cc058a147bdab0aed52fb2b

I decided to remove it to make my refactor easier. It's also not a good idea to keep dead code around

src/components/Semester.vue
```diff
@@ -354,20 +354,17 @@ export default Vue.extend({
...
+    // TODO: unused
     buildIncorrectPlacementCautions() {
       if (this.courses) {
         this.courses.forEach(course => {
           if (!course.semesters.includes(this.type))
+            // @ts-ignore
             course.alerts.caution = `Course unavailable in the ${this.type}`;
         });
       }
     },
```
SamChou19815, line 366: This is unused for a while. Not sure what's happening... It also appears to me that the author never considered the possibility that a class can both be duplicated and unavailable.

willespencer (approved):
Looks good, thanks for improving this!

PR #203, opened 2020-11-19
Migrating OnboardingTransfer.vue to TypeScript

SamChou19815:
### Summary 

I'm about to integrate AP/IB exams into requirement graph, but found that @benjamin-shen's changes broke `OnboardingTransfer.vue`. This isn't caught by CI because `OnboardingTransfer.vue` is in JS. This PR turns `OnboardingTransfer.vue` as much as I can.

It turns out that the existing code inside `OnboardingTransfer.vue` is very messy. I'm not sure how to setup type for many code, since the existing code seems to push all kind of stuff with different types into the same object/array. Therefore, I added a lot of `@ts-ignore`. (@rmz38 might have more context on this)

For the `@ts-ignore` due to changed exam data format, I added `// TODO: UNBREAK reqsData FORMAT` above it. I will fix them in the next diff.

### Test Plan 

This is a type-annotation only change. Since it passes CI, it should be good.

### DIFF
--- a/src/components/Modals/OnboardingTransfer.vue
+++ b/src/components/Modals/OnboardingTransfer.vue
@@ -23,7 +23,7 @@
                   class= "onboarding-section"
                   id="college"
                   v-for="(options, index) in displayOptions.exam"
-                  :key = index
+                  :key="index"
                   :style="{ borderColor: options.type.boxBorder }"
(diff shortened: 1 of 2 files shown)


tcho6319 (approved):
LGTM!

PR #205, opened 2020-11-19
Integrate AP/IB exams into requirement graph

SamChou19815:
### Summary 

Feed the equivalent courses into requirement function, and make AP/IB credits count!

### Test Plan 

It now treats the AP CS credit as CS 1110!

Screen Shot 2020-11-19 at 17 00 57
Screen Shot 2020-11-19 at 17 01 01

### DIFF
--- a/src/requirements/data/exams/ExamCredit.ts
+++ b/src/requirements/data/exams/ExamCredit.ts
@@ -1,4 +1,4 @@
-/* eslint-disable max-len */
+import { AppUser } from '@/user-data';
 import { CourseTaken } from '../../types';
 
 export type ExamRequirements = {
@@ -13,7 +13,7 @@ export type ExamRequirements = {
 export type ExamData = Record<'AP' | 'IB', ExamRequirements[]>;
 
 export type ExamTaken = {
-  readonly name: string;
+  readonly subject: string;
   readonly score: number;
 };
 export type ExamsTaken = Record<'AP' | 'IB', ExamTaken[]>;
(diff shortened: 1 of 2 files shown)


src/requirements/data/exams/ExamCredit.ts
```diff
@@ -60,16 +60,34 @@ function userDataToCourses(
...
+  const userExamData: ExamsTaken = { AP: [], IB: [] };
+  user.exam.forEach(exam => {
+    userExamData[exam.type].push(exam);
+  });
+  user.major.forEach(major =>
+    getCourseEquivalentsFromOneMajor(user.college, major, userExamData).forEach(course => {
+      if (!examCourseIDSet.has(course.courseId)) {
+        examCourseIDSet.add(course.courseId);
+        courses.push(course);
+      }
```
benjamin-shen, line 85: Why is the double counting accounted for here? Doesn't the requirements algorithm take care of that?

SamChou19815: Make sure that the same equivalent course doesn't count twice for requirement like academic credits.

benjamin-shen: Hmm, then what about equivalent courses being double counted with non-AP/IB courses? For example, I'm a CS major and I take CHEM 2090. I also take AP CHEM. Would that similarly inflate requirements like academic credits?

SamChou19815: It might, although I'm not sure about the full implication yet.
It appears to me that we can't simply drop the AP exam when we also have taken the same equivalent class, because you might use AP exam to fulfill one stuff and the actual class to fulfill some other stuff.
This is something need further discussion. However, I think it shouldn't affect this diff, since fulfilling academic credits is usually the least thing people worry about.

benjamin-shen (approved):

PR #207, opened 2020-11-20
Requirements bar see all

hanywang2:
### Summary
This PR implements the see all feature on the requirements menu. When a user clicks on the see all for a given requirement with suggested courses, clicking on the button should navigate the users to viewing all courses that are available to satisfy the requirement.

This pull request is the first step towards implementing requirements redesign V2.0

- [x] Added client-side design setup for the menu
- [x] Add support for navigation between show all steps and non-step
- [x] Drag and drop functionality
- [x] Show ALL courses on clicking see all only
- [x] Pagination support or limit for huge fetch requests
- [ ] Bugs with choices from options for a requirement
- [x] Bug with infinite scroll with redundant courses

Screen Shot 2020-11-20 at 6 51 29 PM
Screen Shot 2020-11-20 at 6 51 27 PM

### Test Plan 

Show all should work as expected from designs on Figma


src/user-data.ts
```diff
@@ -162,6 +162,11 @@ export type AppBottomBarCourse = {
   readonly uniqueID: number;
 };
 
+export type ShowAllCourses = {
+  readonly name: string;
+  readonly courses: AppCourse[];
+};
+
```
SamChou19815, line 169: I don't think this type should be in user-data. User data only contains the type of the data in firestore or its converted form. This type seems to be frontend display-only.

src/components/SubRequirement.vue
```diff
@@ -88,9 +88,12 @@
             :subReqCourseId="id"
             :crseInfoObjects="subReqCrseInfoObjects"
             :subReqCourseObjectsNotTakenArray="subReqCourseObjectsNotTakenArray"
+            :subReqCoursesNotTakenArray="subReqCoursesNotTakenArray"
```
SamChou19815, line 91: why are we passing both around?

src/user-data.ts
hanywang2: O I see that makes sense

src/components/SubRequirement.vue
hanywang2: I'll rename the variables. But the subReqCourseObjectsNotTakenArray is fetched from the DB and only return the first four. The functions to fetch those courses could be moved to `src/components/IncompleteSubReqCourse.vue, but I think it might be used in SubRequirement later

SamChou19815 (approved):

PR #210, opened 2020-11-21
Add course from requirements menu

hanywang2:
### Summary

This PR implements the adding of courses from the requirements menu with the semester selector enabled. The functionality to connect the added data with the requirements of the user is planned for another PR. 

Use process:
- Opens a requirement with self-check fulfillment
- Clicks on the + Course button available
- Model with the semester selector is enabled
- User is able to add a course to the selected semester
- If no semester exists, a new one is created and the course is added to that

Screen Shot 2020-11-22 at 4 59 19 PM
Screen Shot 2020-11-22 at 4 59 13 PM
Screen Shot 2020-11-22 at 4 59 01 PM

**Note**
This PR also addressed the dropdown bug when semester a selector so that it expands outside and on top of the model using the relative and absolute positioning. 
Screen Shot 2020-11-22 at 5 08 01 PM

### DIFF
--- a/src/components/AddCourseButton.vue
+++ b/src/components/AddCourseButton.vue
@@ -1,0 +1,105 @@
+<template>
+  <div
+    class="semester-courseWrapper semester-addWrapper"
+    :class="{ 'semester-addWrapper--compact': compact }"
+    @click="onClick"
+    data-intro-group="pageTour"
+    data-step="3"
+    data-intro='<b>Add your course in this semseter!</b><br>
+      <div class = "introjs-bodytext">To start planning your college career, you should try adding a course in your current semester.</div>'
+    data-disable-interaction="1"
+  >
+    <span class="semester-buttonText" :class="{ 'semester-buttonText--compact': compact }">
+      <slot />
+    </span>
+  </div>
+</template>
+
+<script>
+  import Vue, { PropType } from 'vue';
+
+  export default Vue.extend({
+    props: {
+      compact: Boolean,
+    },
+    methods: {
+      onClick() {
+        this.$emit('open-course-modal');
+      }
+    }
+  })
+</script>
+
+<style scoped lang="scss">
+@import '@/assets/scss/_variables.scss';
+
+@mixin hover-button {
+  border-color: #15a6cf;
+  background: rgba(0, 0, 0, 0.03);
+  color: #15a6cf;
+  cursor: pointer;
+}
+
+.semester {
+  width: fit-content;
+  position: relative;
+  border-radius: 11px;
+
+  &--compact {
+    padding: 0.875rem 1.125rem;
+  }
+
+  &-courseWrapper {
+    margin: 0.5rem 0 0.5rem 0;
+  }
+
+  &-addWrapper {
+    margin-top: -5rem;
+    width: 21.375rem;
+    height: 4.625rem;
+    border-radius: 0.5rem;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    border: 2px dashed #d8d8d8;
+    color: #d8d8d8;
+    margin-left: 1.125rem;
+    margin-right: 1.125rem;
+
+    &--compact {
+      margin-top: -1.2rem;
+      width: 10rem;
+      height: 2rem;
+    }
+
+    &:hover,
+    &:active,
+    &:focus {
+      @include hover-button();
+    }
+  }
+
+  &-buttonText {
+    font-weight: 500;
+    font-size: 16px;
+    line-height: 20px;
+
+    &--compact {
+      font-size: 14px;
+      line-height: 17px;
+    }
+  }
+}
+
+@media only screen and (max-width: 878px) {
+  .semester {
+    &-addWrapper {
+      width: 17rem;
+      &--compact {
+        width: 10rem;
+        height: 2rem;
+      }
+    }
+  }
+}
+</style>
--- a/src/components/AddCourseButton.vue
+++ b/src/components/AddCourseButton.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     class="semester-courseWrapper semester-addWrapper"
-    :class="{ 'semester-addWrapper--compact': compact }"
+    :class="{ 'semester-addWrapper--compact': compact, 'my-2 mx-0': shouldClearPadding }"
     @click="onClick"
     data-intro-group="pageTour"
     data-step="3"
@@ -21,10 +21,11 @@
   export default Vue.extend({
     props: {
       compact: Boolean,
+      shouldClearPadding: Boolean,
     },
     methods: {
       onClick() {
-        this.$emit('open-course-modal');
+        this.$emit('click');
       }
     }
   })
--- a/src/components/AddCourseButton.vue
+++ b/src/components/AddCourseButton.vue
@@ -3,14 +3,14 @@
     class="semester-courseWrapper semester-addWrapper"
     :class="{ 'semester-addWrapper--compact': compact, 'my-2 mx-0': shouldClearPadding }"
     @click="onClick"
-    data-intro-group="pageTour"
-    data-step="3"
-    data-intro='<b>Add your course in this semseter!</b><br>
-      <div class = "introjs-bodytext">To start planning your college career, you should try adding a course in your current semester.</div>'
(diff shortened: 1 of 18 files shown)


src/components/Modals/Modal.vue
```diff
@@ -11,14 +11,15 @@
...
         @duplicateSemester="disableButton"
         @close-current-model="closeCourseModal"
         @updateSemProps="updateSemProps"
         @toggle-left-button="toggleLeftButton"
         ref="modalBodyComponent"
-        :season="seasonCourse"
-        :year="yearCourse"
+        :season="season"
+        :year="year"
         :labels="true"
```
hahnbeelee, line 23: For when you are opening the modal from the req bar, you can get rid of the labels by dynamically setting `:labels="false"`!
image

hahnbeelee:
If you choose a class to add from the req bar, change the semester, don't add a class, exit out of the modal, go back into the modal through any of the + Course buttons from the req bar - THEN the semester you chose in the very beginning is still selected. Is this the type of sticky memory that we want? Or do we want it to default back to Fall 2020 once you exit out?

hahnbeelee:
Once you add a class using the requirements bar, is anything supposed to change in the requirements bar? For example, I just added a class by clicking on "+ Course" under the subreq external specialization. Is there anything happening in the backend indicating that I just chose a class to fulfill this requirement s.t. now 1/3 of the requirements for External Specialization are now fulfilled? Or is that happening later?

image

hahnbeelee (changes requested):
Similar to the other PR I reviewed, I kinda just pointed things out that I saw. Your PR descriptions are pretty short so I wasn't 100% what was/wan't supposed to work but I just wanted to be extra careful and point everything out. However, if what you are addressing is solely only the functionality of utilizing the "+ Course" button from the req bar for self-check and creating a new semester if it doesn't already exist or adding it to the existing semester then it looks good!

src/components/Semester.vue
```diff
@@ -99,20 +100,9 @@
...
-          data-intro='<b>Add your course in this semseter!</b><br>
-            <div class = "introjs-bodytext">To start planning your college career, you should try adding a course in your current semester.</div>'
-          data-disable-interaction="1"
-        >
-          <span class="semester-buttonText" :class="{ 'semester-buttonText--compact': compact }">{{
-            buttonString
-          }}</span>
-        </div>
+        <addcoursebutton :compact="compact" @click="openCourseModal">
+          {{ buttonString }}
```
hahnbeelee, line 104: nice factoring out the AddCourseButton! Given the name I feel like you shouldn't have to input the `buttonString` "+ Course" and it should just be hardcoded into the `AddCourseButton.vue` file?

src/components/AddCourseButton.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+  <div
+    class="semester-courseWrapper semester-addWrapper"
+    :class="{ 'semester-addWrapper--compact': compact, 'my-2 mx-0': shouldClearPadding }"
```
hahnbeelee, line 4: where did the classes `my-2` and `mx-0` come from?

src/components/AddCourseButton.vue
```diff
@@ -0,0 +1,106 @@
...
+    @click="onClick"
+    data-intro-group="pageTour"
+    data-step="3"
+    data-intro='<b>Add your course in this semseter!</b><br>
+      <div class = "introjs-bodytext">To start planning your college career, you should try adding a course in your current semester.</div>'
+    data-disable-interaction="1"
```
hahnbeelee, line 10: I'm sus about this walkthrough code being here considering this is a reusable component

src/components/IncompleteSubReqCourse.vue
```diff
@@ -137,6 +148,16 @@ export default Vue.extend({
...
       this.courseObjects = firstFourCourseObjects;
     },
+    onShowAllCourses(courses: AppCourse[]) {
+      this.$emit('onShowAllCourses', {
+        requirementName: this.subReq.requirement.name,
+        subReqCoursesArray: this.subReqCoursesNotTakenArray,
+      });
+    },
+    onAddCourse() {
+      const dashboardRef = this.$parent.$parent.$parent.$parent as any;
```
hahnbeelee, line 158: I remember peeping one of Theresa's PRs and seeing Sam comment "👀👀" about the `$parent.$parent.$parent.$parent` so here's me doing the same 

👀👀

src/components/SemesterView.vue
```diff
@@ -249,12 +251,13 @@ export default Vue.extend({
       return semester;
     },
     addSemester(type: FirestoreSemesterType, year: number) {
+      console.log(year);
```
hahnbeelee, line 254: remove console.log!

src/components/AddCourseButton.vue
hanywang2: Bootstrap https://getbootstrap.com/docs/4.0/utilities/spacing/

src/components/Semester.vue
hanywang2: That's fair cause I don't think it will be anything else

src/components/Modals/Modal.vue
hanywang2: I just routed isCourseModelSelectingSemester to hiding the labels, which is passed anyways. Works you mentioned

hanywang2:
> If you choose a class to add from the req bar, change the semester, don't add a class, exit out of the modal, go back into the modal through any of the + Course buttons from the req bar - THEN the semester you chose in the very beginning is still selected. Is this the type of sticky memory that we want? Or do we want it to default back to Fall 2020 once you exit out?

@hahnbeelee That's a good question. I thought keeping the memory made more sense because it sticks even after you add a course as well. This way it's easier for a user to add courses to a certain semester instead of always changing it away from the current semester

hanywang2:
> Once you add a class using the requirements bar, is anything supposed to change in the requirements bar? For example, I just added a class by clicking on "+ Course" under the subreq external specialization. Is there anything happening in the backend indicating that I just chose a class to fulfill this requirement s.t. now 1/3 of the requirements for External Specialization are now fulfilled? Or is that happening later?
> 
> image

Happening as soon as we merge this branch into master lol. I chose to do them separately cause our PRs in the past have always been too long

src/components/AddCourseButton.vue
hanywang2: Ooooo yea that is such a good point I forgot about that. I can just remove the props when it's the model from the req menu

src/components/IncompleteSubReqCourse.vue
hanywang2: O LOOOL YES I wanted to pre-comment about this but I forgot

hahnbeelee:
> > If you choose a class to add from the req bar, change the semester, don't add a class, exit out of the modal, go back into the modal through any of the + Course buttons from the req bar - THEN the semester you chose in the very beginning is still selected. Is this the type of sticky memory that we want? Or do we want it to default back to Fall 2020 once you exit out?
> 
> @hahnbeelee That's a good question. I thought keeping the memory made more sense because it sticks even after you add a course as well. This way it's easier for a user to add courses to a certain semester instead of always changing it away from the current semester

hmm makes sense tbh

src/components/IncompleteSubReqCourse.vue
hanywang2: Yes, I do acknowledge that this is an absolutely disgusting practice. I would not recommend doing this on any circumstances... BUT... I actually believe it makes the most sense in this situation because the add course model is so integrated with the way we structure the data up to the individual semester component. If we want to use emit, we would have to move the model component to Dashboard.vue and in doing so would migrate half of the code away from semester, and honestly make things a lot more confusing because we would also need to pass back all the courses data from each semester. 

If we were to rebuild CoursePlan, we would've designed entire components tree with this in mind so it wouldn't have to do it. Otherwise, I imagine the refactor would take at least two days, which at the current stage is not that well worth it compared to .$parent.

hahnbeelee: you've convinced me

hahnbeelee (approved):
noicee B) ill make sure to bring up my other comments elsewhere / to the rest of the team

PR #211, opened 2020-11-22
Add nullify check for cautionString

hanywang2:
### Summary

The recent refactor for caution indicators introduced a minor bug that cancelled all fetchRequests to Firebase Functions to retrieve suggested courses in the requirements menu. It was because this.duplicatedCourseCodeList can sometimes be null.

Adding a null checker fixed that

Screen Shot 2020-11-22 at 5 53 54 PM

### DIFF
--- a/src/components/Course.vue
+++ b/src/components/Course.vue
@@ -85,6 +85,7 @@ export default Vue.extend({
   },
   computed: {
     cautionString() {
+      if (this.duplicatedCourseCodeList == null) return null;
       return this.duplicatedCourseCodeList.includes(`${this.subject} ${this.number}`)
         ? 'Duplicate'
         : null;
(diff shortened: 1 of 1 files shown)


SamChou19815 (commented):
Instead of adding this null check here, we should find out why it's null in the first place. Can you check what's the place where we didn't pass in `duplicatedCourseCodeList`, and see whether that makes sense.

hanywang2:
@SamChou19815 it's undefined when it's courses on the requirements menu. I would argue that makes sense

hanywang2:
I'm a bigger fan of a null check as a sanity check, but we can also just pass a prop that specifically cancels the function. Which do you think is better?

SamChou19815:
> I'm a bigger fan of a null check as a sanity check, but we can also just pass a prop that specifically cancels the function. Which do you think is better?

I think it's better to include the null case in PropType so that in the future typescript can prevent this kind of regression

hanywang2:
@SamChou19815 updated it

SamChou19815 (approved):

PR #212, opened 2020-11-23
[Refactor] Remove SVG Dropdown Redundancies

hanywang2:
### Summary

There are 14 SVG files for dropdown/dropups of the same size in different colors. The graphics can all be achieved by having one dropdown with an open fill attribute and rotating it 180deg when it needs to be a dropup.

**Test**
All dropdowns should be the same color and none of them should change in size or spacing 

Screen Shot 2020-11-23 at 3 45 39 PM


hahnbeelee (approved):
Geez lol thank you for doing this. Looks consistent to me!

PR #213, opened 2020-11-28
Update semester and course data functionally

SamChou19815:
### Summary 

This pull request continues the refactoring that aims to make firestore realtime listening possible.

Right now our semester and course data update course is very imperative and the updates happen all over the place. This PR changes the implementation to make it more functional:

- Instead of pushing to the existing array, we create a new array.
- Instead of editing the field of course directly, we create a new object with the updated field.
- etc

Then, we centralize the imperative update to one place only: `Dashboard.vue`. In this way, we no longer need the `deep` watcher on the `semesters` array, and we have a much cleaner and more explicit data flow.

This update will also make a future PR easier. In a PR after this, I plan to make the firestore the source of truth. With the infra setup in this PR, all I need to do is to do the followings:

1. Convert `AppSemester[]` to `FirestoreSemester[]`
2. Update the user data collection document with the new `FirestoreSemester[]`
3. Firestore realtime listener will handle the rest!

### Test Plan 

- Add semester still works
- Edit semester year and season works
- Delete semester still works
- Add course still works
- Edit course color and credits still works
- Delete course still works.

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -38,6 +38,7 @@
         :isMobile="isMobile"
         :currSemID="currSemID"
         @compact-updated="compactVal = $event"
+        @edit-semesters="editSemesters"
         @updateBar="updateBar"
         @close-bar="closeBar"
         @updateRequirementsMenu="updateRequirementsMenu"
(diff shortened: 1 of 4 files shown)


willespencer (approved):
Test plan works and the improvement looks good! Thanks Sam

PR #215, opened 2020-11-30
CompletedSubReqCourse without Reset Functionality

tcho6319:
### Summary
This pull request is the first step towards implementing the CompletedSubReqCourse feature. 

See the Monday Item: [May PR #3: Show Completed Sub-Reqs with the fulfilling courses](https://courseplan.monday.com/boards/433956786/pulses/754580348)

- [x] Replaced `subReqCoursesNotTakenArray`, with `subReqCoursesArray: SubReqCourseSlot[]`
- [x] CompletedSubReqCourse appears for each CompletedSubReqCourse in Ongoing and Completed SubRequirements
- [x] Shows CompletedSubReqCourse for AP/IB transfer credits
- [x] Created a new component `ReqCourse.vue` for the course card in CompletedSubReqCourse
Screen Shot 2020-12-02 at 9 02 29 PM

- [ ] Reset functionality **(in another PR)**
- [ ] Requirements bar updates on drop of a course from requirements bar into schedule  **(in another PR)**

### Test Plan
Screen Shot 2020-11-30 at 5 21 06 PM
Screen Shot 2020-11-30 at 5 20 55 PM
Screen Shot 2020-12-02 at 9 00 36 PM

Everything else in the requirements bar should behave as before.

### Notes
Instead of the `subReqCoursesNotTakenArray: CrseInfo[][]`, we use `subReqCoursesArray: SubReqCourseSlot[]`, that is an array for every subReqCourse slot, whether it be Completed or Incomplete.

The schema of `subReqCoursesArray` follows from its `SubReqCourseSlot[]` type:
```
type CrseInfo = {
  roster: string;
  crseIds: number[];
};

type CompletedSubReqCourseSlot = {
  isCompleted: true;
  courses: readonly CourseTaken[];
};

type IncompleteSubReqCourseSlot = {
  isCompleted: false;
  courses: CrseInfo[];
};

type SubReqCourseSlot = CompletedSubReqCourseSlot | IncompleteSubReqCourseSlot;
```
This change was made so that each specific course slot could be handled separately depending on whether or not it isCompleted or not.


src/components/CompletedSubReqCourse.vue
```diff
@@ -1 +1,184 @@
...
+
+type CompletedSubReqCourseSlot = {
+  isCompleted: true;
+  courses: readonly CourseTaken[];
+};
+
+type IncompleteSubReqCourseSlot = {
+  isCompleted: false;
+  courses: CrseInfo[];
+};
```
hanywang2, line 67: Can we standardize the types in some global file instead? Redefining them is kinda DRY

src/components/CompletedSubReqCourse.vue
```diff
@@ -1 +1,184 @@
...
+  props: {
+    subReq: Object as PropType<DisplayableRequirementFulfillment>,
+    subReqCourseId: Number,
+    crsesTaken: Array as PropType<readonly CourseTaken[]>,
+    semesters: Array as PropType<readonly AppSemester[]>,
+  },
+  data(): Data {
+    return {
+      semesterType: 'Fall',
+      semesterYear: 1234,
```
hanywang2, line 87: If this is default values that will never be seen by the user, can we set semesterType as '' and semesterYear as 0 instead? Think that would make it less confusing to understand

src/components/CompletedSubReqCourse.vue
```diff
@@ -1 +1,184 @@
...
+  width: 100%;
+  background-color: $inactiveGray;
+}
+
+.completed {
+  &-reqCourses {
+    &-course {
+      &-wrapper {
+        margin-top: 4%;
+        margin-bottom: 4%;
```
hanywang2, line 142: I don't believe that percentage values for margins or paddings are great practice. Is there a strong reason they're used here instead of definate values instead? If not, can we switch them?

src/components/CompletedSubReqCourse.vue
```diff
@@ -1 +1,184 @@
...
+      }
+      &-heading {
+        &-wrapper {
+          display: flex;
+          flex-direction: row;
+          justify-content: space-between;
+          margin-top: 2%;
```
hanywang2, line 149: ditto

src/components/CompletedSubReqCourse.vue
```diff
@@ -1 +1,184 @@
...
+        &-wrapper {
+          display: flex;
+          flex-direction: row;
+          justify-content: space-between;
+        }
+        &-semester {
+          font-size: 14px;
+          line-height: 16px;
+          color: $darkLabelGray;
+          margin-top: 2%;
```
hanywang2, line 167: ditto... you get the idea

src/components/IncompleteSubReqCourse.vue
```diff
@@ -36,7 +37,6 @@
       <div v-if="subReq.fulfilledBy === 'self-check'">
         <addcoursebutton :compact="true" :shouldClearPadding="true" @click="onAddCourse" />
       </div>
-      <div class="separator"></div>
```
hanywang2, line 39: thank you. This has been bothering me since forever

src/components/CompletedSubReqCourse.vue
hanywang2: Don't we already have @/requirements/types?

src/components/Requirements.vue
```diff
@@ -329,7 +349,7 @@ export default Vue.extend({
           .then(result => {
             result.data.courses.forEach((course: FirestoreSemesterCourse) => {
               // @ts-ignore [We should resolve this later]
-              const createdCourse = this.$parent.createCourse(course, true);
+              const createdCourse = this.$parent.createCourse(course, true, false);
```
hanywang2, line 352: Since it's only one $parent and we're at it can we swap out this DOM manip with $emit instead?

src/components/SubRequirement.vue
```diff
@@ -120,6 +138,18 @@ type CrseInfo = {
...
+  isCompleted: true;
+  courses: readonly CourseTaken[];
+};
+
+type IncompleteSubReqCourseSlot = {
+  isCompleted: false;
+  courses: CrseInfo[];
+};
+
+type SubReqCourseSlot = CompletedSubReqCourseSlot | IncompleteSubReqCourseSlot;
```
hanywang2, line 151: Ditto on the DRY type definitions

hanywang2 (changes requested):
Great work overall!

src/components/CompletedSubReqCourse.vue
SamChou19815: Related comment: make arrays and field readonly if possible

src/components/CompletedSubReqCourse.vue
```diff
@@ -1 +1,184 @@
...
+      </div>
+      <div class="completed-reqCourses-course-object-wrapper">
+        <course
+          v-bind="courseObject"
+          :courseObj="courseObject"
+          :id="courseObject.subject + courseObject.number"
+          :uniqueID="courseObject.uniqueID"
+          :compact="true"
+          :active="false"
+          :isCompletedReqCourse="true"
```
willespencer, line 28: Smart to just add another parameter! Looks good 😄

willespencer (commented):
Looks pretty good to me! Functionality wise it is great and a huge improvement to the site, just need the other reviewer's comments resolved before merging!

I do think though that your TODO on updating the bar when a course is dragged in is pretty important, and we talked on slack about getting AP courses to show up as completed sub req courses, but both can be resolved in future PRs 😄

src/components/CompletedSubReqCourse.vue
tcho6319: Good idea! I put these types in @/requirements/types. All fields are readonly except for `crseIds` since there seems to be direct manipulation of the `crseIds` field in `getAllCrseInfoFromSemester()`.

src/components/CompletedSubReqCourse.vue
tcho6319: @hanywang2 That's a good point. I made `semesterType` to be of type `FirestoreSemesterType` so it can only be `"Fall" | "Winter" | "Spring" | "Summer"`. I can change `semesterYear` to be 0 and I'll leave a comment for why `semesterType` is `'Fall'`.

src/components/Requirements.vue
tcho6319: @hanywang2 This is a Monday item that I was trying to work on in a previous PR but I didn't know how to get the created course using $emit. Would we pass the created course as a prop to the child components?

src/user-data.ts
```diff
@@ -131,7 +131,8 @@ export type AppCourse = {
   readonly color: string;
   readonly check: boolean;
   uniqueID: number;
-  readonly isReqCourse: boolean;
+  isReqCourse: boolean;
+  isCompletedReqCourse: boolean;
```
SamChou19815, line 135: why these two fields have to be mutable?
I'm also somehow against the idea of adding more field to this course object. This object is already very bloated and it's becoming a do-it-all object. Can we avoid this?

tcho6319: @SamChou19815 For now, I think that we may have to change the value of the fields when the course is dragged from the requirements bar to the schedule (e.g. so that the triple dots appear on the course card). I will have a better answer when I work on my PR for this week but I will change to readonly if my assumption is incorrect.

As for adding `isCompletedReqCourse`, my reasoning was to add a param to style the Course component for this special case. An option that we can explore is to just create a div in `CompletedSubReqCourse.vue` that looks like a compact, isCompletedReqCourse Course card without actually using the Course component.

SamChou19815: It looks like this field is only useful for certain components, so why can't we pass `isReqCourse` and `isCompletedReqCourse` along with an `AppCourse` in those components, instead of merging all of them together.

tcho6319: @SamChou19815 For now, I will make a separate component for a (completed) requirement course (see screenshot below), which will remove the need for an `isCompletedReqCourse` field. I will do this in a following PR.
Screen Shot 2020-12-01 at 8 25 38 PM

I will have to do more exploring, but I might be able to remove the need for an `isReqCourse` field if the new requirement course component also handles a nontaken requirement course (see below). As mentioned in our dev meeting, this might involve a lot of refactoring so this may not be the immediate solution.
Screen Shot 2020-12-01 at 8 27 29 PM

src/components/Requirements.vue
tcho6319: @hanywang2 The ultimate goal is probably not to call `createCourse()` for requirements bar courses at all because it seems to cause issues when trying to set up the firestore user data listener (See @SamChou19815 's note in #217 ). I will try to address this in my future PR for resolving dragging issues.

src/components/CompletedSubReqCourse.vue
```diff
@@ -1 +1,189 @@
...
+  width: 100%;
+  background-color: $inactiveGray;
+}
+
+.completed {
+  &-reqCourses {
+    &-course {
+      &-wrapper {
+        margin-top: 0.6rem;
+        margin-bottom: 0.6rem;
```
hanywang2, line 147: thank you for using rem. great practice

src/components/ReqCourse.vue
```diff
@@ -0,0 +1,152 @@
...
+          </div>
+        </div>
+        <!-- TODO: Revisit if we want this component for Incomplete Req Courses-->
+        <!-- <div v-if="!compact" class="reqcourse-name">{{ name }}</div>
+        <div class="reqcourse-info">
+          <span v-if="!compact" class="reqcourse-credits">{{ creditString }}</span>
+          <span v-if="!compact && semesterString" class="reqcourse-semesters">{{
+            semesterString
+          }}</span>
+        </div> -->
```
hanywang2, line 29: I'm not a big fan of commented code, especially when it's not substantial for any reason. Won't stop you from merging, but you should consider removing it

src/components/ReqCourse.vue
```diff
@@ -0,0 +1,152 @@
...
+    isCompletedReqCourse: Boolean,
+    compact: Boolean,
+  },
+  computed: {
+    borderColorCSSvar() : any {
+      return {
+        '--border-color': `#${this.color}`,
+      };
+    },
+    cssVars() : any {
```
hanywang2, line 53: can we name this a little bit better?

src/components/ReqCourse.vue
```diff
@@ -0,0 +1,152 @@
...
+<style scoped lang="scss">
+@import '@/assets/scss/_variables.scss';
+
+.reqcourse {
+  width: 21.375rem;
+  border-radius: 0.5rem;
+  display: flex;
+  flex-direction: row;
+  background-color: $white;
+  box-shadow: -4px -4px 10px #efefef, 4px 4px 10px #efefef;
```
hanywang2, line 74: This syntax is generally used for when there are two different shadows. Since it's the same color can we achieve it in the same way? Won't spend too much time on this

hanywang2, line 74: Also can we store #efefef as a variable?

hanywang2 (approved):
Have some very simple suggested changes. But overall LGTM! On a separate note, I think this PR is quite huge. In the future, if we can break this down into multiple smaller PR increments with the new features gated off if necessary. Makes it much easier to review

PR #216, opened 2020-12-01
Updating Engineering School and CS Major data 

einc:
### Summary 

- Updated the requirements for the Engineering school and CS major. In future PRs, will continue updating all the majors currently on CP and adding new ones by following [this](https://docs.google.com/document/d/1ALQw0hy8lnhNBujQxG_YiVQ692FAdlM8-bKjCR8PBkE/edit?usp=sharing) and the various items on Monday about req data. 

- Noted and will continue adding comments/errors found for majors on this [doc](https://docs.google.com/document/d/1ALQw0hy8lnhNBujQxG_YiVQ692FAdlM8-bKjCR8PBkE/edit?usp=sharing)

### Test Plan 

Fulfilled all requirements (except for self check) for Eng and CS 
CS in Engineering


src/requirements/data/colleges/en.ts
```diff
@@ -78,8 +78,8 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
       'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements',
     checker: (course: Course): boolean => course.subject === 'ENGRD',
     operator: 'or',
-    fulfilledBy: 'credits',
-    minCount: 6,
+    fulfilledBy: 'courses',
+    minCount: 2,
   },
   {
     name: 'Liberal Studies Distribution',
```
einc, line 85: this requirement gets incorrectly fulfilled with only 3 courses instead of 6
@SamChou19815 are we using totalCount currently? I feel like you fixed this bug a while ago but I couldn't remember. If aren't using totalCount, should I update minCount to 6 for now?

SamChou19815: I think it's never fixed. Here 3 requirements are fighting each other:

1. 18 credits
2. 6 courses
3. 3 categories

Without a better infra and related design, all we can do to fix this bug is to split it into 3 requirements, each only checking only thing at once.

src/requirements/data/majors/cs.ts
```diff
@@ -29,11 +34,28 @@ const csRequirements: readonly CollegeOrMajorRequirement[] = [
...
     minCount: 5,
   },
+  {
+    name: 'CS Electives',
+    description:
+      'Three 4000+ CS electives each at 3 credits. CS 4090, CS 4998, and CS 4998 are NOT allowed.',
+    source:
+      'http://www.cs.cornell.edu/undergrad/rulesandproceduresengineering/choosingyourelectives',
+    checker: (course: Course): boolean => {
+      if (courseMatchesCodeOptions(course, ['CS 4090', 'CS 4998', 'CS 4998', 'CS 4410'])) {
```
SamChou19815, line 44: I think we also need to exclude algo here

src/requirements/data/colleges/en.ts
einc: Sounds good. For now, I split it into 2 requirements that check for credit & courses. 
Perhaps checking for categories would kind of be similar for handling concentrations for majors like Info Sci and something we do in the future ?

SamChou19815: No, I think for categories you should still create another requirement. If you want to handle it in a way similar to concentrations, then you need to create (6 choose 3) dummy engineering colleges :(

einc: Ah got it! I can create a third requirement! 
However, liberal studies requirements are changing (just the categories part, now it's 6 categories) for those entering in 2020 and later. I wasn't sure how we wanted to do changing requirements yet (for this and overall) so I'll bring it up during the dev meeting tomorrow!

SamChou19815 (approved):

PR #217, opened 2020-12-02
[DO-NOT-MERGE] Setup firestore user data listener

SamChou19815:
### Summary 

Setup firestore user data listener. With this refactor, we can finally move to the following data flow:

```
            listeners   <================
             |                         |
             |                         |
            \/                         |
     vue components ====> firestore update actions

```

### Test Plan 

TODO

### Notes 

The following things are broken now due to bad code elsewhere:

- ~~drag and drop courses is no longer recorded in DB~~
  - ~~also broken on master, probably due to some implicit behavior in the D&D library we are using.~~
  - ~~will be fixed in #218~~
- toggle detail in requirement bar auto collapses after a while.
  - root cause identified: subrequirement is indirectly calling Dashboard's incrementID, which changes the user data!
  - talking with @tcho6319 to try to resolve the issue

### DIFF
--- a/src/components/Semester/SemesterView.vue
+++ b/src/components/Semester/SemesterView.vue
@@ -99,8 +99,6 @@
 
 <script lang="ts">
 import Vue, { PropType } from 'vue';
-// @ts-ignore
-import clone from 'clone';
 import Course from '@/components/Course.vue';
 import Semester from '@/components/Semester/Semester.vue';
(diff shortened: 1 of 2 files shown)


SamChou19815:
The merge conflict is not worth solving since we don't have an urgent need to setup listener and it would better be done when we have a centralized data store.

PR #218, opened 2020-12-02
Unbreak auto-saving drag & drop courses

SamChou19815:
### Summary 

#213 breaks the drag and drop saving functionality, because in drag and drop, the courses are updated by the `v-dragula` directive implicitly, so my migration missed that. This diff uses a watcher to watch for the changes made by dragula and correctly report such changes to parent.

### Test Plan 

On master it will no longer remember your d&d choices. It's fixed in this diff.

### DIFF
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -189,7 +189,20 @@ export default Vue.extend({
     semesters: Array as PropType<readonly AppSemester[]>,
     isFirstSem: Boolean,
   },
-
+  watch: {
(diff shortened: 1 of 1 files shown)


tcho6319 (approved):
LGTM. Though I noticed that the bug where the requirements course card does not disappear on drop into the schedule reappeared. The bug is also on master so it was probably from a previous PR that was merged in recently. 
Screen Shot 2020-12-02 at 12 07 03 AM

PR #219, opened 2020-12-03
Store toggleableRequirementChoices in firestore

SamChou19815:
### Summary 

The PR that refactors the data flow (#217) is currently blocked due to some abused functions, so I decided to go with storing toggleableRequirementChoices first without waiting for the refactor to happen.

In this PR, I lifted the state of toggleableRequirementChoices to the root Dashboard.vue, and update the user data collection whenever there is an update from requirement component.

### Test Plan 

Test using A&S's biological science's general chemistry requirement. After changing your choice, reload to ensure it's remembered.

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -20,11 +20,13 @@
           class="dashboard-reqs"
           v-if="loaded && (!isTablet || (isOpeningRequirements && isTablet))"
           :semesters="semesters"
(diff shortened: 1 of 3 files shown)


hahnbeelee (approved):
- The A&S Biological Science Major w/ Gen Chemistry req stores memory when changing between option 1 and 2 and refreshing
- And the req graph changes accordingly 👍

PR #220, opened 2020-12-05
Edit Requirements from Add Modal

hahnbeelee:
### Summary 
- [x] Factor out reqs object from `Requirements.vue` into `Dashboard.vue` s.t. it can be accessible in both the requirements bar and semester view
- [x] Made `getReqsRelatedToCourse` function
- [x] Visual clean up
    - [x] if no requirements or potential requirements fulfilled don't show anything (AKA if array length === 0)
    - [x] if a requirement is selected/unselected send that info somewhere
#### Saving for a future PR
- [ ] Detect binary select - illegal counting
Editing the requirements graph - happening in another PR
- [ ] When the "Add" button is pressed, the requirements that are selected are altered to be fulfilled. - closely related to what people are working on in the requirements bar
    - changing requirements after adding works - but look into how a user could edit a requirement (not done anywhere yet)
    - adding from self check - not implemented yet

### Test Plan
- Go through the process of adding a course
  - Check that the associated requirements show up
  - Edit the requirements and observe the activity 
  - Add the class && look at the console log to see an array of the selected requirements 
 
Overall: Just mess around with adding a course and see how the flow is affected and if it's awkward at all

### Notes 
- The requirement map doesn't actually change when you edit the requirements from the add modal. It's all just frontend.
- My biggest concern is the flow - ~~specifically the BACK button I don't like what it's doing right now, but I'm going to meet with Ein today to talk about it.~~ We talked! changes to the designs/flow will be made.


src/components/Modals/Modal.vue
```diff
@@ -117,11 +124,18 @@ export default Vue.extend({
...
+          this.$refs.modalBodyComponent.next();
+        } else {
+          const dropdown = document.getElementById(`dropdown-${this.semesterID}`);
+          const title = dropdown.value;
 
-        const key = title.substring(0, title.indexOf(':'));
-        this.addCourse();
+          const key = title.substring(0, title.indexOf(':'));
+          const selectedReqs = this.$refs.modalBodyComponent.getSelectedReqs();
+          console.log(selectedReqs);
```
willespencer, line 136: Console log

willespencer (approved):
Looks really good, the frontend is clean 😄 ! Could you remove the console log now that I have done the test plan, and also update Monday with item(s) for the unresolved work for future PRs? Afterwards it will be good to merge!

SamChou19815:
@hahnbeelee ready?

PR #221, opened 2020-12-05
Add Reset Functionality on Req Menu

hanywang2:
### Summary 
This PR added the functionality to reset completed subrequirements from the requirements menu. This is an extension of Theresa's completed-sub-req-course branch. The reset functionality works as follows:

The user is able to click on the `reset` button on a completed subreq.
Screen Shot 2020-12-05 at 1 23 59 AM

The course disappears from the semesterview
Screen Shot 2020-12-05 at 1 24 05 AM

Note that the complete implementation of the reset functionality includes a confirmation model. That should be built in a separate PR.

This pull request is the first step towards implementing the reset feature

- [x] Add general reset functionality
- [ ] Add reset confirmation model (in separate PR)

### Test Plan 

The reset button should work as described. Everything else should run properly as well


src/components/Dashboard.vue
```diff
@@ -635,6 +636,16 @@ export default Vue.extend({
...
+
+    deleteCourseFromSemesters(subject: string, number: string) {
+      const updatedSemesters = this.semesters.map(semester => {
+        const coursesWithoutDeleted = semester.courses.filter(
+          course => course.subject !== subject && course.number !== number
+        );
+        return { ...semester, courses: coursesWithoutDeleted };
+      });
+      this.editSemesters(updatedSemesters);
+    },
```
tcho6319, line 648: For the case of duplicate non AP/IB credits (e.g. having CS 1110 twice in the schedule), maybe we can check the `uniqueID` of the course in addition to the `subject` and `number`. This may involve adding a `courseUniqueID` data field in `CompletedSubReqCourse.vue`.

tcho6319 (commented):
Overall, LGTM! The Reset functionality does not seem to work in the case of AP/IB credits. The functionality for AP/IB would be different since it is not the process of removing the course from the schedule. Would this be handled in a different PR?

willespencer (changes requested):
Agree with @tcho6319's notes on using courseUniqueID and also think we need to consider AP/IB separately (but since this PR does not break anything with AP/IB, that does not need to be addressed with this PR). 

One thing I found when testing though was that resetting one course of a requirement seemed to reset all fulfilled courses in that requirement. Would it be better if it just reset the one corresponding course?

src/components/Dashboard.vue
hanywang2: Yea so I think we'll need support for AB/IB credits first? But it would definitely be for another PR.

src/components/CompletedSubReqCourse.vue
```diff
@@ -126,6 +129,9 @@ export default Vue.extend({
       this.courseSubject = crseTaken.subject;
       this.courseNumber = crseTaken.number;
     },
+    onReset() {
+      this.$emit('deleteCourseFromSemesters', this.courseSubject, this.courseSubject);
+    },
```
tcho6319, line 134: Reset functionality doesn't seem to work because you're missing uniqueID. I think this would involve creating a new data field like `courseUniqueID` and then setting `this.courseUniqueID to be the `uniqueID` of the course that is found in `setCourseAndSemesterForNonTransferCredits()`.

tcho6319 (changes requested):
Thanks for making the changes! I think you're missing courseUniqueID in CompletedSubReqCourse.vue

src/components/CompletedSubReqCourse.vue
hanywang2: Lol I didn't check off the file on GitHub desktop. Sorry about that. Should be pushed now

src/components/CompletedSubReqCourse.vue
```diff
@@ -125,6 +131,15 @@ export default Vue.extend({
       this.color = this.transferCreditColor;
       this.courseSubject = crseTaken.subject;
       this.courseNumber = crseTaken.number;
+      this.courseUniqueId = crseTaken.courseId;
```
tcho6319, line 134: This might have to change when handling AP/IB credits in the separate PR. `courseId` might be `crseId` which is the same for crosslisted courses. Non AP/IB courses sets `this.courseUniqueId = course.uniqueID` so we wouldn't risk deleting the wrong duplicate course. Though, I think we should be wary of the AP/IB case.

tcho6319 (approved):
LGTM! I don't seem to have the problem where all fulfilled courses get reset (as mentioned by @willespencer). Though I added a comment that should be considered when doing Reset for AP/IB credits.

src/components/Dashboard.vue
```diff
@@ -635,6 +636,18 @@ export default Vue.extend({
     naIfEmptyStringArray(arr: readonly string[]) {
       return arr && arr.length !== 0 && arr[0] !== '' ? arr : ['N/A'];
     },
+
+    deleteCourseFromSemesters(subject: string, number: string, uniqueID: number) {
+      const updatedSemesters = this.semesters.map(semester => {
+        const coursesWithoutDeleted = semester.courses.filter(course => {
+          return (
+            course.subject !== subject || course.number !== number || course.uniqueID !== uniqueID
```
willespencer, line 644: Can we just check course.uniqueID here instead of subject and number as well? I think that is the standard across the codebase, and you would not have to emit and pass through as many props

src/components/CompletedSubReqCourse.vue
tcho6319: After discussion with @willespencer : Removing this line might be better because I think resetting AP/IB course in CompletedSubReqCourse would be the equivalent of removing that AP/IB test from the Onboarding component so we shouldn’t really ever set `this.courseUniqueId = crseTaken.courseId`.

willespencer (changes requested):
Almost good! Please just address Theresa and I's nitpicks and make a Monday item for handling AP/IB resetting differently and we should be good to go 😄

src/components/Dashboard.vue
hanywang2: That sounds good

willespencer (approved):
Looks good to go!

PR #222, opened 2020-12-06
Added jest tests on requirement IDs

SamChou19815:
### Summary 

Since we are now storing requirement IDs into the database for toggleable requirements, we need stable IDs for them. This PR adds some test to enforce these desired properties.

The first test uses [Snapshot tests](https://jestjs.io/docs/en/snapshot-testing). It will fail whenever the ID list changes. I added some comments in the test to tell you what to do when this test fails.

The second test is pretty straightforward, it just asserts that the IDs are unique.

### Test Plan 

`npm run test`

The following screenshot shows a test failure when I intentionally changed a requirement name:

Screen Shot 2020-12-05 at 21 25 02


src/requirements/__test__/requirement-data-id.test.ts
```diff
@@ -0,0 +1,46 @@
...
+].sort((a, b) => a.localeCompare(b));
+
+it('ID of the requirements are unchanged.', () => {
+  /**
+   * What should you do when this test fails:
+   *
+   * First, you need to unbreak the test. You only need to run `npm run test -- -u`.
+   *
+   * 1. If course plan has not been launched yet, don't worry.
+   * 2. If course plan is launched.
```
willespencer, line 29: Appreciate this comment, thank you!

willespencer (approved):
Looks good! Quick question though, does this mean that we need to consistently run `npm test` after changing requirements data or does the CI check this?

SamChou19815:
> Looks good! Quick question though, does this mean that we need to consistently run `npm test` after changing requirements data or does the CI check this?

The CI will check this, although it doesn't hurt if you run the test by yourself, so you can have less red-cross in CI logs.

PR #223, opened 2020-12-08
Loading Req Courses

tcho6319:
### Summary

This pull request is the implementation of loading cards for the requirements bar courses in `IncompleteSubReq.vue` using the npm package `skeleton-loader-vue` ([link](https://www.npmjs.com/package/skeleton-loader-vue)). 

Relevant Monday Item: [Loading bar/circle should show while courses are being fetched](https://courseplan.monday.com/boards/433956786/pulses/848282872)

- [x] Installed skeleton-loader-vue
- [x] Display 4 loading courses while waiting for courses on the requirements bar (specifically in `IncompleteSubReq.vue`) to be fetched.

### Test Plan
ezgif com-gif-maker (1)

Loading skeleton should appear while courses for the requirements bar are being fetched. Once courses are fetched, loading skeleton should be replaced by course cards.

### Notes
- Should we create variables that maintain the width and height of our Course cards (both the min and regular versions)?


willespencer (approved):
The loading cards look really good! Should be good to merge 😄 

I do agree with your note, especially because you introduced those numbers in a new place. They could go in the same file we have all of the colors now? Could you make a Monday item for doing that, with maybe some other potential ideas of numbers to make variables (like full sized course cards)?

PR #224, opened 2020-12-09
Generate courses.json with courseId

SamChou19815:
### Summary 

In the dev meeting today, @hahnbeelee makes us realized that we need a way to get course ID from a course without the need to fetch the roster API.

I was originally thinking about replacing the course json's course id list with course code list, but I then realized that it will mess up the AP/IB credits work. In addition, with careful thinking, I still believe courseID is the right thing to record in the generated json, because it matches the equality notation in the graph algorithm.

Therefore, I decided to go with the original approach: attaching course ID to the courses.json. Then @hahnbeelee can use the course ID over there to directly add courses. Maybe @tcho6319 can also use the same file to easily find all the classes that have the same courseID.

### Test Plan 

`npm run courses-gen`


src/requirements/courses-json-generator.ts
```diff
@@ -0,0 +1,26 @@
+import { writeFileSync } from 'fs';
+import filteredAllCourses from './filtered-all-courses';
+
+type CourseJson = {
+  [courseCode: string]: {
+    i: number; // course id
+    t: string; // title
+    r: string; // roster (aka semester)
```
SamChou19815, line 8: following the convention of the original json: use single letter field name to keep the json small

hahnbeelee (approved):
looks good to me!

PR #225, opened 2020-12-09
Hoist requirement data to the root Dashboard.vue

SamChou19815:
### Summary 

I need to implement warnings that are coming from requirement computation, so I need to hoist the data up to the root so that relevant information can be computed higher up and passed down to semesters and classes components.

Since now the requirements are already in the root, there is no need for the `updateRequirementsMenu` hack that only changes the key. We can call the `recomputeRequirements` functions directly instead. As a result, I removed all the `updateRequirementsMenu` and some dead code associated with it. 

### Test Plan 

All the requirements are correctly updated when classes and toggleable requirement changes.

### Notes

I keep this diff small so that @hahnbeelee can easily merge this into her branch once it's in master. There is only one merge conflicts that needs to be fixed:

Screen Shot 2020-12-09 at 13 56 45

which you should remove the line that calls `updateRequirementsMenu`.

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -19,6 +19,7 @@
         <requirements
           class="dashboard-reqs"
           v-if="loaded && (!isTablet || (isOpeningRequirements && isTablet))"
+          :reqs="reqs"
           :semesters="semesters"
           :toggleableRequirementChoices="toggleableRequirementChoices"
           :user="user"
@@ -44,7 +45,6 @@
         @edit-semesters="editSemesters"
         @updateBar="updateBar"
         @close-bar="closeBar"
-        @updateRequirementsMenu="updateRequirementsMenu"
         @increment-semID="incrementSemID"
       />
     </div>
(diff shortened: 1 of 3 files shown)


src/components/Dashboard.vue
```diff
@@ -637,6 +637,77 @@ export default Vue.extend({
...
+        this.getCourseCodesArray(),
+        this.toggleableRequirementChoices,
+        this.user.college,
+        this.user.major,
+        this.user.minor
+      );
+      // Turn result into data readable by requirements menu
+      const singleMenuRequirements = groups.map(group => {
+        const singleMenuRequirement: SingleMenuRequirement = {
+          ongoing: [],
```
rmz38, line 673: I know the type definition exists, but maybe to make it easier for the first time seeing a SingleMenuRequirement, you could add a comment for what should be added to the ongoing[] list and completed[] list?

hahnbeelee (approved):
Looks good to me other than addressing Rose's comment. Thank you for being considerate about my PR - really appreciated.
What I did to test:
Biological Science major & edited the toggle-able requirements and refreshed. Added courses and saw the requirements change on the left. Everything seems to be working as desired.

src/components/Dashboard.vue
SamChou19815: @hahnbeelee maybe you can add some short explanation for me. If I try to add stuff there it might create large merge conflicts for you.

hahnbeelee: ok yeah i can definitely do that!

PR #226, opened 2020-12-11
Factor out CourseCaution component

SamChou19815:
### Summary 

We will add conditional logic on whether there is a single caution or multiple caution soon, so it's better to factor this out to reduce the component complexity in the course component. Also renamed Caution component to SemesterCaution to disambiguate.

### Test Plan 

Works as usual:

Screen Shot 2020-12-11 at 14 10 44


src/components/CourseCaution.vue
```diff
@@ -0,0 +1,125 @@
...
+@import '@/assets/scss/_variables.scss';
+
+.course {
+  &-iconWrapper {
+    font-style: normal;
+    display: flex;
+    margin-left: 0.2rem;
+    align-items: center;
+
+    // TODO: styling for info icon on course card
```
willespencer, line 31: We don't have them on the site right now, but we did have info icons in the design, which I think is what this CSS styling is leftover from. Do we want it in this new component or back in Course?

willespencer (approved):
Looks good! Just left one question for you.

src/components/CourseCaution.vue
SamChou19815: I will leave this here. Not sure how in the future whether we will be reusing the code in this component or start new.

PR #227, opened 2020-12-11
Convert all course related component to typescript

SamChou19815:
### Summary 

Depending on #226. Doing some cleanup.

### Test Plan 

This is a type only change. Everything passes CI.

### DIFF
--- a/src/components/Course.vue
+++ b/src/components/Course.vue
@@ -41,23 +41,24 @@
   </div>
 </template>
 
-<script>
-import Vue from 'vue';
+<script lang="ts">
+import Vue, { PropType } from 'vue';
 import CourseMenu from '@/components/Modals/CourseMenu.vue';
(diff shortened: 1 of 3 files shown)


src/components/Course.vue
```diff
@@ -81,13 +82,13 @@ export default Vue.extend({
     };
   },
   computed: {
-    cautionString() {
+    cautionString(): string | null {
```
hanywang2, line 85: Why not just string and if there is no caution it's just ''?

hanywang2 (approved):
LGTM just got one question

src/components/Course.vue
SamChou19815: That is very sketchy. I believe whether a variable exhibits useful information should be reflected by the type, instead of using a dummy value.

PR #229, opened 2020-12-13
Setup sample double-countable requirement data

SamChou19815:
### Summary 

Attach a field called `allowCourseDoubleCounting` to requirements that allow it. I also setup the rule to ensure that all non-first major, and minors are automatically double countable.

### Test Plan 

Testing on these courses and console log the `illegallyDoubleCountedCourses` list.

Screen Shot 2020-12-13 at 00 25 12

The list is empty as expected.


src/requirements/data/university/index.ts
```diff
@@ -8,6 +8,7 @@ const universityRequirements: readonly CollegeOrMajorRequirement[] = [
       'All incoming freshmen are required to take two credits (two courses) of Physical Education, ' +
       'one credit each semester of the first year on campus.',
     source: 'http://courses.cornell.edu/content.php?catoid=41&navoid=11637',
+    allowCourseDoubleCounting: true,
```
rmz38, line 11: I don't know if it's just because I'm traumatized by CS4620, but I think allowCourseDoubleCounting can be a long name. If you think it belongs, it's fine.

rmz38 (commented):
For the src/requirements/decorated-requirements.json file diff, is there a reason it was deleted and readded? Otherwise it could be an accidental change and need to be reverted.

Otherwise everything looks great!

src/requirements/data/university/index.ts
SamChou19815: I tend to prefer this long name for clarify. Previously people seem to be confused whether the allow double counting is a concept on the requirement or the course, so I think extra long name can serve as a clarification.

SamChou19815:
> For the src/requirements/decorated-requirements.json file diff, is there a reason it was deleted and readded? Otherwise it could be an accidental change and need to be reverted.
> 
> Otherwise everything looks great!

This is intentional. We decided to check in this generated file a long time ago.

src/requirements/reqs-functions.ts
```diff
@@ -303,9 +306,16 @@ export function computeRequirements(
...
-    // TODO: Replace this dummy implementation once we decided how to determine if a requirement is
-    // double-countable. Meanwhile, make it always return true to match the old behavior.
-    allowDoubleCounting: () => true,
+    allowDoubleCounting: requirement => {
+      // All minor requirements are automatically double-countable.
+      if (requirement.sourceType === 'Minor') return true;
+      if (requirement.sourceType === 'Major') {
+        if (majors == null) throw new Error();
+        // If it's not the first major, then it's double countable.
+        if (requirement.sourceSpecificName !== majors[0]) return true;
```
willespencer, line 315: Just want to clarify, does this mean that anything can be double counted by the second major that fulfills a requirement in the first major, or does it also mean that requirements in the second major are double countable with each other?

src/requirements/reqs-functions.ts
```diff
@@ -303,9 +306,16 @@ export function computeRequirements(
...
     },
     getCorrespondingRequirementAndAllRelevantCoursesUnderFulfillmentStrategy: it => it,
-    // TODO: Replace this dummy implementation once we decided how to determine if a requirement is
-    // double-countable. Meanwhile, make it always return true to match the old behavior.
-    allowDoubleCounting: () => true,
+    allowDoubleCounting: requirement => {
+      // All minor requirements are automatically double-countable.
+      if (requirement.sourceType === 'Minor') return true;
+      if (requirement.sourceType === 'Major') {
+        if (majors == null) throw new Error();
```
willespencer, line 313: Do we want anything more descriptive thrown even if it should not happen?

willespencer (commented):
Just left a couple more clarifying question on top of Rose's!

src/requirements/reqs-functions.ts
SamChou19815: short answer: all the non-first majors are treated the same as minors.
then all the minor requirements allow double counting automatically. e.g. if a course A satisfy a minor requirement, it can still be used to satisfy other requirements that doesn't allow double counting.

src/requirements/reqs-functions.ts
SamChou19815: clarified

willespencer (approved):

PR #231, opened 2020-12-13
AP/IB Transfer Data & Improvements

benjamin-shen:
### Summary 

This pull request adds all the preliminary AP/IB data to `ExamCredit.ts`. It also starts to handle more complicated requirements cases.

- [x] Add AP/IB data according to [this spreadsheet](https://docs.google.com/spreadsheets/d/e/2PACX-1vTTSMiw1JqB9b-jyUChWiF_L77F8KalKvPZGRDowfwJAXhlWpGxRiacRRYGUxNuHtH1PAHqfQHty4UR/pubhtml).
- [x] Add special courses for FWS-fulfilling and credit-only courses
- [x] Implement basic function to forfeit ap/ib credit if taking an equivalent course
- [x] Remove much of the equivalent course functionality in the onboarding modal
- [x] Export only the necessary information to the onboarding modal

### Test Plan 

Go into the onboarding modal and test a few AP/IB courses to see if they fulfill the requirements you expect them to fulfill (some examples below). Some special cases (noted at the bottom) are not implemented, but please comment if there's something weird missing from the list.

AP English (A&S) fulfills FWS and total academic credit
image

AP Computer Science A (Eng) fulfills intro programming and total academic credit
image

### Notes 

If you don't have a major, no equivalent courses are computed. This should be fixed (high priority).

We still have some more complicated/special cases not fully handled in this PR:
- composite logic for AP phys I and phys II
- taking a course forfeits credit (partially handled)
- splitting credit from course equivalents (partially handled)
- ap/ib credit should not fulfill a&s credit requirement
- ap/ib equivalent courses (non-credit and credit) should not both show up in credit requirements

We also might want to flesh out the onboarding (/onboarding transfer) modal. For example, computing credits immediately after the user selects the exam and score. Also, removing `equivCourses` from the `FirestoreAPIBExam` schema.


src/requirements/data/checkers-common.ts
```diff
@@ -30,6 +30,13 @@ export const courseMatchesCode = (course: Course, code: string): boolean =>
 export const courseMatchesCodeOptions = (course: Course, codeOptions: readonly string[]): boolean =>
   codeOptions.some(code => ifCodeMatch(`${course.subject} ${course.catalogNbr}`, code));
 
+/**
+ * @param course course object with useful information retrived from Cornell courses API.
+ * @returns true if the course is AP/IB equivalent course or credit
+ */
+export const courseIsAPIB = (course: Course): boolean =>
+  [10, 11].includes(course.crseId) || ['AP', 'IB', 'CREDITS'].includes(course.subject);
```
SamChou19815, line 38: 10, 11 should be put into constants to make it less confusing

src/requirements/data/exams/ExamCredit.ts
```diff
@@ -1,16 +1,17 @@
-import { AppUser } from '../../../user-data';
+import { AppUser, FirestoreAPIBExam } from '../../../user-data';
 import { CourseTaken } from '../../types';
 
 export type ExamRequirements = {
   readonly name: string;
   readonly fulfillment: {
-    readonly courseEquivalents: Record<string, number>;
+    readonly courseEquivalents?: Record<string, number | number[]>;
```
SamChou19815, line 7: why this instead of just `number[]`

benjamin-shen: A large majority of them have a single equivalent course and I didn't want to write the brackets for every exam. It adds some complexity in the function, but makes the data easier to read and modify imo. Do you think I should change it to just `number[]`?

src/requirements/requirement-json-generator.ts
```diff
@@ -5,18 +5,49 @@ import {
...
 } from './types';
 import sourceRequirements from './data';
 import filteredAllCourses from './filtered-all-courses';
 
+/**
+ * Special (synthetic) courses, as used in AP/IB equivalent courses generation.
+ */
+const specialCourses: Course[] = [
+  {
+    subject: 'CREDITS',
```
rmz38, line 18: What special course is CREDITS? Adding a comment to explain or changing the numbers to constants like in Sam's comment could help explain

src/components/Modals/OnboardingTransfer.vue
```diff
@@ -506,18 +483,6 @@ export default Vue.extend({
       };
       this.displayOptions.exam.push(exam);
     },
-    getCourseFromExam(type: 'AP' | 'IB', subject: string) {
```
rmz38, line 509: By deleting this function, what would replace the work it does? It seems like the current transfer page needs this function when the component is mounted, or I might need to change that for my pull request.

rmz38 (changes requested):

src/components/Modals/OnboardingTransfer.vue
benjamin-shen: What exactly does the function do and why does the transfer page need it? If it's to show the equivalent course(s), I would want to have a larger discussion about it. I removed it because the implementation isn't correct/consistent with the logic of `ExamCredit.ts`, and if there's a need for such a function I think it should go in that file.

src/requirements/data/exams/ExamCredit.ts
SamChou19815: I think so, having a consistent data format is ideal, especially when we completely control the data and have the freedom to change them.

src/components/Modals/OnboardingTransfer.vue
rmz38: It gets the course an exam would give credit for and it's only useful for graying out courses in the dropdown if another entered exam would already give credit ex. If you entered AP Chem already then the chemistry IB test will be grayed out. I think it's fine to take it out then

SamChou19815 (approved):

rmz38 (approved):

PR #234, opened 2020-12-14
Accessibility/Trivial fixes 2 

hahnbeelee:
### Summary
- Change `../assets` to `@/assets` so that less things break when files are moved
- added aria to progress bars
- Add alt text to `img`'s

### Test Plan 
- look over changes and make sure that they make sense


src/components/BottomBarTabView.vue
```diff
@@ -30,11 +30,13 @@
           v-if="!seeMoreOpen"
           class="bottombarSeeMoreTab-arrow"
           src="@/assets/images/uparrow-white.svg"
+          alt="up arrow"
```
SamChou19815, line 33: I think it should see something like "collapse see more" instead of just repeating the name of the icon file.

src/components/BottomBarTabView.vue
```diff
@@ -30,11 +30,13 @@
...
         />
         <img
           v-if="seeMoreOpen"
           class="bottombarSeeMoreTab-arrow"
           src="@/assets/images/downarrow-white.svg"
+          alt="down arrow"
```
SamChou19815, line 39: similar problem here and elsewhere

SamChou19815 (approved):
looks good, but some a11y nits

PR #235, opened 2020-12-14
Added description to toggleable requirements + New majors data

einc:
### Summary 

- [x] Added optional description to toggle-able requirements. Once an option ("Biology) is chosen, an optional description of that option is now shown ("Take 1: BIOG...").

It seems like we can implement Concentrations this temporarily. It seems likely that we would run into issues for double-counting requirements and cases where we would need to nest toggleable requirements so this would be a temporary fix. But for launch, majors like Info Sci would at least show concentrations!  

Screen Shot 2020-12-14 at 2 39 50 PM
Screen Shot 2020-12-14 at 2 39 40 PM

- [x] Making progress on adding [these](https://docs.google.com/spreadsheets/d/1yJ71OIA3STd22d4SmUWev3Mr6e_3jWKxF1MwgJNcwG0/edit#gid=0) requirements. Added Biomedical Engineering, Chemical Engineering, Civil Engineering, and DEA majors

- [x] Removed instances of `totalCount` from requirements data as this is not used anymore

### Test Plan 

- [ ] Check toggle-able requirements for Biomedical Engineering Concentrations and DEA Natural Science Requirement. 
- [ ] Look at Biomedical Engineering, Chemical Engineering, Civil Engineering, and DEA majors & double check there are no breaking changes

### Notes 
- [ ] Currently, I put in placeholders for AP/IB credit information and will check in with @benjamin-shen  to see what the best way to add AP/IB credit information is. 
- [ ] Double check with @tcho6319  if any changes need to be made from the Requirements data side with her new PR of how electives are shown (as she showed in standup). The BME and ChemE major had a lot of examples similar to CS Electives.


src/requirements/data/majors/ce.ts
```diff
@@ -0,0 +1,126 @@
+import { CollegeOrMajorRequirement } from '../../types';
+import { includesWithSingleRequirement, includesWithSubRequirements } from '../checkers-common';
+
+const civilRequirements: readonly CollegeOrMajorRequirement[] = [
+  {
+    name: 'Engineering Distribution',
```
SamChou19815, line 6: this requirement overlaps with the engineering distribution requirement, so it's better to mark it as allowDoubleCounting. You might need to merge in latest master

src/requirements/data/majors/bme.ts
```diff
@@ -0,0 +1,142 @@
...
+import { includesWithSingleRequirement, includesWithSubRequirements } from '../checkers-common';
+
+const biomedicalEngineeringRequirements: readonly CollegeOrMajorRequirement[] = [
+  {
+    name: 'Engineering Distriubtions',
+    description:
+      'ENGRD 2202. ENGRD 2020 is recommended and satisfies the Common ' +
+      'Curriculum distribution requirement and also fulfills a required Major course. It is best taken during semester 3 and must be completed before semester 5',
+    source: 'http://courses.cornell.edu/preview_program.php?catoid=41&poid=19808',
+    checker: includesWithSingleRequirement('ENGRD 2202'),
```
SamChou19815, line 11: same with this

src/requirements/data/majors/ce.ts
```diff
@@ -0,0 +1,126 @@
...
+      'BEE 4310 ',
+      'CEE 4330 ',
+      'CEE 4370 ',
+      'CEE 4450 ',
+      'CEE 4530 ',
+      'CEE 4620 ',
+      'CEE 4780 ',
+      'CEE 5970 ',
+      'CEE 6550 ',
+      'CEE 6570 ',
```
SamChou19815, line 115: the spaces don't look right

src/requirements/data/majors/dea.ts
```diff
@@ -0,0 +1,221 @@
...
+      ['DEA 1101'],
+      ['DEA 1110'],
+      ['DEA 1150'],
+      ['DEA 1500'],
+      ['DEA 2030'],
+      [' DEA 2510'],
+      [' DEA 2730'],
+      [' DEA 3590'],
+      [' DEA 4040'],
+      [' DEA 5304']
```
SamChou19815, line 27: 🤔

src/requirements/data/majors/dea.ts
```diff
@@ -0,0 +1,221 @@
...
+      },
+      // TODO: Placeholder for AP Physics
+      'AP Physics score of 5': {
+        description: 'AP Physics score of 5',
+        checker: includesWithSubRequirements(['PHYS 1101']),
+        counting: 'courses',
+        operator: 'and',
+        minCount: 1,
+      },
+    },
```
SamChou19815, line 189: should this be automatically handled by the above requirement?

src/requirements/types.ts
```diff
@@ -72,6 +72,7 @@ type RequirementFulfillmentInformation<T = {}> =
           readonly totalCount?: number;
           readonly counting: 'credits' | 'courses';
           readonly operator: 'and' | 'or';
+          readonly description?: string;
```
SamChou19815, line 75: maybe we should enforce that description is always here?

src/requirements/data/majors/dea.ts
einc: I'll double check with @benjamin-shen 
I wasn't sure how users would choose to accept/not accept their AP credit without toggle-able requirements.

benjamin-shen: I'm not sure if I'm interpreting your question correctly.

Equivalent course credit for AP exams should in general be automatically accepted. I can't think of any scenarios where there are multiple fulfillment strategies for AP/IB equivalent courses. In this case, with toggleable requirements (if I'm understanding correctly), I believe users can choose Physics and their AP credit (or IB credit, or any transfer credit that has an acceptable course id) would automatically apply.

More in-depth explanation: Sam is right, the "Physics" fulfillment option should automatically cover "AP Physics score of 5". When running the requirements json generator, the Physics checker `includesWithSubRequirements(['PHYS 1101', ...])` should add the course id for PHYS 1101 to the `decorated-requirements.json`. Then, if the user takes AP Physics I, `getCourseEquivalentsFromUserExams` should add the equivalent course (with the PHYS 1101 course id) to the courses taken, which then behaves as expected in the requirements graph/algorithm. I'm not exactly sure how the toggleable requirements work but naturally it seems like it would be the same logic: the checker would convert the course code to the course id, which would then match AP Physics' (or whatever exam's) equivalent course's course id.

So @einc I think you can just get rid of all the // TODO: Placeholder's.

einc: That makes sense! Sorry one more question, so if the user chooses not to use AP credit for a course, what would that look right? I think ideally, we would want to give users the option (if possible)? 

For example, a user could have AP CS credit, but choose not to take it and instead take CS 1110.
Would pressing "reset" remove the AP credit (or not because the transfer credit is a part of the requirements precomputation)? Or would the user have to remove the AP credit during onboarding?

benjamin-shen: I don't think any of that is actually (explicitly) implemented. How *should* that work (on a high level)?

Currently, to forfeit the entire ap exam, the user would have to remove it from the onboarding page (because if they're adding it, we should assume they want at least some credit from it?). The reset button doesn't do anything right now but maybe it can remove the exam from the user on Firebase?

If there are cases where we need more control (eg. choose not to use an exam's equivalent course for one requirement, but use it for another requirement), I imagine we would need to update the way exams are stored (maybe with an additional `requirementsFulfilled` field) and use that to change the way course equivalents are generated (maybe change the `CourseTaken` type). Then, AP/IB equivalent courses would be able to act more like regular courses. But I'm pretty sure the granularity for regular courses (the "backend" to Hahnbee's work on the new add course modal) isn't implemented yet. So this is definitely a next-semester solution, unless someone has any better ideas.

SamChou19815 (approved):
looks good.

PR #238, opened 2020-12-24
Add package-lock.json check in CI

SamChou19815:
### Summary 

Looks like the members are still using different versions of npm and is causing `package-lock.json` merge conflicts all the time. For example, the master's `package-lock.json` seems to be produced by npm v7.

It turns out that `npm ci` doesn't enforce `package-lock.json` change on CI, so I added another step to detect `package-lock.json` change using git.

### Test Plan 

The first commit fails the CI check. After updating the `package-lock.json` in the second commit, it no longer fails the check.


SamChou19815:
Closing it since it doesn't seem to work. NPM's behavior is so erratic. Maybe we should switch to yarn

PR #240, opened 2020-12-29
Split firestoreCourseToAppCourse into cornellCourseRosterCourseToAppCourse and firestoreCourseToAppCourse

SamChou19815:
### Summary 

For quite a long time, the `createCourse` function inside `Dashboard.vue` is doing two things at once: converting course data from firestore to `AppCourse` and converting course roster course data into `AppCourse`. When I did some cleanup earlier, such code structure is preserved to keep the diff small. However, now it's the time to cleanup these data, so that we have a better idea what we store in the db.

This diff first splits the type `FirestoreSemesterCourse` into `FirestoreSemesterCourse` and `CornellCourseRosterCourse`, and then splitted their converter functions. As a result of the split, the converter function becomes much cleaner and much safer.

With this refactors, it becomes clear that all the calls to `createCourse` via `this.$parent....createCourse(...)` is for the purpose of converting `CornellCourseRosterCourse` to `AppCourse`, so I renamed the function to reflect that.

### Test Plan 

Check that fetching course information in the requirement bar and add course still works.

### Notes 

In a future diff, I plan to move the `createCourse` function out of the Dashboard component, and make the function and related state available globally.

### DIFF
--- a/src/components/Semester/SemesterView.vue
+++ b/src/components/Semester/SemesterView.vue
@@ -342,7 +342,7 @@ export default Vue.extend({
      * Works in conjunction with addCourse()
      * CHANGE WILL ALTER DATA STRUCTURE
      */
-    toFirebaseCourse(course: AppCourse) {
+    toFirebaseCourse(course: AppCourse): FirestoreSemesterCourse {
       return {
         crseId: course.crseId,
         code: `${course.subject} ${course.number}`,
@@ -359,7 +359,7 @@ export default Vue.extend({
         lastRoster: course.lastRoster,
         color: course.color,
         uniqueID: course.uniqueID,
-      } as FirestoreSemesterCourse;
+      };
     },
     /**
      * Updates semester user data
--- a/src/containers/Dashboard.vue
+++ b/src/containers/Dashboard.vue
@@ -25,7 +25,6 @@
           :user="user"
           :key="requirementsKey"
           :startTour="startTour"
-          @createCourse="createCourse"
           @showTourEndWindow="showTourEnd"
           @on-toggleable-requirement-choices-change="chooseToggleableRequirementOption"
           @deleteCourseFromSemesters="deleteCourseFromSemesters"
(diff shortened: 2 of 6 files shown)


src/components/Semester/SemesterView.vue
```diff
@@ -342,7 +342,7 @@ export default Vue.extend({
      * Works in conjunction with addCourse()
      * CHANGE WILL ALTER DATA STRUCTURE
      */
-    toFirebaseCourse(course: AppCourse) {
+    toFirebaseCourse(course: AppCourse): FirestoreSemesterCourse {
       return {
         crseId: course.crseId,
         code: `${course.subject} ${course.number}`,
```
SamChou19815, line 348: This function proves that the type declaration of `FirestoreSemesterCourse` is correct. Here we are simply adding a type annotation to show that things type check.

Since this function is unchanged, the stuff we store into the database is also unchanged. Therefore, this migration is safe!

src/user-data.ts
```diff
@@ -169,110 +174,81 @@ export type AppToggleableRequirementChoices = Readonly<Record<string, string>>;
...
   addColor: (subject: string) => string
 ): AppCourse => {
-  const uniqueID = course.uniqueID || incrementID();
-
-  const subject = (course.code && course.code.split(' ')[0]) || course.subject;
-  const number = (course.code && course.code.split(' ')[1]) || course.catalogNbr;
-
-  // TODO: same field?
-  const name = course.titleLong || course.name;
+  const uniqueID = incrementID();
```
tcho6319, line 192: Even though `incrementID()` is used for `cornellCourseRosterCourseToAppCourse`, since it is decoupled from `firestoreCourseToAppCourse` now, would we still have to worry about `incrementID()` changing user data as mentioned in #217?

src/containers/Dashboard.vue
```diff
@@ -301,11 +298,14 @@ export default Vue.extend({
     /**
      * Creates a course on frontend with either user or API data
      */
-    createCourse(course: FirestoreSemesterCourse, isRequirementsCourse: boolean): AppCourse {
+    createAppCourseFromCornellRosterCourse(
```
tcho6319, line 301: Would this be the function that would be made global so that we can remove references to `$parent`?

tcho6319 (approved):
Just left some comments with clarification questions but LGTM! Thank you so much Sam

src/user-data.ts
SamChou19815: Yes, since all those functions are still calling the roster conversion one

src/containers/Dashboard.vue
SamChou19815: Yes, this is my intention.

PR #242, opened 2021-01-02
Do not check for subjectColors nullability in Dashboard

SamChou19815:
## Summary

This field always exists since it's initialized to empty object in data function.

## Test Plan

It still type checks, so it's safe.


tcho6319 (approved):
LGTM

PR #243, opened 2021-01-02
Avoid storing firebaseSems in Dashboard.vue data

SamChou19815:
## Summary

This is only useful for end onboarding, so let's just inline the value over there. This refactoring helps to reduce the number of data stored inside Dashboard, so we can have a better idea of what should go into the global store.

## Test Plan

Try to delete your data in dev firestore, and see that the onboarding flow is unaffected.


willespencer (approved):
Tested and looks good!

PR #246, opened 2021-01-03
Avoid storing semester id in AppSemester

SamChou19815:
## Summary

Right now the semester id is only used in two different ways:

- Use as a unique id for the semester for updating/editing, which can be replaced by a combination of year + season(type)
- Use as an array index like thing, which can be used to decide whether the edit color menu appear on the left or right. This can be replaced by an actual array index.

Keeping the id makes semester stuff tightly coupled with the root Dashboard.vue, so it's better to replace it.

## Test Plan

- Test semester creating, editing, deleting works
- Test edit semester and create semester still have correct duplication error warning
- Test edit color menu still appears on the correct side.

### DIFF
--- a/src/components/Course.vue
+++ b/src/components/Course.vue
@@ -29,7 +29,7 @@
     </div>
     <coursemenu
       v-if="menuOpen"
-      :semId="semId"
+      :semesterIndex="semesterIndex"
       :isCompact="compact"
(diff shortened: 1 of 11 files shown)


tcho6319 (approved):
LGTM! Thanks Sam

PR #248, opened 2021-01-05
Split firestore collections

SamChou19815:
### Summary 

Depends on #247.

Right now all the firestore data is stored inside a single collection. This is bad for performance, especially for realtime listening, because if you only care about semester realtime update, but an unrelated change in `uniqueIncrementer` would force a rerender.

This PR splits the collections and provides a script to migrate old data. The script is already successfully run against dev db data, so reviewers can just play with the staging preview site.

### Test Plan 

Try to see editing onboarding data, creating new subject, editing courses, etc still works.

### Breaking Changes  

This is a breaking change. However, a script is provided in functions folder to auto-migrate the old data.


tcho6319 (approved):
LGTM! I think the code will be a lot more organized too if we ever need to update specific firestore collections.

PR #251, opened 2021-01-29
Cleanup Onboarding components (wave 1/n)

SamChou19815:
### Summary 

The current onboarding components are in a mess. Many code are untyped or with `@ts-ignore`...

This diff cleans them up with the following major changes:

1. Collapsed some different css names with exactly the same style into a single class, so that it becomes very clear that some code can be reused (plan for next diff)
2. Removed some unused code in OnboardingBasic.
3. Remove all the id property in vue templates: they are never used...
4. OnboardingBasic.vue is now in TS!
5. Fix all the type errors in OnboardingBasic.

Before this change, the data of OnboardingBasic is a mess. In order to accomodate the incoming user.majors and user.minors being arrays, the code added a lot of if-else condition here and there to deal with it. It leaves us with some code that is almost impossible to statically type. I changed it so that the data is flattened when the component mounts instead of when every time data changes, so we no longer need to deal with two different data format everywhere. As a result, the code is much cleaner and the data is in a consistent format that's easier to write type definitions.

Linter warnings: 47 -> 44

### Test Plan 

Play with the onboarding component to ensure all the data are still saved and there are no visual regressions

### Notes 

After this PR, it becomes clear that we can extract out a common dropdown component for onboarding instead of passing the big display options. I will address it in a future PR.

### DIFF
--- a/src/components/Modals/Onboarding/OnboardingBasic.vue
+++ b/src/components/Modals/Onboarding/OnboardingBasic.vue
@@ -37,38 +37,30 @@
             <div
               class="onboarding-select onboarding-input"
               :class="{ 'onboarding-select--disabled': Object.keys(colleges).length <= 0 }"
-              id="college"
               v-for="(options, index) in displayOptions.college"
               :key="index"
               :style="{ borderColor: options.boxBorder }"
               v-click-outside:[index]="closeCollegeDropdownIfOpen"
             >
               <div
-                class="onboarding-dropdown-placeholder college-wrapper"
+                class="onboarding-dropdown-placeholder college-major-minor-wrapper"
                 @click="showHideCollegeContent(index)"
               >
                 <div
-                  class="onboarding-dropdown-placeholder college-placeholder"
-                  id="college-placeholder"
+                  class="onboarding-dropdown-placeholder college-major-minor-placeholder"
                   :style="{ color: options.placeholderColor }"
                 >
                   {{ options.placeholder }}
                 </div>
                 <div
-                  class="onboarding-dropdown-placeholder college-arrow"
-                  id="college-arrow"
+                  class="onboarding-dropdown-placeholder college-major-minor-arrow"
                   :style="{ borderTopColor: options.arrowColor }"
                 ></div>
               </div>
-              <div
-                class="onboarding-dropdown-content college-content"
-                id="college-content"
-                v-if="options.shown"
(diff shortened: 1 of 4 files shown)


src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -522,43 +413,41 @@ export default Vue.extend({
...
       this.closeDropdownIfOpen('college', i);
     },
-    closeMajorDropdownIfOpen(event, i) {
+    closeMajorDropdownIfOpen(_: unknown, i: number) {
       this.closeDropdownIfOpen('major', i);
     },
-    closeMinorDropdownIfOpen(event, i) {
+    closeMinorDropdownIfOpen(_: unknown, i: number) {
       this.closeDropdownIfOpen('minor', i);
     },
```
tcho6319, line 424: I feel like these functions are repetitive and could be refactored into one function. Could this be handled when factoring out a dropdown component?

tcho6319: Similarly for the functions `selectCollege()`, `selectMajor()`, `selectMinor()` and `removeCollege()`, etc.

src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -461,35 +380,8 @@ export default Vue.extend({
...
-            acronym: option.acronym,
-            fullName: option.placeholder,
-          };
-
-          list.push(obj);
-        }
-      });
-
-      return list;
-    },
```
tcho6319, line 489: Just checking but is this dead code?

tcho6319 (commented):
LGTM Sam! Left a couple quick clarification questions and then should be good for approval. Thank you so much for doing this.

src/components/Modals/Onboarding/OnboardingBasic.vue
SamChou19815: Yes this is dead. The vue template isn't using it

src/components/Modals/Onboarding/OnboardingBasic.vue
SamChou19815: Yes all of them will be handled

tcho6319 (approved):
LGTM!

PR #253, opened 2021-01-30
Cleanup Onboarding components (wave 3/n)

SamChou19815:
### Summary 

This PR cleans up the onboarding basic component by separating the dropdown logic into its own components. Here is the new flow:

The lowest level is `OnboardingSingleDropdown`. This component handles all the color changes when selecting and unselecting items in dropdown.
Above `OnboardingSingleDropdown` is `OnboardingMultiDropdown`. It simply wraps a list of `OnboardingSingleDropdown` and also an add dropdown button. Both `OnboardingSingleDropdown` and `OnboardingMultiDropdown` does not mutate the user choices on their own.
Finally, `OnboardingBasic` contains one `OnboardingSingleDropdown` for college and two `OnboardingMultiDropdown` and major and minor. It is mostly responsible for maintaining the state and communicating with its parent `Onboarding.vue`.

After this refactor, the separation of concern is achieved and the flow is much cleaner.

### Test Plan 

Test that editing college, major, and minor in onboarding still works. Also ensure that data is correctly saved.

### DIFF
--- a/src/components/Modals/Onboarding/Onboarding.vue
+++ b/src/components/Modals/Onboarding/Onboarding.vue
@@ -84,20 +84,16 @@ export default Vue.extend({
   },
   data() {
     return {
-      // TODO: Get real college, major, and minor lists
       currentPage: 1,
-      colleges: {},
-      majors: {},
-      minors: {},
       firstName: this.userData.firstName,
       middleName: this.userData.middleName,
       lastName: this.userData.lastName,
       tookSwim: '',
       placeholderText,
+      collegeDropdown: {},
+      majorDropdowns: [],
+      minorDropdowns: [],
       displayOptions: {
-        college: [],
-        major: [],
-        minor: [],
         exam: [],
         class: [],
       },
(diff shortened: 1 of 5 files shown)


src/components/Modals/Onboarding/Onboarding.vue
```diff
@@ -86,18 +86,15 @@ export default Vue.extend({
     return {
       // TODO: Get real college, major, and minor lists
```
tcho6319, line 87: Is this comment irrelevant now? Can we delete it?

tcho6319 (changes requested):
LGTM and test plan seems to work! Just left a quick comment about a TODO comment in the code and then should be good for approval.

src/components/Modals/Onboarding/Onboarding.vue
SamChou19815: @tcho6319 removed.

SamChou19815:
also #253 and #254 are also ready for review.

tcho6319 (changes requested):
I noticed a discrepancy between the staging site for this PR (left) and master (right). For the left side: When I select "+ add another major" in basic onboarding but don't actually select another major and proceed to the review page, I see "Select one" in my list of majors. This happens for minors as well. The right side (master staging site) shows the intended behavior.

Screen Shot 2021-01-31 at 1 28 30 PM

SamChou19815:
@tcho6319 should be fixed now

tcho6319 (approved):
Looks good!

PR #254, opened 2021-01-31
Cleanup Onboarding components (wave 4/n)

SamChou19815:
### Summary 

Depends on #253.

This diff cleans up the `OnboardingTransfer` component by creating the following components:

- `OnboardingTransferExamPropertyDropdown.vue`, which serves as an editor for exam subject and score,
- `OnboardingTransferSwimming.vue`, which serves as an editor as `tookSwimTest`

Using the above components allow us to move all the display props down to those children components, and greatly simplified the state of the `OnboardingTransfer` component. Now the `OnboardingTransfer` component only serves as a container for states.

### Test Plan 

Check that transfer credits and AP/IB exam data editor are working fine and data are stored.

### DIFF
--- a/src/components/Modals/Onboarding/OnboardingTransfer.vue
+++ b/src/components/Modals/Onboarding/OnboardingTransfer.vue
@@ -1,41 +1,9 @@
 <template>
   <div class="onboarding">
-    <div class="onboarding-section">
-      <div class="onboarding-subHeader">
-        <span class="onboarding-subHeader--font">Cornell Swimming Test </span>
-      </div>
-      <div class="onboarding-inputs onboarding-inputs--name">
-        <div class="onboarding-inputWrapper">
-          <label class="onboarding-label"
-            >Have you taken the Swim Test? (Choose yes if you are a transfer)</label
-          >
-          <div class="onboarding-inputs--radioWrapper">
-            <label class="onboarding-inputs--radio--radioText" for="yes">
-              <input
-                class="onboarding-inputs--radio"
-                type="radio"
-                v-on:click="updateSwimYes"
-                v-model="tookSwimTest"
-                value="yes"
-              />
-              <img class="checkmark" :src="swimYesImage" alt="checkmark" />
-              Yes
-            </label>
-            <label class="onboarding-inputs--radio--radioText" for="no">
-              <input
-                class="onboarding-inputs--radio"
-                type="radio"
-                v-on:click="updateSwimNo"
-                v-model="tookSwimTest"
-                value="no"
-              />
-              <img class="checkmark" :src="swimNoImage" alt="checkmark" />
-              No
-            </label>
-          </div>
-        </div>
-      </div>
-    </div>
+    <onboarding-transfer-swimming
+      :tookSwimTest="tookSwimTest === 'yes'"
+      @update-swim="updateSwim"
+    />
     <div class="onboarding-section">
       <div class="onboarding-subHeader">
         <span class="onboarding-subHeader--font"> Transfer Credits (Optional)</span>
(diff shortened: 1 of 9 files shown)


tcho6319 (changes requested):
Found some bugs when following the test plan:
- Toggling Cornell Swimming Test 
    - Buggy toggling the swimming test. The new choice for the swim test seems to show in the Review page but we can’t click “Finish”.

https://user-images.githubusercontent.com/54298311/106394978-abff1a80-63cd-11eb-96bd-19cae589ee14.mov

- Adding AP/IB Credits
    - Doesn’t seem to work. Once we add the exams and then try to hit “Next >”, we get following console message

image

https://user-images.githubusercontent.com/54298311/106395048-0c8e5780-63ce-11eb-9197-7833f6e60140.mov

- Transferred Course Credits
    - Can’t seem to add equivalent Cornell course (via “Add” button or “+ add another subject” button) anymore
   
image

SamChou19815:
@tcho6319 I think all the above mentioned problems should be fixed now, mostly by removing some code regarding the unused field `equivCourse`.
One problem remains is the transfer class one. This diff doesn't touch the transfer class related logic, and it is already broken in master, due to recent addition of the binary choice modal that requires requirement information. I believe it should be the job of another PR to fix it.

src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -727,42 +288,26 @@ export default Vue.extend({
...
-      if (this.displayOptions.class.length === 0) {
+      this.classes.splice(index, 1);
+      if (this.classes.length === 0) {
         this.addTransfer();
       }
-      this.getCredits();
     },
     addTransfer() {
       // @ts-ignore
       this.displayOptions.class.push({ class: placeholderText, credits: 0 });
```
tcho6319, line 304: Since `displayOptions` is removed from data but not from this function, I think we get an error message in the console when we click "+ add another subject" or "Add" for Transferred Course Credits and a new box does not appear as shown in the screen recording. Do you think this can be resolved in this PR since it doesn't have to do with transfer logic?

tcho6319 (changes requested):
Toggling Cornell Swimming Test works now! I think some of the bugs are still unresolved but show different behavior. I've attached some screen recordings.
- Adding AP/IB Credits
    - Doesn’t seem to work properly (but in a different way from before). When I try to select exams, I get directed to the bottom of the onboarding screen and the exams aren’t selected. They don’t show up in Review page as well.

https://user-images.githubusercontent.com/54298311/106486794-048ef000-6480-11eb-8a9c-8d6ca3158400.mov

- Transferred Course Credits
    - Although we can’t add the equivalent Cornell course so that it shows on the Review page on master, the frontend still shows boxes being added. In the screen recording, left is staging site for this PR and right is master. I left a comment that points to a potential issue for this discrepancy. Do you think we could resolve this bug in this PR since it’s not transfer logic but just the functionality of adding another subject box in Transferred Course Credits?

https://user-images.githubusercontent.com/54298311/106486861-17a1c000-6480-11eb-83d7-fe253317588d.mov

SamChou19815:
@tcho6319  should be fixed now

src/components/Modals/Onboarding/OnboardingTransferExamPropertyDropdown.vue
```diff
@@ -0,0 +1,100 @@
...
+    availableOptions: { type: Array as PropType<readonly unknown[]>, required: true },
+    choice: { type: [String, Number], required: true },
+  },
+  data() {
+    return {
+      shown: false,
+      stopClose: false,
+      boxBorder: '',
+      arrowColor: '',
+      placeholderColor: '',
```
tcho6319, line 60: Can we have the `placeholderColor` be `lightPlaceholderGray` if the placeholder is not "Select one" but is a previously inputted exam? The right (master) panel of the screenshot shows what we would want if the user had just opened up onboarding and saw the AP exams that they'd inputted before. 
Screen Shot 2021-02-01 at 2 14 10 PM

tcho6319 (changes requested):
Thanks for making all the fixes! Some new things popped up since the fixes were made, but I think once these are addressed, the PR will be good for approval.

SamChou19815:
@tcho6319 all fixed

tcho6319 (approved):
LGTM! Thank you so much, Sam.

PR #255, opened 2021-01-31
Add type definitions for scss variables and dragula

SamChou19815:
## Summary

Depends on #254.

Add type definitions so that we can have less ts-ignore.

## Test Plan

Everything still type checks.


tcho6319 (approved):
LGTM!

PR #256, opened 2021-02-02
[cleanup] Eliminate horribly dynamic untyped Modal.vue

SamChou19815:
### Summary 

Currently, `Modal.vue` is very dynamic and impossible to make it managed by TS. It used vue's [dynamic components](https://vuejs.org/v2/api/#is) to dynamically switch between `NewCourse.vue` or `NewSemester.vue` as its inner components. `NewCourse.vue` and `NewSemester.vue` have almost completely different set of props and emits, so as a result, the `Modal.vue` must take all these props and emits and handle all of them. This makes this component very difficult to reason.

This PR changes this sad situation by using [slots](https://vuejs.org/v2/guide/components-slots.html) to achieve the same goal of having a common wrapper component. I made the `FlexibleModal.vue` component as a wrapper component, and the `NewCourseModal` and `NewSemesterModal` in separate.

Then we can do things like

```html
<!-- NewSemesterModal.vue -->
<flexible-modal title="New Semester" props...>
  <new-semester props... />
</flexible-modal>

<!-- NewCourseModal.vue -->
<flexible-modal title="New Course" props...>
  <new-course props... />
</flexible-modal>
```

Then each modal component can have its own set of props and handlers without a lot of tricky if-else.

### Test Plan 

- Adding semester stills works.
- Adding course still works.

Screen Shot 2021-02-02 at 15 01 08
Screen Shot 2021-02-02 at 15 01 18


src/components/Modals/FlexibleModal.vue
```diff
@@ -0,0 +1,117 @@
...
+    display: flex;
+    justify-content: flex-end;
+  }
+
+  &-button {
+    width: 4.75rem;
+    height: 1.8rem;
+    color: #5b676d;
+    border-radius: 3px;
+    border: 1px solid #3d3d3d;
```
tcho6319, line 96: Not sure if this is out of scope of this PR, but we should change the color values to the scss variables. I believe #3d3d3d is $primaryGray. Some of the color values like #5b676d do not seem to have variables so maybe we want to add them.

tcho6319 (approved):
Testing Plan works! Thank you so much for your work, Sam. I didn't know about slots before but I think they make more sense for the way we want to handle modals. I left a comment about suggested changes, though they can be made in another PR if you think they're outside of the scope for this cleanup.

src/components/Modals/FlexibleModal.vue
SamChou19815: changed

PR #257, opened 2021-02-03
[easy][cleanup] Eliminate all unnecessary id from vue templates

SamChou19815:
### Summary 

There are a lot of id declaration in the vue templates. In most cases, they shouldn't be there, because it's usually a sign of DOM manipulation. In this PR I purged almost all of them. In cases where id is used for styling, I replaced that with class names.

There are only 3 left:

1. dropdown-<id>. Sadly, it's used to retrieve values from new course component. I plan to eventually replace it with emits
2. id="app". This is required for setup vue.
3. id="onboarding-transfer-swimming-yes" and id="onboarding-transfer-swimming-no". They are there for the labels.

### Test Plan 

- 404 page is fine
- login page is fine
- dashboard is fine
- modals are fine

### DIFF
--- a/src/components/BottomBar/BottomBarTabView.vue
+++ b/src/components/BottomBar/BottomBarTabView.vue
@@ -8,7 +8,6 @@
       >
         <bottombartab
           v-bind="bottomCourse"
-          :id="index"
           :subject="bottomCourse.subject"
           :number="bottomCourse.number"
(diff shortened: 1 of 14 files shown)


tcho6319 (approved):
Testing Plan seems to work! Thanks Sam.

PR #258, opened 2021-02-04
Convert NewCourse.vue to TypeScript

SamChou19815:
### Summary 

This PR converts `NewCourse.vue` to TypeScript. It doesn't attempt to make it any better. It just adds type and some type casts until things can type check. The goal is to give me some type information for this file, so that I can later more confidently get rid of all the DOM manipulations.

### Test Plan 

This is mostly type-only changes. Now everything still type checks. You can still add courses.

Screen Shot 2021-02-03 at 20 56 59

### DIFF
--- a/src/components/Modals/NewCourse/NewCourse.vue
+++ b/src/components/Modals/NewCourse/NewCourse.vue
@@ -84,8 +84,8 @@
   </div>
 </template>
 
-<script>
-import Vue from 'vue';
+<script lang="ts">
+import Vue, { PropType } from 'vue';
 import coursesJSON from '@/assets/courses/courses.json';
 import EditRequirement from '@/components/Modals/NewCourse/EditRequirement.vue';
 import BinaryButton from '@/components/Modals/NewCourse/BinaryButton.vue';
@@ -94,6 +94,9 @@ import fall from '@/assets/images/fallEmoji.svg';
 import spring from '@/assets/images/springEmoji.svg';
 import winter from '@/assets/images/winterEmoji.svg';
 import summer from '@/assets/images/summerEmoji.svg';
+import { FirestoreSemesterType } from '@/user-data';
+import { SingleMenuRequirement } from '@/requirements/types';
+import { CourseJson } from '@/requirements/courses-json-generator';
 
 Vue.component('editRequirement', EditRequirement);
 Vue.component('binaryButton', BinaryButton);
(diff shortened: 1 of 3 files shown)


src/components/Modals/NewCourse/NewCourse.vue
```diff
@@ -277,38 +282,35 @@ export default Vue.extend({
...
           /* and and make the current item more visible: */
-          addActive(x);
+          addActive(xList);
         } else if (e.keyCode === 13) {
           /* If the ENTER key is pressed, prevent the form from being submitted, */
           e.preventDefault();
           if (currentFocus > -1) {
             /* and simulate a click on the "active" item: */
-            if (x) x[currentFocus].click();
+            if (xList) xList[currentFocus].click();
```
tcho6319, line 305: Observed some weird behavior. Left is PR, right is master. 

My actions: Start typing in class, use keyboard (down arrow key) to move to the first option, press `Enter` key (so that the first option is the text of the search bar), press `Enter` key again, I get this console message that points to this line. I can't seem to replicate this on master, so maybe it has to do with how `xList` is defined in this cleanup?

https://user-images.githubusercontent.com/54298311/106843685-952e2180-6674-11eb-8cad-d0f0ed17335d.mov

tcho6319 (changes requested):
Overall LGTM! Observed some weird behavior that should be checked out/addressed.

src/components/Modals/NewCourse/NewCourse.vue
```diff
@@ -234,8 +239,8 @@ export default Vue.extend({
           }
 
           // Sort both results by title
-          code.sort((first, second) => first.title - second.title);
-          title.sort((first, second) => first.title - second.title);
+          code.sort((first, second) => first.title.localeCompare(second.title));
+          title.sort((first, second) => first.title.localeCompare(second.title));
```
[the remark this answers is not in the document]
SamChou19815: I believe the current code here is simply wrong... You can't subtract two strings...
localeCompare is used to compare strings. Assuming English, then it should result in alphabetical order.

src/components/Modals/NewCourse/NewCourse.vue
SamChou19815: Actually it's probably not worth fixing it in this diff. I already put up the follow up diff that doesn't have this issue: https://github.com/cornell-dti/course-plan/pull/259, and it has a much better implementation. Once that's approved I can merge it together.

tcho6319 (approved):
LGTM!

PR #259, opened 2021-02-04
Cleanup NewCourse component by eliminating DOM manipulation

SamChou19815:
## Summary

Depends on #258.

The NewCourse component is also a mess. It's used in two different places:

- NewCourseModal for supporting the main functionality.
- OnboardingTransfer for selecting equivalent classes.

I first cleaned up the DOM manipulation by replacing that with standard vue state based approach. It turns out it's much simpler to write code in this way.

After the above cleanup, it becomes clear that OnboardingTransfer component is only using the new course component for selecting a course. All the extra requirement functionality are not needed. Therefore, I replaced the use of `newCourse` component to be `course-selector`. This change also unbreaks the functionality of adding transfer classes.

Then we are able to clean up a lot of if-else just to make the new-course component work for onboarding transfer.

## Test Plan

- check that OnboardingTransfer works. Try to add a transfer class.
- check that NewCourseModal works. Try to add a class inside a semester.

Screen Shot 2021-02-03 at 23 22 02
Screen Shot 2021-02-03 at 23 22 13
Screen Shot 2021-02-03 at 23 22 31

### DIFF
--- a/src/components/Modals/NewCourse/CourseSelector.vue
+++ b/src/components/Modals/NewCourse/CourseSelector.vue
@@ -1,0 +1,160 @@
+<template>
+  <div class="autocomplete">
+    <input
+      v-bind:class="['search-box', searchBoxClassName]"
+      ref="dropdownInput"
+      v-model="searchText"
+      :placeholder="placeholder"
+      @keyup.esc="onEscape"
+      @keyup.enter="onEnter"
+      @keyup.up="changeFocus(currentFocus - 1)"
+      @keyup.down="changeFocus(currentFocus + 1)"
+    />
+    <div v-if="matches.length > 0" class="autocomplete-items">
+      <div
+        v-for="(matchingCourse, index) in matches"
+        :key="matchingCourse.id"
+        v-bind:class="['search-result', currentFocus === index ? 'autocomplete-active' : '']"
+        @click="selectCourse(matchingCourse)"
+      >
+        {{ matchingCourse.title }}
+        <input type="hidden" :value="matchingCourse.title" :name="matchingCourse.roster" />
+      </div>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import coursesJSON from '@/assets/courses/courses.json';
+
+import { CourseJson } from '@/requirements/courses-json-generator';
+
+export type MatchingCourseSearchResult = {
+  readonly title: string;
+  readonly roster: string;
+  readonly id: number;
+};
+
+const getMatchingCourses = (
+  searchText: string,
+  courses: CourseJson
+): readonly MatchingCourseSearchResult[] => {
+  // search after value length of 2 to reduce search times of courses
+  if (!searchText || searchText.length < 2) return [];
+  /* code array for results that contain course code and title array for results that contain title */
+  const code: MatchingCourseSearchResult[] = [];
+  const title: MatchingCourseSearchResult[] = [];
+
+  for (const attr in courses) {
+    if (courses[attr]) {
+      const result = {
+        title: `${attr}: ${courses[attr].t}`,
+        roster: courses[attr].r,
+        id: courses[attr].i,
+      };
+      if (attr.toUpperCase().includes(searchText) && attr !== 'lastScanned') {
+        code.push(result);
+      } else if (courses[attr].t && courses[attr].t.toUpperCase().includes(searchText)) {
+        title.push(result);
+      }
+    }
+  }
+
+  // Sort both results by title
+  code.sort((first, second) => first.title.localeCompare(second.title));
+  title.sort((first, second) => first.title.localeCompare(second.title));
+
+  /* prioritize code matches over title matches */
+  // limit the number of results to 10
+  return code.concat(title).slice(0, 10);
+};
+
+export default Vue.extend({
+  props: { searchBoxClassName: String, placeholder: String, autoFocus: Boolean },
+  data() {
+    return {
+      searchText: '',
+      currentFocus: -1,
+    };
+  },
+  computed: {
+    matches(): readonly MatchingCourseSearchResult[] {
+      return getMatchingCourses(this.searchText.toUpperCase(), coursesJSON);
+    },
+  },
+  mounted() {
+    // Activate focus and set input to empty
+    const input = this.dropdownInputRef();
+    if (this.autoFocus) input.focus();
+  },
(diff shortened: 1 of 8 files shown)


src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -0,0 +1,156 @@
...
+  /*the container must be positioned relative:*/
+  position: relative;
+  display: inline-block;
+  width: 100%;
+  margin-top: 0.5rem;
+  padding-bottom: 12px;
+}
+.autocomplete-items {
+  position: absolute;
+  border: 1px solid #d4d4d4;
```
tcho6319, line 130: Can we have scss variables for these colors?

src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -0,0 +1,156 @@
...
+  <div class="autocomplete">
+    <input
+      v-bind:class="['search-box', searchBoxClassName]"
+      ref="dropdownInput"
+      v-model="searchText"
+      :placeholder="placeholder"
+      @keyup.esc="onEscape"
+      @keyup.enter="onEnter"
+      @keyup.up="currentFocus -= 1"
+      @keyup.down="currentFocus += 1"
```
tcho6319, line 11: I think the way we handle incrementing/decrementing `currentFocus` might need to be adjusted so that we have the looping mechanism that we see on the right panel when using the up and down arrow keys. Maybe once `currentFocus` is at the max intended value (I think maybe 9 since we show 10 results?), we don't keep incrementing. I think `addActive()` from the `NewCourse.vue` tried to handle it.

Left is PR and right is master. 
https://user-images.githubusercontent.com/54298311/106916504-419ff000-66d5-11eb-8f3d-c5b1e364d08e.mov

src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -0,0 +1,156 @@
+<template>
+  <div class="autocomplete">
+    <input
+      v-bind:class="['search-box', searchBoxClassName]"
+      ref="dropdownInput"
+      v-model="searchText"
+      :placeholder="placeholder"
+      @keyup.esc="onEscape"
```
tcho6319, line 8: This might be out of scope for this PR, but the escape key doesn't seem to close the course modal. This bug also exists on master so I don't know if we ever addressed it.

tcho6319 (changes requested):
Thank you so much for this cleanup, Sam! I noticed some bugs when following the Test Plan:

- After adding in a subject for transfer credit course and clicking Finish, wipes out Major and Minor.

https://user-images.githubusercontent.com/54298311/106919177-e4f20480-66d7-11eb-89f3-b3c07b4cd898.mov

- In the case where we have added an equivalent Cornell class at a previous point (e.g. “CS 1110” here), can we have the font color be the same as the one that Chemistry (i.e. an IB exam that I’d filled out before) is?
image

I also had a question:
- Would there be any case where the number of credits for an Equivalent Cornell Class would be different depending on the user’s college? For example: Say CS 1110 is a 4 credit class, but I can only have it be applied for 3 credits. Is that possible?

src/components/Modals/NewCourse/CourseSelector.vue
SamChou19815: fixed

src/components/Modals/NewCourse/CourseSelector.vue
SamChou19815: added

tcho6319 (changes requested):
Thanks for all the fixes! Just wanted to bump the font color discrepancy that I mentioned in the earlier review:

- In the case where we have added an equivalent Cornell class at a previous point (e.g. “CS 1110” here), can we have the font color be the same as the one that Chemistry (i.e. an IB exam that I’d filled out before) is?
image

And the question I asked in the earlier review:
- Would there be any case where the number of credits for an Equivalent Cornell Class would be different depending on the user’s college? For example: Say CS 1110 is a 4 credit class, but I can only have it be applied for 3 credits. Is that possible?

Once these are addressed, this should be good to go!

SamChou19815:
> Thanks for all the fixes! Just wanted to bump the font color discrepancy that I mentioned in the earlier review:
> 
> * In the case where we have added an equivalent Cornell class at a previous point (e.g. “CS 1110” here), can we have the font color be the same as the one that Chemistry (i.e. an IB exam that I’d filled out before) is?
>   image
> 
> And the question I asked in the earlier review:
> 
> * Would there be any case where the number of credits for an Equivalent Cornell Class would be different depending on the user’s college? For example: Say CS 1110 is a 4 credit class, but I can only have it be applied for 3 credits. Is that possible?
> 
> Once these are addressed, this should be good to go!

Sorry missed these. The first issue should be fixed now.

For the second question, I don't have a good answer yet. Currently I'm simply maintaining the existing behavior, and we currently don't have any infra on this to support it. If we were to support it in the future, it would probably be like the exam credits json.

src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -327,3 +336,27 @@ export default Vue.extend({
...
+    line-height: 17px;
+    color: $black;
+    width: 100%;
+    border-radius: 3px;
+    padding: 0.5rem;
+    border: 0.5px solid $darkPlaceholderGray;
+    border-radius: 0px;
+    background-color: $white;
+    &::placeholder {
+      color: $black;
```
tcho6319, line 353: Can we make the color `$lightPlaceholderGray` to match how we handle the font color of AP/IB exams that were filled in at a previous time?
Screen Shot 2021-02-05 at 10 00 21 AM

tcho6319 (changes requested):
Thanks for making the changes! Requested a change to match the font color for previously filled in AP/IB exams in Onboarding. We should be good to go after the requested change is addressed!

Re: computing course credits for Equivalent Cornell Courses: sounds good for now. I'll add an item to Notion to make sure we discuss how we want to prioritize this for launch.

src/components/Modals/Onboarding/OnboardingTransfer.vue
SamChou19815: @tcho6319 color fixed in latest commit

tcho6319 (approved):
LGTM! Thank you so much, Sam!

PR #260, opened 2021-02-05
Setup firebase hosting preview

SamChou19815:
### Summary 

Setup the official PR preview solution from firebase so that we don't need to rely on the github pages deploy hack.

### Test Plan 

CI

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -1,6 +1,6 @@
 {
   "projects": {
-    "default": "cornell-courseplan",
+    "default": "cornelldti-courseplan-dev",
     "staging": "cornelldti-courseplan-dev",
     "production": "cornell-courseplan"
   }
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -8,18 +8,16 @@ import Page404 from '@/containers/404.vue';
 
 Vue.use(Router);
 
-const baseURL = process.env.BASE_URL;
-
 const router = new Router({
   mode: 'history',
   routes: [
     {
-      path: `${baseURL}/login`,
+      path: '/login',
       name: 'Login',
       component: Login,
     },
     {
-      path: `${baseURL}/`,
+      path: '/',
       name: 'Dashboard',
       component: Dashboard,
       meta: {
@@ -27,7 +25,7 @@ const router = new Router({
       },
     },
     {
-      path: `${baseURL}/*`,
+      path: '/*',
       name: '404',
       component: Page404,
     },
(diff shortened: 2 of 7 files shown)


src/router/index.ts
```diff
@@ -8,26 +8,24 @@ import Page404 from '@/containers/404.vue';
...
       meta: {
         requiresAuth: true,
       },
     },
     {
-      path: `${baseURL}/*`,
+      path: '/*',
       name: '404',
       component: Page404,
     },
```
SamChou19815, line 31: the base_url is used as a hack for github pages deployment. Now it's no longer needed.

vue.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  publicPath: process.env.PUBLIC_URL || '',
+  publicPath: '',
```
tcho6319, line 2: Just checking if my understanding is correct but we don't need `process.env.PUBLIC_URL` anymore because we're not relying on env variables to do github page deployments for PRs right?

src/router/index.ts
tcho6319: What was the downside of using `baseURL` (i.e. env variables)?

tcho6319 (commented):
Thanks for setting this up Sam! This is new to me so I just wanted to confirm/clarify some things:
1. By using Firebase hosting preview, we've replaced github pages for PR and master staging sites right?
2. It seems the deployment bot will automatically create a preview URL. So devs will not have to do anything special when making PRs?
3. To double check: This wouldn't change anything about the process in deploying to prod right?

.firebaserc
```diff
@@ -1,6 +1,6 @@
 {
   "projects": {
-    "default": "cornell-courseplan",
+    "default": "cornelldti-courseplan-dev",
```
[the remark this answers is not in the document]
SamChou19815: It's the prod instance name. It's better not to make the prod as default so you don't accidentally deploy to prod

vue.config.js
SamChou19815: Yes that's right.

src/router/index.ts
SamChou19815: No downside. It's simply unnecessary now.

SamChou19815:
> Thanks for setting this up Sam! This is new to me so I just wanted to confirm/clarify some things:
> 
> 1. By using Firebase hosting preview, we've replaced github pages for PR and dev staging sites right?
> 
> 2. It seems the deployment bot will automatically create a preview URL. So devs will not have to do anything special when making PRs?
> 
> 3. To double check: This wouldn't change anything about the process in deploying to prod right?

Yes all of these are correct. I will delete the gh-pages when its merged

tcho6319 (approved):
LGTM! Excited to use this instead of github pages :)

PR #262, opened 2021-02-07
Eliminate selected state since it can be derived from selectedCourse

SamChou19815:
### Summary 

It's always true that `selected === (selectedCourse != null)`, so let's just get rid of `selected`.

Also did some misc cleanup like removing the need for passing placeholder text around.

### Test Plan 

Requirement editor still shows up after a course is selected.

Screen Shot 2021-02-07 at 12 05 56

### DIFF
--- a/src/components/Modals/NewCourse/NewCourse.vue
+++ b/src/components/Modals/NewCourse/NewCourse.vue
@@ -1,16 +1,15 @@
 <template>
   <div class="newCourse">
-    <div class="newCourse-text">{{ text }}</div>
-    <!-- TODO: for some reason this breaks the dropdown <div v-if="selected" class="newCourse-name newCourse-requirements-container">{{ selectedCourse }}</div> -->
+    <div class="newCourse-text">Search Course Roster</div>
     <course-selector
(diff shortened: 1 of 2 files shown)


src/components/Modals/NewCourse/NewCourse.vue
```diff
@@ -125,13 +116,8 @@ export default Vue.extend({
     };
   },
   computed: {
-    text(): string {
-      return 'Search Course Roster';
-    },
```
tcho6319, line 130: What's the benefit of not having this text be a computed property?

tcho6319 (commented):
LGTM! I just left a quick comment but once that's addressed, should be good for approval.

src/components/Modals/NewCourse/NewCourse.vue
SamChou19815: It just bloats the component with computed properties. Usually it's only needed if you need some translations or conditional text

tcho6319 (approved):
LGTM! Thanks Sam

PR #263, opened 2021-02-07
Rename EditRequirement to EditSingleRequirement

SamChou19815:
Depends on #262.

### Summary 

I'm about to introduce a RequirementEditor that supports all the requirement edits, so I renamed this to disambiguate.

### Test Plan 

NewCourse modal still shows the requirement editor correctly.


tcho6319 (approved):
LGTM! So to confirm: Will this Requirement Editor be handling the case where a user wants to edit an existing course, specifically which requirements the course is applied to? For example: When I added Course A, I applied it to Req 1. But now, I want to edit Course A so that it applies to Req 2 not Req 1.

SamChou19815:
> LGTM! So to confirm: Will this Requirement Editor be handling the case where a user wants to edit an existing course, specifically which requirements the course is applied to? For example: When I added Course A, I applied it to Req 1. But now, I want to edit Course A so that it applies to Req 2 not Req 1.

I plan to make it more general purposed. I will make it take in a bunch of requirements and report what user selected. It doesn't contain any requirement logic, so it should be able to apply in all cases.

PR #265, opened 2021-02-09
Rename the confusing operator field to subRequirementProgress

### LINKED ISSUE #138: Existing requirement computation issues

SamChou19815:
### Summary 

Close #138. This PR addresses the first and final unresolved issue in that github issue. The `operator: 'and' | 'or'` is very confusingly named and every time it took me a while to understand what they mean. This PR changes them to

`subRequirementProgress: 'every-course-needed' | 'any-can-count'`

I believe this wording can make the code reader know just be reading that this field tells how subrequirement progress stat is computed.

### Test Plan 

All requirements are computed as usual:

Screen Shot 2021-02-08 at 21 24 15


tcho6319 (approved):
LGTM. This naming seems much more intuitive. Thanks Sam!

PR #267, opened 2021-02-09
Inline NewCourse.vue into NewCourseModal.vue

SamChou19815:
### Summary 

This PR inlines NewCourse into NewCourseModal. This can help us avoid some awkward component communication via ref and also allows us to find some duplicated states that can be easily derived. e.g.

- left button text is always cancel when the course is not selected, always add when the course is selected.
- right button text is always next is edit mode, and add when not in edit mode.
- right button is only disabled when a course is not selected.
- goBack state is unused

### Test Plan 

Local:
local
Master:
master

### Notes

In a future PR, I plan to fix the correctness issue of the requirement choices by using the graph.


src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -60,31 +95,74 @@ export default Vue.extend({
     isCourseModelSelectingSemester: Boolean,
     reqs: Array as PropType<readonly SingleMenuRequirement[]>,
```
tcho6319, line 96: Can we do `type` and `required` for the props declarations?

src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -4,51 +4,86 @@
     contentClass="content-course"
     :leftButtonText="leftButtonText"
     :rightButtonText="rightButtonText"
-    :rightButtonIsDisabled="isDisabled"
+    :rightButtonIsDisabled="selectCourse == null"
```
tcho6319, line 7: Do you mean `selectedCourse == null`? I'm not sure if this is working properly. When I first open up the New Course Modal, the Add button should be grayed out. (Left is PR, right is master).
Screen Shot 2021-02-09 at 2 43 53 PM

tcho6319 (changes requested):
Thanks Sam! I requested 2 smaller changes, but once those are addressed, the PR should be good for approval.

src/components/Modals/NewCourse/NewCourseModal.vue
SamChou19815: changed

src/components/Modals/NewCourse/NewCourseModal.vue
SamChou19815: should be fixed now

tcho6319 (approved):
LGTM! Thank you :)

PR #268, opened 2021-02-09
[easy] Update type definition of dragula to avoid any

SamChou19815:
### Summary 

Two fewer `any`s, two fewer linter warnings!

### Test Plan 

Things still type-check.


tcho6319 (approved):
LGTM!

PR #269, opened 2021-02-09
Optimize the size of course json

SamChou19815:
### Summary 

Optimize the course json by keeping only the field we need and throw away all others. I also throw away some empty dummy field like `catalogDistr: ""`.

Final results: (top: before size, bottom: final size)

test

### Test Plan 

Requirement generation and the bottom bar course still works, which shows that no useful information is lost.


SamChou19815:
big diff is dying because the json is still too big to it :(

tcho6319 (changes requested):
Thanks for the update Sam! 

1. To confirm: The decorated requirements json is up-to-date with SP21 data right?
2. I'm getting a 500 error when I try to see the courses on the requirements sidebar. I don't see this on the master staging site. When looking at the [FetchCourses() logs](https://console.firebase.google.com/u/0/project/cornelldti-courseplan-dev/functions/logs?search=&&severity=DEBUG), I wonder if it has to do with changes made to the filtered course json files..
Screen Shot 2021-02-09 at 3 19 47 PM

SamChou19815:
@tcho6319 it turns out there is nothing wrong with the code. The function hasn't been deployed so it doesn't have the latest sp21 data in this PR. I redeployed the functions and the error goes away.

tcho6319 (approved):
Good catch! Yes, I forgot that we had to re-deploy FetchCourses() since we updated the data. LGTM!

PR #270, opened 2021-02-09
Factor out requirement graph builder code into its own file

SamChou19815:
### Summary 

We will need the requirement graph not just as a helper structure to compute progress but also as a store for some other purposes. For example, to accurately decide the requirement choices for `NewCourseModal`, we need to inspect the graph. Therefore, the code that computes the graph must be extracted out so that we can build the graph independently of the progress report.

This PR factors out the relevant code into `requirement-graph-builder-from-user-data.ts`. In the process, I found that `computeRequirementMap` is no longer used, so I deleted it.

### Test Plan 

Should have no change on requirement progress. Check the requirement progress to verify nothing changed.

### DIFF
--- a/src/containers/Dashboard.vue
+++ b/src/containers/Dashboard.vue
@@ -122,7 +122,7 @@ import {
   createAppUser,
   AppToggleableRequirementChoices,
 } from '@/user-data';
-import { computeRequirements } from '@/requirements/reqs-functions';
+import computeRequirements from '@/requirements/reqs-functions';
 import { CourseTaken, SingleMenuRequirement } from '@/requirements/types';
 import getCourseEquivalentsFromUserExams from '@/requirements/data/exams/ExamCredit';
 import getCurrentSeason, { checkNotNull, getCurrentYear, getSubjectColor } from '@/utilities';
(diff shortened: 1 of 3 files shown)


tcho6319 (approved):
LGTM. Test plan seems to work! I have some clarification questions for my own understanding: 
1. Is the difference between the requirements map (what is used for the New Add Modal at the moment and is used for Requirements Sidebar) and the requirements graph that the map is generated from the graph (that is, the map is a summary of the decisions made after the graph computation)?
2. To confirm: This PR does not change the structure of the requirements map, but the requirements graph builder is now factored out so that we can use it for inspecting the requirements graph.

SamChou19815:
> LGTM. Test plan seems to work! I have some clarification questions for my own understanding:
> 
> 1. Is the difference between the requirements map (what is used for the New Add Modal at the moment and is used for Requirements Sidebar) and the requirements graph that the map is generated from the graph (that is, the map is a summary of the decisions made after the graph computation)?
> 2. To confirm: This PR does not change the structure of the requirements map, but the requirements graph builder is now factored out so that we can use it for inspecting the requirements graph.

1. The add modal doesn't use a map. It uses a list which looks like: `[{req1: [course1, c2], req2: [c2, c3]}]`, which only provides a view from requirement to courses. The graph instead provides a two-sided view, and the view from course to requirements is the useful side, since you are exactly trying to match a list of requirement to a course there.
2. This is correct.

PR #271, opened 2021-02-10
Setup global data store (1/n)

SamChou19815:
### Summary 

This PR setups the global data store using Vuex. Right now I moved `toggleableRequirementChoices`, `subjectColors`, `uniqueIncrementer` into the store. As a result, we are able to remove various very ugly calls like

`this.$parent.$parent.$parent.createAppCourseFromCornellRosterCourse(...)`

since all the data necessary for creating a new course is available globally, so we don't need to rely on the top-level Dashboard anymore.

In addition, I setup `toggleableRequirementChoices`, `subjectColors`, `uniqueIncrementer` subscribers, so now the realtime firestore data is the source of truth. We no longer need to update them via the toplevel Dashboard anymore. We can simply call the functions to update firestore directly, and the Vuex store and firestore subscriber will take care of the rest.

I moved around some functions into new ts files mostly to avoid cyclic imports.

### Test Plan 

- change your major to A&S biological science, and play with the general chemistry toggleable requirement to verify toggleable requirement still works.
- add a course with a new subject you never added, verify that the new color is generated without issue.
- add a course and verify that the ID is still unique


src/user-data-converter.ts
```diff
@@ -0,0 +1,210 @@
...
+    major,
+    majorFN,
+    minor,
+    minorFN,
+    exam: 'exam' in data && data.exam.length > 0 ? [...data.exam] : [],
+    transferCourse: 'class' in data && data.class.length > 0 ? [...data.class] : [],
+    tookSwim: data.tookSwim,
+  };
+  return user;
+};
```
SamChou19815, line 210: moved from user-data to avoid cyclic import

src/utilities.ts
```diff
@@ -60,62 +60,6 @@ export function getRostersFromLastTwoYears(): string[] {
...
-  } else {
-    const colorIndex = Object.keys(subjectColors).length;
-    const key = Object.keys(subjectColors)[colorIndex % colors.length];
-    randomColor = subjectColors[key];
-  }
-
-  subjectColors[subject] = randomColor;
-  return randomColor;
-};
-
```
SamChou19815, line 118: moved into global-firestore-data to avoid cyclic import

src/containers/Login.vue
```diff
@@ -233,7 +233,6 @@ export default Vue.extend({
         .then(doc => {
           if (doc.exists) {
             this.performingRequest = false;
-            this.$store.commit('setCurrentUser', user);
```
SamChou19815, line 236: removed because it's never read before this commit, and now it's auto set when the user is logged in

src/store.ts
```diff
@@ -1,30 +1,99 @@
...
+import { checkNotNull } from './utilities';
 
 Vue.use(Vuex);
 
-type State = {
-  currentUser: firebase.User | null;
+type SimplifiedFirebaseUser = { readonly displayName: string; readonly email: string };
+
+export type VuexStoreState = {
+  currentFirebaseUser: SimplifiedFirebaseUser;
```
SamChou19815, line 13: renamed currentUser to currentFirebaseUser to avoid confusion when we later add in AppUser.

src/store.ts
```diff
@@ -1,30 +1,99 @@
...
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+let firestoreListenerUnsubscriber = (): void => {};
+
 fb.auth.onAuthStateChanged(user => {
   if (user) {
-    store.commit('setCurrentUser', user);
+    const simplifiedUser = {
+      displayName: checkNotNull(user.displayName),
+      email: checkNotNull(user.email),
+    };
```
SamChou19815, line 69: Enforce non-null display name and email here once and for all, instead of checking it everywhere else

src/store.ts
```diff
@@ -1,30 +1,99 @@
...
+    store.commit('setCurrentFirebaseUser', simplifiedUser);
+    const toggleableRequirementChoiceUnsubscriber = fb.toggleableRequirementChoicesCollection
+      .doc(simplifiedUser.email)
+      .onSnapshot(snapshot => {
+        const toggleableRequirementChoices = snapshot.data() || {};
+        store.commit('setToggleableRequirementChoices', toggleableRequirementChoices);
+      });
```
tcho6319, line 76: To confirm: Does this piece of code basically update the global data store when it listens to the changes that were made to the toggleable requirement choices in firestore?

src/components/Requirements/Requirements.vue
```diff
@@ -231,11 +235,10 @@ export default Vue.extend({
...
             result.data.courses.forEach((course: CornellCourseRosterCourse) => {
-              // @ts-ignore [We should resolve this later]
-              const createdCourse = this.$parent.createAppCourseFromCornellRosterCourse(
-                course,
-                true
-              );
+              const createdCourse = cornellCourseRosterCourseToAppCourse(course, true);
+              // compact field is implicitly used in Course.vue as a prop :(((
+              // We need to cleanup this!
+              // @ts-ignore
```
tcho6319, line 241: I'll add this as an item on Notion!

src/store.ts
```diff
@@ -1,30 +1,99 @@
...
 });
 
+export const subscribeToggleableRequirementChoicesChange = (
+  handler: (state: VuexStoreState) => void
+): (() => void) =>
+  store.subscribe((payload, state) => {
+    if (payload.type === 'setToggleableRequirementChoices') {
+      handler(state);
+    }
+  });
```
tcho6319, line 59: I see a subscriber for toggleable requirements, but I'm still confused about how we have subscribers for subject colors and unique incrementer.

src/containers/Dashboard.vue
```diff
@@ -198,76 +186,53 @@ export default Vue.extend({
   },
   mounted() {
     this.getInformationFromUser();
+    subscribeToggleableRequirementChoicesChange(() => {
+      // Avoid recomputing requirement before user data is initialized.
+      if (this.user.college !== '') this.recomputeRequirements();
+    });
```
tcho6319, line 192: To confirm: Does this piece of code mean that we are listening for changes to toggleable requirements, and if there is a change, we recompute requirements, provided that user data is initialized?

tcho6319 (commented):
Wow thanks for this Sam! Test plan seems to work. I left comments/questions to clarify my understanding, but once those are addressed, this PR should be good for approval.

src/store.ts
SamChou19815: Yes

src/store.ts
SamChou19815: Right now we don't. There wasn't any need now, since none of the component need to rerender whenever subject color or unique incrementer changes.

src/containers/Dashboard.vue
SamChou19815: Yes.

tcho6319 (approved):
LGTM! Tysm Sam :)

PR #272, opened 2021-02-10
Fix two easy @ts-expect-error

SamChou19815:
### Summary 

- NewSemester.vue: we can use string as key, so no need to convert to number.
- Semester.vue: we can just make the object field upper case, so we don't need to do the lower case conversion which TS doesn't understand.

### Test Plan 

- Check the console to see no key collision warning in NewSemester modal.
- Check semesters are still correctly ordered.


tcho6319 (approved):
LGTM! Test plan seems to work on my end.

PR #273, opened 2021-02-10
Derive college/major/minor full name instead of storing them

SamChou19815:
Depends on #271.

### Summary 

These names can be derived from the acronym and the requirement json directly, there is no need to store them. Removing them cleans up a lot of boilerplate code to process and pass them around.

This change will change the data format by making the ***FN field in firestore obsolete. It is safe and doesn't require any data migration since existing data with those field will just be ignored.

### Test Plan 

- go through onboarding to ensure full names are displayed correctly.
- check the requirement menu to see full names are displayed correctly.


tcho6319 (approved):
LGTM! Test plan seems to work as well. Thanks Sam :)

PR #274, opened 2021-02-10
Split AppUser into FirestoreUsername and AppOnboardingData

SamChou19815:
Depends on #273.

### Summary 

In an effort to make the format of the app data and firestore data more consistent with each other, this diff splits the AppUser into `FirestoreUsername` and `AppOnboardingData`, which is very close to the firestore data.

After this diff, we will be able to integrate the onboarding data into the global data store.

### Test Plan 

- go through onboarding to ensure full names are displayed correctly.
- check the requirement menu to see full names are displayed correctly.


tcho6319 (approved):
Looks great Sam! Thanks.

PR #275, opened 2021-02-11
Unify AppCourse & FirestoreSemesterCourse, AppSemester & FirestoreSemester

SamChou19815:
### Summary 

In the push to store some important data globally, we want the data in firestore to be consistent with the data used in the app, so that we don't have to do tricky transformations all the time. Semester and course data is such important global data.

`AppCourse` and `FirestoreSemesterCourse` are very similar to each other, except a few fields.  `AppSemester` and `FirestoreSemester` are almost identical, except that one contains `AppCourse` and the other contains `FirestoreSemesterCourse`.

This diff unifies them into the same type. In particular, it abandons AppCourse and AppSemester and switches them to the firestore version throughout the app. The `AppCourse` data has these fields that are different from `FirestoreSemesterCourse`:

- `check: boolean`. Unused field, so i removed it.
- `isReqCourse: boolean`. This is not necessary to the course, but would be more appropriate if it's a parameter passed into the course component, so I removed it.
- `subject, number` instead of `code`. In the places where we need subject and number, we split from the code.

After the above changes are done, `AppCourse` and `FirestoreSemesterCourse` are completely identical and we can only keep one type!

There was also a leftover from previous PR when I found that we are doing `course.compact = true;`. It turns out it's unnecessary, since the relevant field is passed into the course component already.

### Test Plan 

- add/edit/delete course still works.
- course drag and drop still works
- courses in the requirement menu still correctly show up

### Notes 

In this PR, I also want to get rid of the `v-bind="course"` on the course component, but failed to do so. It turns out that the sad drag and drop library depends on it :(

### DIFF
--- a/src/components/Requirements/CompletedSubReqCourse.vue
+++ b/src/components/Requirements/CompletedSubReqCourse.vue
@@ -37,7 +37,7 @@
 import Vue, { PropType } from 'vue';
 import ReqCourse from '@/components/Requirements/ReqCourse.vue';
 import { DisplayableRequirementFulfillment, CourseTaken } from '@/requirements/types';
-import { AppSemester, FirestoreSemesterType } from '@/user-data';
+import { FirestoreSemester, FirestoreSemesterType } from '@/user-data';
 
 Vue.component('reqcourse', ReqCourse);
 
@@ -57,7 +57,7 @@ export default Vue.extend({
     subReq: Object as PropType<DisplayableRequirementFulfillment>,
     subReqCourseId: Number,
     crsesTaken: Array as PropType<readonly CourseTaken[]>,
-    semesters: Array as PropType<readonly AppSemester[]>,
+    semesters: Array as PropType<readonly FirestoreSemester[]>,
   },
   data(): CompletedSubReq {
     return {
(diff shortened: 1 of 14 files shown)


src/user-data-converter.ts
```diff
@@ -23,9 +20,8 @@ const createCourseCreditRange = (course: CornellCourseRosterCourse): readonly [n
 };
 
 export const cornellCourseRosterCourseToAppCourse = (
```
tcho6319, line 22: Should we rename this `cornellCourseRosterCourseToFirestoreSemesterCourse` to be consistent?

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -100,19 +100,18 @@ export default Vue.extend({
       if (!this.$data.scrollable) event.preventDefault();
     },
     setCourseAndSemesterForNonTransferCredits(crseTaken: CourseTaken) {
+      const courseTakenCode = `${crseTaken.subject} ${crseTaken.code}`;
```
tcho6319, line 103: CompletedSubReqs seem to show with the default data values. I think `courseTakenCode` should be `${crseTaken.subject} ${crseTaken.number}`
Screen Shot 2021-02-11 at 11 10 30 AM

tcho6319 (changes requested):
Thanks for this refactor Sam! I noticed a bug and made a suggestion but once those are addressed, the PR should be good to go.

src/user-data-converter.ts
SamChou19815: renamed

src/components/Requirements/CompletedSubReqCourse.vue
SamChou19815: fixed

tcho6319 (approved):
LGTM! Tysm Sam.

PR #276, opened 2021-02-11
Setup global data listeners for all data except semesters

SamChou19815:
Depends on #275.

### Summary 

This PR setups up all firestore listeners except semesters and then connect them to the root Dashboard. Semester data is hard to setup listener due to drag and drop mess.

For all the listeners I set up, when all of the listeners finish the first `onSnapshot`, it will call the onLoad callback to inform the application that the first batch of data has been loaded (so that the application has full data to work with, like requirement computation). Dashboard turns this into a promise and combines this with the promise of semester data loading, and then wait on these promises before declaring the page is loaded.

### Test Plan 

- add/edit/delete course still works.
- course drag and drop still works
- courses in the requirement menu still correctly show up

### Notes

To integrate semester into this, we need to replace the drag and drop library first.

### DIFF
--- a/src/containers/Dashboard.vue
+++ b/src/containers/Dashboard.vue
@@ -99,25 +99,18 @@ import surfing from '@/assets/images/surfing.svg';
 
 import '@/vueDragulaConfig';
 import {
-  db,
-  usernameCollection,
-  semestersCollection,
-  onboardingDataCollection,
-} from '@/firebaseConfig';
-import {
   FirestoreUserName,
-  FirestoreOnboardingUserData,
   FirestoreSemester,
   AppOnboardingData,
   AppBottomBarCourse,
   FirestoreSemesterCourse,
 } from '@/user-data';
-import { createAppOnboardingData } from '@/user-data-converter';
 import computeRequirements from '@/requirements/reqs-functions';
 import { CourseTaken, SingleMenuRequirement } from '@/requirements/types';
 import getCourseEquivalentsFromUserExams from '@/requirements/data/exams/ExamCredit';
+import store, { initializeFirestoreListeners, subscribeRequirementDependencyChange } from '@/store';
+import { semestersCollection } from '@/firebaseConfig';
 import getCurrentSeason, { getCurrentYear } from '@/utilities';
-import store, { subscribeToggleableRequirementChoicesChange } from '@/store';
 
 Vue.component('course', Course);
 Vue.component('semesterview', SemesterView);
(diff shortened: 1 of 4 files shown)


src/containers/Dashboard.vue
```diff
@@ -181,56 +163,60 @@ export default Vue.extend({
...
-        } else {
-          this.startOnboarding();
-        }
-      });
-    },
-
     editSemesters(newSemesters: readonly FirestoreSemester[]) {
+      console.log(
+        newSemesters.map(it => ({ a: `${it.type}${it.year}`, c: it.courses.map(c => c.code) }))
+      );
```
tcho6319, line 217: Can we remove this console log?

src/containers/Dashboard.vue
```diff
@@ -181,56 +163,60 @@ export default Vue.extend({
...
       this.semesters = newSemesters;
-      this.recomputeRequirements();
+      // this.recomputeRequirements();
```
tcho6319, line 219: Can we remove this comment?

tcho6319 (changes requested):
Thanks Sam! Everything looks mostly good except a couple comments and a bug I noticed:

1. Req bar updates seem kind of buggy. Though I'm not sure if this is because we don't have a global data listener for semesters. 
PR #275 staging site. The req bar does update on drop: 

https://user-images.githubusercontent.com/54298311/107668259-7a544200-6c5e-11eb-8708-2464602906a8.mov

This PR staging site. The req bar doesn't update on drop. Also, the small compact card is of the crosslisted course, not the one that was dragged. However, the small compact card does disappear when you refresh the page.:

https://user-images.githubusercontent.com/54298311/107668893-2bf37300-6c5f-11eb-8f5e-602b68bfe7a3.mov

Once the comments are addressed, the PR should be good to go!

src/containers/Dashboard.vue
SamChou19815: removed

src/containers/Dashboard.vue
SamChou19815: ahh i'm doing some debugging, and it turns out this is the reason why the requirement bar doesn't update

SamChou19815:
> Thanks Sam! Everything looks mostly good except a couple comments and a bug I noticed:
> 
> 1. Req bar updates seem kind of buggy when I drag a course from the req bar to the semesterview. Though I'm not sure if this is because we don't have a global data listener for semesters.
>    PR #275 staging site. The req bar does update on drop:
> 
>  Screen.Recording.2021-02-11.at.11.41.36.AM.mov 
> This PR staging site. The req bar doesn't update on drop. Also, the small compact card is of the crosslisted course, not the one that was dragged. However, the small compact card does disappear when you refresh the page.:
> 
>  Screen.Recording.2021-02-11.at.11.48.23.AM.mov 
> Once the comments are addressed, the PR should be good to go!

the bad behavior of the double card is also on master :(, although i'm not sure when it's there. It's probably introduced by me at some point... I'm relatively confident that migrating to a new drag and drop library can fix that.

tcho6319 (approved):
LGTM! Thank you Sam. As for the bad double card behavior: this is one of the issues that we've had for a while (probably because of dragula) so I am hopeful that replacing the drag&drop library will resolve the issue.

PR #277, opened 2021-02-11
[easy] Tweak linter config

SamChou19815:
### Summary 

Easy linter cleanup, also fixed some linter warnings introduced in last PR.

### Test Plan 

CI.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,24 +29,10 @@ module.exports = {
     '@typescript-eslint/ban-ts-comment': ['warn'],
     // TODO: change no-console to error out for prod
     'no-console': ['warn'],
-    // TODO: fix copied js code that makes this warn necessary
-    'no-use-before-define': ['warn'],
-    'linebreak-style': 0,
-    //TODO: make requests asynchronous
     'no-await-in-loop': ['error'],
-    // don't require .vue extension when importing
-    'import/extensions': [
-      'error',
-      'always',
-      {
-        js: 'never',
-        ts: 'never',
-        vue: 'never',
-      },
-    ],
     // disallow reassignment of function parameters
     // disallow parameter object manipulation except for specific exclusions
-    'no-param-reassign': [2, { props: false }],
+    'no-param-reassign': ['error', { props: false }],
     // allow optionalDependencies
     'import/no-extraneous-dependencies': [
       'error',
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -29,24 +29,10 @@ module.exports = {
     '@typescript-eslint/ban-ts-comment': ['warn'],
     // TODO: change no-console to error out for prod
     'no-console': ['warn'],
-    // TODO: fix copied js code that makes this warn necessary
-    'no-use-before-define': ['warn'],
```
SamChou19815, line 33: already listed below as `off` due to TS. Removing it

.eslintrc.js
```diff
@@ -29,24 +29,10 @@ module.exports = {
...
-    'linebreak-style': 0,
```
SamChou19815, line 34: enforced by prettier

.eslintrc.js
```diff
@@ -29,24 +29,10 @@ module.exports = {
...
-    //TODO: make requests asynchronous
```
SamChou19815, line 35: todo is resolved

.eslintrc.js
```diff
@@ -29,24 +29,10 @@ module.exports = {
...
-    // don't require .vue extension when importing
-    'import/extensions': [
-      'error',
-      'always',
-      {
-        js: 'never',
-        ts: 'never',
-        vue: 'never',
-      },
-    ],
```
SamChou19815, line 46: already in the overrides below

.eslintrc.js
```diff
@@ -29,24 +29,10 @@ module.exports = {
...
     // disallow reassignment of function parameters
     // disallow parameter object manipulation except for specific exclusions
-    'no-param-reassign': [2, { props: false }],
+    'no-param-reassign': ['error', { props: false }],
```
SamChou19815, line 35: trying to be consistent of using string instead of numbers for linter error levels

.eslintrc.js
```diff
@@ -70,7 +56,6 @@ module.exports = {
       },
     ],
     'max-classes-per-file': ['off'],
-    'max-len': ['error', { code: 200 }],
```
SamChou19815, line 73: already enforced by prettier

.eslintrc.js
tcho6319: What requests did this TODO refer to? Or was it more of a general todo to make all requests async?

tcho6319 (approved):
LGTM! Thanks Sam. I just left a clarification question.

.eslintrc.js
SamChou19815: we used to do something like

```js
for (...) {
  await ...
}
```

This is bad because you are not making things run in parallel. Now we no longer have this kind of code.

PR #279, opened 2021-02-12
Cleanup Vue components

SamChou19815:
### Summary 

Cleanup vue components by

- Declaring each component prop as required or give it a default value
- Declare components in `components: {...}`, instead of `Vue.component('name', Component)`. This gives better type-checking and autocompletion.

Unfortunately, the Vuetr plugin has [a bug](https://github.com/vuejs/vetur/issues/2344) with imported types in props. Therefore, I have to make all the common types global. 

### Test Plan 

Move around the app and see no warnings in console.

### DIFF
--- a/src/components/Modals/NewCourse/NewCourseModal.vue
+++ b/src/components/Modals/NewCourse/NewCourseModal.vue
@@ -1,7 +1,7 @@
 <template>
   <flexible-modal
     title="Add Course"
-    contentClass="content-course"
+    content-class="content-course"
     :leftButtonText="leftButtonText"
     :rightButtonText="rightButtonText"
     :rightButtonIsDisabled="selectedCourse == null"
@@ -21,12 +21,12 @@
     <div v-if="isCourseModelSelectingSemester && selectedCourse == null">
       <div class="newCourse-title">Add this class to the following semester</div>
       <div class="newCourse-semester-edit">
-        <newSemester
+        <new-semester
           :type="season"
           :year="year"
           :isCourseModelSelectingSemester="isCourseModelSelectingSemester"
           @updateSemProps="updateSemProps"
-        ></newSemester>
+        />
       </div>
     </div>
     <div v-if="selectedCourse != null">
@@ -60,8 +60,9 @@ import Vue, { PropType } from 'vue';
 import SelectedRequirementEditor, {
   RequirementWithID,
 } from '@/components/Modals/NewCourse/SelectedRequirementEditor.vue';
-import { FirestoreSemesterType, CornellCourseRosterCourse } from '@/user-data';
 import { SingleMenuRequirement } from '@/requirements/types';
+import FlexibleModal from '@/components/Modals/FlexibleModal.vue';
+import NewSemester from '@/components/Modals/NewSemester.vue';
 import CourseSelector, {
   MatchingCourseSearchResult,
 } from '@/components/Modals/NewCourse/CourseSelector.vue';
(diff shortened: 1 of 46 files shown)


src/router/index.ts
```diff
@@ -4,7 +4,7 @@ import Router from 'vue-router';
 import Login from '@/containers/Login.vue';
 import Dashboard from '@/containers/Dashboard.vue';
 import Page404 from '@/containers/404.vue';
-import store from '@/store';
+import store from '../store';
```
tcho6319, line 7: Is it better to use `../` here instead of `@/`? I thought `@` referred to `src`.

src/containers/Dashboard.vue
```diff
@@ -57,18 +57,18 @@
...
       :exit="congratsExit"
       :buttonText="congratsButtonText"
       @hide="showTourEndWindow = false"
       v-if="showTourEndWindow"
     >
-    </tourwindow>
+    </tour-window>
     <div>
-      <bottombar
+      <bottom-bar
```
tcho6319, line 71: Without declaring the name of the component as in the old style (`Vue.component('bottombar', BottomBar);`), how do we know that bottom-bar refers to the component BottomBar? Just curious!

src/containers/Dashboard.vue
```diff
@@ -97,27 +96,15 @@ import TourWindow from '@/components/Modals/TourWindow.vue';
...
 import computeRequirements from '@/requirements/reqs-functions';
 import { CourseTaken, SingleMenuRequirement } from '@/requirements/types';
 import getCourseEquivalentsFromUserExams from '@/requirements/data/exams/ExamCredit';
 import store, { initializeFirestoreListeners, subscribeRequirementDependencyChange } from '@/store';
 import { semestersCollection } from '@/firebaseConfig';
 import getCurrentSeason, { getCurrentYear } from '@/utilities';
 
-Vue.component('course', Course);
 Vue.component('semesterview', SemesterView);
 Vue.component('requirements', Requirements);
```
tcho6319, line 107: Are we waiting on handling these until after the migration to the global data store?

src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <flexible-modal
     title="Add Course"
-    contentClass="content-course"
+    content-class="content-course"
```
tcho6319, line 4: Isn't the required prop name `contentClass`?

tcho6319 (commented):
Thanks for this great improvement in the codebase Sam! I left some comments but once those are addressed, the PR should be good to go!

src/router/index.ts
SamChou19815: It is used to make the ts project for requirement generator happy

src/containers/Dashboard.vue
SamChou19815: It's some standard case transform vue performs. CamelCase to lower-case-with-dash

src/containers/Dashboard.vue
SamChou19815: Yes

src/components/Modals/NewCourse/NewCourseModal.vue
SamChou19815: Also some case transformation on props

tcho6319 (approved):
LGTM! Thank you so much Sam.

PR #280, opened 2021-02-12
Add to the 120 credits on repeated courses

### LINKED ISSUE #141: [Requirements Handling] Add to the 120 credits on repeated courses
According to the university's information on the credits, the number of the total academic requirements needed increases for repeated courses.

Screen Shot 2020-09-24 at 11 33 32 AM

SamChou19815:
### Summary 

Resolve #141.

Implemented according to my proposal during the meeting:

- Lifting the total credits requirement out of the generated requirement json and special case it on frontend
- As a result, it will not become part of requirement graph as desired.
- In addition, the requirement json dropped from 2.5M to 800k!

### Test Plan 

Repeated courses now correctly count towards 120 credits:

repeated-courses

AAP: no college req as expected.
AAP
A&S: correctly counted 
AS
Business school: correctly counted 
BUSINESS
CALS: correctly counted
CALS
Engineering: no such requirement as expected
EN
Human ecology: correctly counted 
HE
ILR: correctly counted 
ILR


src/requirements/__test__/__snapshots__/requirement-data-id.test.ts.snap
```diff
@@ -9,7 +9,6 @@ Array [
   "College-AG-Physical and Life Sciences",
   "College-AG-Quantitative Literacy",
   "College-AG-Social Sciences and Humanities",
-  "College-AG-Total Academic Credits",
```
SamChou19815, line 12: snapshot is updated as expected, since there are no longer part of the requirement json any more. Since it's before launch, no migration is necessary. For more background, see https://github.com/cornell-dti/course-plan/pull/222

src/requirements/reqs-functions.ts
```diff
@@ -8,6 +9,121 @@ import {
...
+          'PE courses do not count.',
+        source: 'http://courses.cornell.edu/content.php?catoid=41&navoid=11561',
+      };
+      break;
+    case 'AS':
+      requirement = {
+        ...requirementCommon,
+        id: 'College-AS-total-credits',
+        description:
+          '120 academic credits are required' +
```
tcho6319, line 66: Should it be `'120 academic credits are required. '`

src/requirements/reqs-functions.ts
```diff
@@ -8,6 +9,121 @@ import {
...
+          'PE courses and courses numbered 1000-1099 do not count towards the 120 credits',
```
tcho6319, line 67: I think this string is missing a period.

src/requirements/reqs-functions.ts
```diff
@@ -8,6 +9,121 @@ import {
...
+        source: 'http://courses.cornell.edu/content.php?catoid=41&navoid=11570#credit-req',
+      };
+      break;
+    case 'HE':
+      requirement = {
+        ...requirementCommon,
+        id: 'College-HE-total-credits',
+        description:
+          '120 academic credits are required' +
+          'PE courses and courses numbered 1000-1099 do not count towards the 120 credits',
```
tcho6319, line 77: I think some of the concatenated strings in the descriptions for the different colleges are missing periods and spaces at the end.

src/requirements/reqs-functions.ts
```diff
@@ -8,6 +9,121 @@ import {
...
+  (!courseIsAPIB(course) && course.subject !== 'PE' && !course.number.startsWith('10'));
+
+const getTotalCreditsFulfillmentStatistics = (
+  college: string,
+  courses: readonly CourseTaken[]
+): FulfillmentStatistics | null => {
+  const requirementCommon = {
+    sourceType: 'College',
+    sourceSpecificName: college,
+    name: 'Total Academic Credits',
```
tcho6319, line 42: Would the `name` field be how the frontend would special case the 120 total academic credits?

tcho6319 (changes requested):
Thank you so much Sam! I requested some smaller changes but everything looks mostly good. To note: I'm not sure if we can say that #141 would be completely resolved by this PR though, because the issue discusses how if the user repeats a course, the min total number of academic credits would increase by the number of credits of that course. Though the credits of repeated courses are taken into account for the total academic credits, the number still remains at 120.

src/requirements/reqs-functions.ts
SamChou19815: fixed

src/requirements/reqs-functions.ts
SamChou19815: fixed

src/requirements/reqs-functions.ts
SamChou19815: fixed

src/requirements/reqs-functions.ts
SamChou19815: No, it's handled by completely remove them from the requirement json, and inject it as the first college requirement in the requirement menu.

SamChou19815:
> Thank you so much Sam! I requested some smaller changes but everything looks mostly good. To note: I'm not sure if we can say that #141 would be completely resolved by this PR though, because the issue discusses how if the user repeats a course, the min total number of academic credits would increase by the number of credits of that course. Though the credits of repeated courses are taken into account for the total academic credits, the number still remains at 120.

@tcho6319  I'm not sure if I understand the question. Is there an counterexample that shows this PR doesn't solve everything in that issue.

tcho6319:
> > Thank you so much Sam! I requested some smaller changes but everything looks mostly good. To note: I'm not sure if we can say that #141 would be completely resolved by this PR though, because the issue discusses how if the user repeats a course, the min total number of academic credits would increase by the number of credits of that course. Though the credits of repeated courses are taken into account for the total academic credits, the number still remains at 120.
> 
> @tcho6319 I'm not sure if I understand the question. Is there an counterexample that shows this PR doesn't solve everything in that issue.

For example, since I have CS 1110 twice in my schedule, that would mean that I repeated the course. Based on the policy re: repeated courses, the total number of academic credits that I would need to graduate should increase from 120 to 124 because CS 1110 is 4 credits. In the screenshot, we see that the minimum number of credits remains 120.
Screen Shot 2021-02-12 at 8 32 59 PM

Also, this is more related to the second point about 10xx level courses, but the required number of credits I need to graduate is 121 total academic credits because I took ENGRG 1094 (a supplemental class for lin alg), which is 1 credit. 
image

tcho6319 (approved):
LGTM! Thanks Sam. I shared some examples of why I think this PR might not fully address #141 above but I think this PR is good on its own. I think tackling changing the Total Academic Credits minimum might be better handled in a separate PR.

SamChou19815:
It turns out that handling the increase isn't hard, so I just included it there.

PR #282, opened 2021-02-12
Add semesters data to global store

SamChou19815:
### Summary 

Unlike other data, we don't rely on real time listening here. It's because real time listening still has a delay, which makes drag and drop really wracky. Therefore, we still synchronously update the store and at the same time update firestore to make things consistent. However, with this setup we can now read semesters from everywhere!

Remaining TODOs:

Integrate requirement data into global store.

### Test Plan 

- add/edit/remove semesters
- drag and drop courses
- open the requirement menu to check all courses are loaded

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -2,7 +2,7 @@ import Vue from 'vue';
 import Vuex, { Store } from 'vuex';
 
 import * as fb from './firebaseConfig';
-import { checkNotNull } from './utilities';
+import getCurrentSeason, { checkNotNull, getCurrentYear } from './utilities';
 
 Vue.use(Vuex);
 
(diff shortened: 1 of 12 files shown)


src/store.ts
```diff
@@ -93,18 +99,24 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
...
       userNameInitialLoadFinished &&
       onboardingDataInitialLoadFinished &&
+      semestersInitialLoadFinished &&
       toggleableRequirementChoiceInitialLoadFinished &&
       subjectColorInitialLoadFinished &&
-      uniqueIncrementerInitialLoadFinished
+      uniqueIncrementerInitialLoadFinished &&
+      !emitted
     ) {
+      emitted = true;
```
SamChou19815, line 119: lack of emitted is a bug. Without this flag set to true after first load emit, it will keep calling onLoad whenever there is an update in any listeners.

tcho6319 (approved):
Thanks for this Sam! 

I noticed the strangest bug both on this PR's staging site and master :( but I'll note it here. When I type in anything else for the add modal, it's fine but when I start to type in "wines", I get a console error. Any idea on what could be the reason? I'm good with approving this PR though because I don't think the bug has to do with changes from this PR.

https://user-images.githubusercontent.com/54298311/107839051-5b959e80-6d77-11eb-89de-e7195cd043d7.mov

SamChou19815:
@tcho6319 fixed in #284.

PR #285, opened 2021-02-13
Enforce single-select on non-double-countable requirements in add modal

SamChou19815:
### Summary 

This PR implements my proposal that enforces that we can only select one non-double-countable requirement in the add modal, and all the double-countable requirements are auto-selected.

The purpose of this PR is not to implement a good design, but is designed to be quickly mergable to unblock @benjamin-shen's work of saving the choice and to inform design (@einc) what should be done. (especially the wording might need a lot of tweak)

### Test Plan 

Example of MATH 4710:

Screen Shot 2021-02-12 at 22 07 32
Screen Shot 2021-02-12 at 22 07 35

Also an example of cs 3110 that has non-double-countable non-self-check requirements:

Screen Shot 2021-02-12 at 22 15 34
Screen Shot 2021-02-12 at 22 15 39


src/components/Modals/NewCourse/RequirementRadioSelectButton.vue
```diff
@@ -1,79 +0,0 @@
-<template>
```
SamChou19815, line 1: according to the meeting, no more radio select.

src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -133,19 +126,27 @@ export default Vue.extend({
...
+            if (!subRequirement.allowCourseDoubleCounting) {
+              potentialRequirements.push({ id: subreqs[j].id, name: subRequirement.name });
+            }
           }
+          this.requirementsThatAllowDoubleCounting = requirementsThatAllowDoubleCounting;
           this.relatedRequirements = relatedRequirements;
           this.potentialRequirements = potentialRequirements;
-          this.selectedRequirementIDs = relatedRequirements.map(it => it.id);
+          this.selectedRequirementID =
+            relatedRequirements.length > 0 ? relatedRequirements[0].id : '';
```
SamChou19815, line 149: It's important to note that the algorithm used here is still flaky. It doesn't account for toggleable requirements at all. Fixing it would be the job of another diff.

einc:
Screen Shot 2021-02-13 at 2 54 58 PM

I added a CS major in COE with a ISST minor. ENGRD 2700 fulfills the "probability" & "engineering distribution" for CS major and the "Required Courses" requirement for the ISST minor. All minors by default are double-countable, right?
In this example, should Required Courses also be pre-selected?

SamChou19815:
> Screen Shot 2021-02-13 at 2 54 58 PM
> 
> I added a CS major in COE with a ISST minor. ENGRD 2700 fulfills the "probability" & "engineering distribution" for CS major and the "Required Courses" requirement for the ISST minor. All minors by default are double-countable, right?
> In this example, should Required Courses also be pre-selected?

@einc  This is a good point, but it's not planned to be addressed by this PR. In this PR, I'm just naively looking at the allowDoubleCounting flag in the requirement json, while the correct way to do this is to call the same functions that accurately decide whether a requirement double counting as in the requirement graph algo. This is the scope of another PR.

src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -191,17 +192,16 @@ export default Vue.extend({
       this.reset();
       this.closeCurrentModal();
     },
-    onSelectedChange(selected: readonly string[]) {
-      this.selectedRequirementIDs = selected;
+    onSelectedChange(selected: string) {
+      this.selectedRequirementID = selected;
     },
     reset() {
```
tcho6319, line 198: Should we also reset `this.requirementsThatAllowDoubleCounting = []`?

tcho6319 (commented):
Thanks Sam! I just have one comment but this PR should be good at least to unblock @benjamin-shen's work.

src/components/Modals/NewCourse/NewCourseModal.vue
SamChou19815: No requirements that allow double counting is immutable. A requirement either allows or doesn't allow double counting, regardless of user choices

tcho6319 (approved):
LGTM! I'll prob make Notion items for the modifications we want to do to the Add Modal so that it is more polished, based on the discussion threads.

PR #288, opened 2021-02-13
[easy] Remove the deprecated forcedReload argument

SamChou19815:
### Summary 

deprecated

It is deprecated, and according to https://stackoverflow.com/questions/55127650/location-reloadtrue-is-deprecated, we probably don't need it.

### Test Plan 

👀


tcho6319 (approved):
LGTM. From our prior use of `forcedReload`, it seems we wouldn't want to perform a forced reload anyway so I think we're ok to remove it.

PR #290, opened 2021-02-14
Cleanup requirement computation

SamChou19815:
### Summary 

This PR cleans up requirement computation code by eliminating some unnecessary data transformation. It turns out that we don't really need `SingleMenuRequirement`, because it's very similar to `GroupedRequirementFulfillmentReport`. The only significant difference is that `SingleMenuRequirement` has `ongoing, completed, fulfilled, required` field that tells the fulfillment stat of requirements, but they can be easily derived in the places that need them.

In this PR, I compute the `ongoing, completed` list on demand inside `RequirementView` and `filfilled, required` inside `RequirementHeader`. As a result, we have a much slimmer set of data to pass around.

For other changes, I will explain in the review comments on why they are safe.

### Test Plan 

Side by side comparison of this PR vs master. All requirement items are identical.

Screen Shot 2021-02-13 at 19 22 22

### Notes 

After this PR, we can finally make requirements into the global store.

PR #291, opened 2021-02-14
Setup global types for requirements

SamChou19815:
### Summary 

Move requirement types into a global typedef file, so that they can be used everywhere without import. This is used to unblock final setup to add requirements into global store.

Depends on #290.

### Test Plan 

This is mostly a type only change, so check everything still type checks.
I also migrated more components to avoid using `Vue.component('...', Component)`. We can check it works by ensuring every part of the app still loads without error.

### DIFF
--- a/src/components/Requirements/SubRequirement.vue
+++ b/src/components/Requirements/SubRequirement.vue
@@ -74,7 +74,7 @@
       <div v-if="displayDescription" class="subreqcourse-wrapper">
         <div v-for="(subReqCourseSlot, id) in subReqCoursesArray" :key="id">
           <div v-if="subReqCourseSlot.isCompleted" class="completedsubreqcourse-wrapper">
-            <completedsubreqcourse
+            <completed-sub-req-course
               :subReqCourseId="id"
               :crsesTaken="subReqCourseSlot.courses"
               @deleteCourseFromSemesters="deleteCourseFromSemesters"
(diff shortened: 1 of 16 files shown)


src/components/Requirements/SubRequirement.vue
```diff
@@ -107,17 +107,11 @@ import CompletedSubReqCourse from '@/components/Requirements/CompletedSubReqCour
...
-  SubReqCourseSlot,
-  CrseInfo,
-} from '@/requirements/types';
+import { SubReqCourseSlot, CrseInfo } from '@/requirements/types';
 import { clickOutside } from '@/utilities';
 
 import { cornellCourseRosterCourseToFirebaseSemesterCourse } from '@/user-data-converter';
 
-Vue.component('completedsubreqcourse', CompletedSubReqCourse);
 Vue.component('incompletesubreqcourse', IncompleteSubReqCourse);
```
tcho6319, line 115: Is there a blocker for migrating IncompleteSubReqCourse as was done for CompletedSubReqCourse?

tcho6319 (approved):
LGTM! I found a bug that's on this PR's staging site and on master's as well. I don't think it's related to #290 or #291 since it's on master so I would be ok with approving this PR as it seems mostly type changes anyway.

I observe the following buggy behavior for CALS and A&S credits but not for other subreqs. I haven't tried other colleges but I wonder if it has to do with college credits. Do you have any guesses?
 

https://user-images.githubusercontent.com/54298311/107996030-646fb580-6fad-11eb-9ed2-b02ffb39e726.mov

src/components/Requirements/SubRequirement.vue
SamChou19815: Yes, this component is still using these imported types:

https://github.com/cornell-dti/course-plan/blob/f046f281be638dbc9e494aa557acfbd8c32b21de/src/requirements/types.ts#L195-L205

I plan to investigate in another PR whether these types should be moved out to global types.

SamChou19815:
> LGTM! I found a bug that's on this PR's staging site and on master's as well. I don't think it's related to #290 or #291 since it's on master so I would be ok with approving this PR as it seems mostly type changes anyway.
> 
> I observe the following buggy behavior for CALS and A&S credits but not for other subreqs. I haven't tried other colleges but I wonder if it has to do with college credits. Do you have any guesses?
> 
>  Screen.Recording.2021-02-15.at.4.38.50.PM.mov

This looks like the subreq courses is a little buggy...

PR #292, opened 2021-02-14
Move requirement data into Vuex global store

SamChou19815:
### Summary 

Finally, we can move all the requirements into the global store!

Depends on #291.

### Test Plan 

Add and remove a course to see requirements still change in real time.

### DIFF
--- a/src/requirements/requirement-graph-builder-from-user-data.ts
+++ b/src/requirements/requirement-graph-builder-from-user-data.ts
@@ -1,4 +1,3 @@
-import store from '../store';
 import RequirementFulfillmentGraph from './requirement-graph';
 import buildRequirementFulfillmentGraph from './requirement-graph-builder';
 import requirementJson from './typed-requirement-json';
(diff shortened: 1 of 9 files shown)


src/requirements/requirement-graph-builder-from-user-data.ts
```diff
@@ -22,16 +21,16 @@ function forfeitTransferCredit(coursesTaken: readonly CourseTaken[]): readonly C
 }
 
 export default function buildRequirementFulfillmentGraphFromUserData(
-  coursesTaken: readonly CourseTaken[]
+  coursesTaken: readonly CourseTaken[],
+  { college, major: majors, minor: minors }: AppOnboardingData,
+  toggleableRequirementChoices: AppToggleableRequirementChoices
```
tcho6319, line 26: Why do we not directly use the data from `store.state`? Is this to avoid circular import?

tcho6319 (approved):
Yay :) LGTM! Tysm Sam. Test plan seems to work well, and I left a clarification question.

src/requirements/requirement-graph-builder-from-user-data.ts
SamChou19815: Yes.

PR #293, opened 2021-02-15
[easy] Factor all fetch courses firebase function call into a typed version

SamChou19815:
### Summary 

Instead of calling `FetchCourses` function inside the components and worry about types on their own, I setup a typed version of this function inside `firebaseConfig.ts`, so that type safety is enforced. I also moved the import of `firebase/functions` into `firebaseConfig.ts`, to ensure that it's only initialized once.

### Test Plan 

Courses are still fetched.

Screen Shot 2021-02-15 at 18 25 31
Screen Shot 2021-02-15 at 18 25 36

### DIFF
--- a/src/components/Requirements/Requirements.vue
+++ b/src/components/Requirements/Requirements.vue
@@ -59,9 +59,7 @@
 </template>
 
 <script lang="ts">
-import firebase from 'firebase/app';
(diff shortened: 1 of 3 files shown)


tcho6319 (approved):
LGTM! Test plan works as well. Thanks Sam

PR #294, opened 2021-02-16
[easy] Cleanup dependencies in firebase functions

SamChou19815:
### Summary 

- bumped firebase deps
- Removed eslint dependencies: the linter config is certainly not respected and unused. It's better to rewrite the functions in TS at some later time, and then unifies everything using the root linter config.

### Test Plan 

I redeployed the functions and verified that the FetchCourses function still works on master.

### Notes

This cleanup should be able to remove the dependency vulnerability warning on `minimalist`.


tcho6319 (approved):
LGTM. Thank you Sam. I'll make a Notion item for converting the functions to TS and linking them to the root linter config.

PR #295, opened 2021-02-16
Fix stale data issue of CompletedSubReqCourse

SamChou19815:
### Summary 

Fix the issue mentioned in https://github.com/cornell-dti/course-plan/pull/291#pullrequestreview-590755830.

The root cause is that the component will not be rerendered when props change, because it derives data only once in `mounted`. This PR changes it to use computed properties to avoid this issue.

### Test Plan 

https://user-images.githubusercontent.com/4290500/108120584-83368080-706f-11eb-8dc1-fb22d86b3d02.mov

### DIFF
--- a/src/components/Requirements/CompletedSubReqCourse.vue
+++ b/src/components/Requirements/CompletedSubReqCourse.vue
@@ -18,15 +18,15 @@
       </div>
       <div class="completed-reqCourses-course-object-wrapper">
         <req-course
-          :color="color"
-          :subject="courseSubject"
-          :number="courseNumber"
+          :color="courseColorAndSemesterLabel.color"
+          :subject="courseTaken.subject"
+          :number="courseTaken.number"
           :compact="true"
           :isCompletedReqCourse="true"
           class="completed-reqCourses-course-object"
         />
         <div class="completed-reqCourses-course-object-semester">
-          {{ semesterLabel }}
+          in {{ courseColorAndSemesterLabel.semesterLabel }}
         </div>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -37,94 +37,55 @@
...
-    crsesTaken: { type: Array as PropType<readonly CourseTaken[]>, required: true },
-  },
-  data(): CompletedSubReq {
-    return {
-      semesterType: 'Fall', // 'Fall' is an arbitrary FirestoreSemesterType value that user will never see
-      semesterYear: 0,
-      color: '',
-      transferCreditColor: 'DA4A4A', // Arbitrary color for transfer credit
-      courseSubject: '',
-      courseNumber: '',
```
SamChou19815, line 65: these two field can be get from `courseTaken` directly

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -37,94 +37,55 @@
...
-      courseUniqueId: 0,
```
SamChou19815, line 66: If it's transfer, keep it 0, if it's not, then we read from the `courseTaken` object.

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -37,94 +37,55 @@
...
-  },
-  mounted() {
-    const crseTaken = this.crsesTaken[0];
-    this.isTransferCredit = crseTaken.subject === 'AP' || crseTaken.subject === 'IB';
-    if (!this.isTransferCredit) {
-      this.setCourseAndSemesterForNonTransferCredits(crseTaken);
-    } else {
-      this.setCourseAndSemesterForTransferCredits(crseTaken);
-    }
+    courseTaken: { type: Object as PropType<CourseTaken>, required: true },
```
SamChou19815, line 48: we are always only use the first element in array, so we should just pass that element.

SamChou19815 (commented):
some explanations

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -37,94 +37,55 @@
...
-      this.courseNumber = crseTaken.number;
-    },
+  },
+  methods: {
     onReset() {
-      this.$emit('deleteCourseFromSemesters', this.courseUniqueId);
+      this.$emit(
+        'deleteCourseFromSemesters',
+        this.isTransferCredit ? 0 : this.courseTaken.courseId
+      );
```
tcho6319, line 88: Resetting doesn't seem to work. I think it's because in Dashboard.vue, the `deleteCourseFromSemesters()` method still uses `uniqueID`. 


https://user-images.githubusercontent.com/54298311/108152331-d7a92280-70a6-11eb-8fc3-d7af95550488.mov

tcho6319 (changes requested):
Thanks Sam! I requested changes because the Reset functionality does not seem to work. This PR seems to resolve the smaller bug where the border color of the req course card updates when the course card in the semesterview is updated.

src/components/Requirements/CompletedSubReqCourse.vue
SamChou19815: should be fixed now

tcho6319 (approved):
LGTM! Reset works again :)

PR #296, opened 2021-02-16
Connect swim test choice in onboarding to requirements

SamChou19815:
### Summary 

This PR connects swim test choice to onboarding requirements. Instead of creating some equivalent courses, I decide just to special case swim test requirement for the following reasons:

1. We must ensure that the dummy swim test course does not get counted as a PE course for PE requirement.
2. The swim test course should not count towards total credit.
3. It also doesn't help to make swim test requirement part of requirement graph.

Depends on #295.

### Test Plan 

When selected took swim test in onboarding:
selected

When unselected took swim test in onboarding:
unselected


src/requirements/data/constants.ts
```diff
@@ -7,3 +7,28 @@ export const CREDITS_COURSE_ID = 10;
...
+  'SP18',
+  'SU18',
+  'FA18',
+  'SP19',
+  'SU19',
+  'FA19',
+  'SP20',
+  'FA20',
+  'SP21',
+];
```
tcho6319, line 34: Were these roster values hardcoded? Do you think it would be good to generate this list of rosters based on the rosters we get from the fetcher (i.e. the filtered-course files)?

src/requirements/requirement-fronend-computation.ts
```diff
@@ -123,6 +128,45 @@ const getTotalCreditsFulfillmentStatistics = (
...
+    source: 'http://courses.cornell.edu/content.php?catoid=41&navoid=11637',
+    courses: [Object.fromEntries(ALL_ROSTERS.map(roster => [roster, [SWIM_TEST_COURSE_ID]]))],
+    subRequirementProgress: 'any-can-count',
+    fulfilledBy: 'courses',
+    minCount: 1,
+  };
+  const swimClasses = courses.filter(it => it.courseId === SWIM_TEST_COURSE_ID);
+  if (tookSwimTest) {
+    swimClasses.push({
+      roster: 'SP21',
```
tcho6319, line 153: To confirm: Would we want the roster field value to be the most recent roster we have course data for?

tcho6319 (commented):
LGTM for the most part, Sam! Thank you. I made a comment about how we want to retrieve all the rosters.

src/requirements/requirement-fronend-computation.ts
SamChou19815: updated to use generated roster json

tcho6319 (approved):
Awesome. Looks great thanks Sam.

PR #297, opened 2021-02-16
Refactor IncompleteSubReqCourse to avoid relying on imported types

SamChou19815:
## Summary

This diff continues the work of #279. It refactors the IncompleteSubReqCourse component so that the prop no longer relies on `CrseInfo` and `SubReqCourseSlot`, so that we can cleanup the last `Vue.component(...)` call that doesn't work well with Vuetr.

I also moved the type of `SubReqCourseSlot` from `src/requirement/types` into `SubRequirement.vue`, since only that component and `Requirement.vue` are using it.

## Test Plan

See all courses and incomplete subreq courses still load

### DIFF
--- a/src/components/Requirements/SubRequirement.vue
+++ b/src/components/Requirements/SubRequirement.vue
@@ -81,17 +81,18 @@
             />
           </div>
           <div v-if="!subReqCourseSlot.isCompleted" class="incompletesubreqcourse-wrapper">
-            <incompletesubreqcourse
+            <incomplete-sub-req-course
               :subReq="subReq"
               :subReqCourseId="id"
-              :crseInfoObjects="subReqCourseSlot.courses"
+              :courseIDs="subReqCourseSlot.courses.map(it => it.crseIds)"
               :subReqFetchedCourseObjectsNotTakenArray="subReqFetchedCourseObjectsNotTakenArray"
-              :subReqCoursesArray="subReqCoursesArray"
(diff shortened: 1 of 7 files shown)


src/components/Requirements/SubRequirement.vue
```diff
@@ -106,15 +107,23 @@ import CompletedSubReqCourse from '@/components/Requirements/CompletedSubReqCour
 import IncompleteSubReqCourse from '@/components/Requirements/IncompleteSubReqCourse.vue';
 import DropDownArrow from '@/components/DropDownArrow.vue';
 
-import { SubReqCourseSlot, CrseInfo } from '@/requirements/types';
+import { CrseInfo } from '@/requirements/types';
```
tcho6319, line 110: Just a little confused on my end. Isn't `CrseInfo` an imported type? How are we able to keep this import statement for `SubRequirement.vue` and remove its `Vue.component()` call, but we wanted to get rid of `CrseInfo` for `IncompleteSubReqCourse.vue`?

tcho6319 (approved):
LGTM Sam! Thanks. I just left a question because I was a little confused, but this PR looks good to go.

src/components/Requirements/SubRequirement.vue
SamChou19815: The bug with the vuetr plugin is that imported type cannot appear in props.

PR #289, opened 2021-02-13
[easy] Strictify linter config from vue/essential to vue/recommended

SamChou19815:
### Summary 

This PR strictifies linter config from `vue/essential` to `vue/recommended`. It results in the additional "strongly recommended" and "recommended" rules that you can see here: https://eslint.vuejs.org/rules/.

The fix in this PR is auto generated by `npm run lint`. Here is a list of issues fixed.

- `v-bind:xxx="..."` to `:xxx="..."` shorthand
- `v-on:click="..."` to `@click="..."` shorthand
- No `this` in templates.

This PR also adds two v-html warnings. They should be fixed later (opportunity for someone else to get cleanup credit), since they allow XSS attack. Once they are fixed, the linter level should be turned to error to ensure we don't do this again.

Depends on #287.

### Test Plan 

CI

PR #298, opened 2021-02-17
Remove v-html directives

willespencer:
### Summary 

Remove two uses of v-html directives to fix a couple linter warnings due to recent linter config changes (see #289). The linter restricts this because of possible security issues from injecting html (see https://vuejs.org/v2/guide/security.html#Injecting-HTML).

- [x] Remove usage in CourseCaution (currently only text passed through, not html)
- [x] Remove usage in TourWindow (created a separate image tag to handle that html)
- [x] Set no-v-html to cause a linter error in the future

Remaining TODOs:

- [ ] resolve remaining linter warnings (currently 18 left) and TODOs in .eslintrc.js

See Notion item to handle all linter warnings here: https://www.notion.so/Clean-Code-Week-NPM-run-lint-warnings-and-errors-f01ddbf6aa9047739d17a4f104773584

### Test Plan 

1. Confirm caution modals still say 'Duplicate' when two courses are duplicated in a plan
2. Confirm the first page of the walkthrough is unchanged, and the last page still has the emoji icon

### DIFF
--- a/src/components/Modals/TourWindow.vue
+++ b/src/components/Modals/TourWindow.vue
@@ -40,8 +40,8 @@ export default Vue.extend({
     text: { type: String, required: true },
     exit: { type: String, required: true },
     buttonText: { type: String, required: true },
-    alt: { type: String, required: false, default: ''  },
-    image: { type: Object, required: false, default: null  },
+    alt: { type: String, required: false, default: '' },
+    image: { type: String, required: false, default: null },
   },
   data() {
     return {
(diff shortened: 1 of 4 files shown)


src/components/Modals/TourWindow.vue
```diff
@@ -37,6 +40,8 @@ export default Vue.extend({
     text: { type: String, required: true },
     exit: { type: String, required: true },
     buttonText: { type: String, required: true },
+    alt: { type: String, required: false, default: ''  },
+    image: { type: Object, required: false, default: null  },
```
SamChou19815, line 44: I think the type should be `String`

SamChou19815 (approved):
Verified that things still work.

tcho6319 (approved):
LGTM! The Test Plan works. Thanks Will!

PR #299, opened 2021-02-17
Cleanup BottomBar courses

SamChou19815:
### Summary 

This PR moves some of the data transformation code for `AppBottomBarCourse` into their own components, so that `Dashboard.vue` doesn't need to contain random data format changing code that doesn't seem to be very top-level. Also, this makes the data format of BottomBar course closer to firebase course, which paves way for unifying the two.

It also resolves this [notion task](https://www.notion.so/82bf355c4b7b49f99691eb9f3afb203c?v=f61b6258c5634200a9f45e381a0fb501&p=40fbeb2b11384fc391790a4ae173c60b) to deduplicate a lot of vue template copy-pasta when dealing with small-width screen.

### Test Plan 

Desktop comparison against master:
desktop
Mobile comparison against master:
mobile

### Note

This PR does introduce some behavior change. In the master version on mobile, the info is below the reviews, while in this PR, info is above the reviews. I can't find any design on figma regarding the mobile behavior, so I'm not sure whether the original behavior is a bug.

### DIFF
--- a/src/components/BottomBar/BottomBarCourse.vue
+++ b/src/components/BottomBar/BottomBarCourse.vue
@@ -1,146 +1,17 @@
 <template>
   <div class="bottombarcourse">
     <div class="bottombarcourse-wrapper">
-      <div
-        :class="
-          isSmallerWidth
-            ? 'bottombarcourse-bar-info-noOverflow'
-            : 'bottombarcourse-bar-info-overflow'
-        "
-      >
-        <div class="info">
-          <div class="info-section-wrapper">
-            <div>
-              <div class="section">
-                <h1 class="info-head">Credits</h1>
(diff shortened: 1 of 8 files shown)


tcho6319 (changes requested):
Thanks Sam! How the BottomBar looks on master right now at the different breakpoints is the intended design I believe (@einc, can you confirm?). So, info (the gray box) should be below the reviews (white box). Do you think you can change it so that the ordering of the bottom bar course is the same as master's? Other than that, the PR should be good!

SamChou19815:
@tcho6319 fixed in last commit by creating new components

tcho6319 (approved):
LGTM. Thanks Sam!

PR #300, opened 2021-02-17
Remove console logs

willespencer:
### Summary 

Remove 4 console logs in components to decrease linter warnings from 16 -> 12. This [notion item](https://www.notion.so/82bf355c4b7b49f99691eb9f3afb203c?v=f61b6258c5634200a9f45e381a0fb501&p=3068e663c70f45edb54c2eb67e3d6ed9) explains why we can just remove them, but they were all just logging errors either from firebase or our from our own returned requirements promises.

- [x] Removed all console logs in src/components

### Test Plan 

Confirm that everything still works where console logs were (so logging in, requirements, and sub req courses)

### Notes 

I chose not to change console logs to error in the future instead of just provide warnings so that people can still log when testing their code. Do people agree with this?

### DIFF
--- a/src/components/NavBar.vue
+++ b/src/components/NavBar.vue
@@ -28,9 +28,7 @@ export default Vue.extend({
       firebase
         .auth()
         .signOut()
-        .then(
-          () => window.location.reload(),
-        );
+        .then(() => window.location.reload());
     },
     editProfile() {
       this.$emit('editProfile');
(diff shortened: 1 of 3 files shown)


SamChou19815 (approved):
LGTM.

SamChou19815:
> I chose not to change console logs to error in the future instead of just provide warnings so that people can still log when testing their code. Do people agree with this?

I agree. Sometimes when you set error level to error, the dev server won't let you test your apps anymore... Since github will annotate all the console.log warnings in the code changes page, we can still easily find them.

tcho6319 (approved):
LGTM! And yes, I agree. I think showing warnings for console logs instead of erroring out would be better for devs when debugging/testing.

PR #301, opened 2021-02-18
Setup shared bottom bar state

SamChou19815:
### Summary 

The state of bottom bar is complicated. We have bottom courses and see all courses. We have the course that needs to be focused. In addition, bottom bar courses can be updated from everywhere, both inside the bottom bar by clicking and removing, and outside of it by clicking a course in semester view.

It's important to realize that we shouldn't really partition the courses into `bottomBarCourses` and `seeAllCourses`. After all, they logically belong to the same ordered list. We should only partition it inside `BottomBarTabView`, when the courses absolutely needed to be separated into the first few that can be fit into the tab and the rest that must appear in see all. Refactoring it to use a single list removes a lot of hard-to-reason code that if-else on whether a class belong to `bottomBarCourses` or `seeAllCourses`.

To support updating the bottom bar courses from everywhere, I setup another vue instance that acts as the scoped bottom bar courses storage and as a message bus. I decide not to put the data into the Vuex, since Vuex is mostly used for storing data that has strong dependency between each other, especially for requirements.

Setting up the shared bottom bar state makes one thing clear: we can easily update the state with 4 basic operations:

1. report color change of a firestore course
2. add a course to bottom bar
3. change course to focus
4. delete a course

Each of these basic operations is very easy to implement, when we have the data globally. Therefore, I implemented them inside `BottomBarState.ts`, and let all the other components call functions inside this file.

In the end, we have a much better codebase such that:

- `Dashboard.vue` is no longer cluttered with bottom bar update code.
- No more ugly deep props passing and event emitting related to bottom bar.
- Much cleaner bottom bar course logic without a lot of if-else.

### Test Plan 

- click to add courses to bottom bar. try to remove the courses and click on a tab to change focus.
- focus the last course in the bottom bar and check the focus has been transferred to previous course.
- resize the window to see the focused class between updated, when the original focused one is the 3rd or 4th.
- click on semester to see bottom bar collapses.

### Notes

As a result of this PR, 4 linter warnings related to mutating bottom bar related props have been eliminated!

### DIFF
--- a/src/components/BottomBar/BottomBarTabView.vue
+++ b/src/components/BottomBar/BottomBarTabView.vue
@@ -2,7 +2,7 @@
   <div class="bottombartabview">
     <div class="bottombartabview-bottomCourseWrapper">
       <div
-        v-for="(bottomCourse, index) in bottomCourses"
+        v-for="(bottomCourse, index) in firstFewCourses"
         :key="index"
         class="bottombartabview-courseWrapper"
       >
@@ -12,16 +12,14 @@
           :tabIndex="index"
           :bottomCourseFocus="bottomCourseFocus"
           :isExpanded="isExpanded"
-          @bottomBarTabToggle="bottomBarTabToggle"
-          @deleteBottomTab="deleteBottomTab"
-          @toggleFromTab="toggleFromTab"
-          @updateBarTabs="updateBarTabs"
+          @on-change-focus="() => changeBottomBarCourseFocus(index)"
+          @on-delete="() => deleteBottomBarCourse(index)"
         />
       </div>
     </div>
     <div v-if="seeMoreCourses.length > 0" class="bottombartabview-seeMoreWrapper">
       <div class="bottombarSeeMoreTab" @click="bottomBarSeeMoreToggle">
-        <div class="bottombarSeeMoreTab-name">{{ seeMoreString }}</div>
+        <div class="bottombarSeeMoreTab-name">See More</div>
         <img
           v-if="!seeMoreOpen"
           class="bottombarSeeMoreTab-arrow"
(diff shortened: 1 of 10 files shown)


src/containers/Dashboard.vue
```diff
@@ -15,7 +15,7 @@
           class="dashboard-nav"
           @editProfile="editProfile"
           @toggleRequirementsBar="toggleRequirementsBar"
-          :isBottomPreview="bottomBar.isPreview"
+          :isBottomPreview="false"
```
tcho6319, line 18: If `isBottomPreview` is always going to be `false`, can we just remove that prop and any of its dependencies from NavBar entirely?

src/components/BottomBar/BottomBarTabView.vue
```diff
@@ -70,97 +76,33 @@ export default Vue.extend({
...
-    isExpanded: { type: Boolean, required: true },
     maxBottomBarTabs: { type: Number, required: true },
   },
 
   computed: {
-    seeMoreString() {
-      return 'See More';
+    bottomCourseFocus(): number {
+      const focus = immutableBottomBarState.bottomCourseFocus;
+      return focus < this.maxBottomBarTabs ? focus : this.maxBottomBarTabs - 1;
```
tcho6319, line 85: I think the course focus may switch correctly to the right bottombartab but not the actual bottombarcourse content and title. When I select PE 1130, a course that's 3rd or 4th and minimize the screen width, shouldn't the course focus be on PAM 2100?

https://user-images.githubusercontent.com/54298311/108453811-37373780-7239-11eb-9bff-dfd15b35ac04.mov

tcho6319 (changes requested):
Thanks Sam! I found a bug and made a small suggestion.

src/containers/Dashboard.vue
SamChou19815: removed

tcho6319 (approved):
LGTM! Ty Sam

PR #302, opened 2021-02-18
Accurately compute new course modal options

SamChou19815:
### Summary 

This PR factors out from subroutine in requirement graph computation, and make the new course modal options generator call these subroutine to get an accurate list of options.

After this PR, all types of requirements can have a list of correct options, instead of just requirements that can be fulfilled by courses (as opposed to credits or toggleable).

In addition, during the generation of `RequirementWithIDSourceType`, I also patched the `allowDoubleCounting` field to make it allow double counting when the course is minor or second major. After this, reading this field will always give you the correct answer whether the requirement is double countable. Therefore, the bug of not classifying minor requirement as double-countable requirement is also fixed.

### Test Plan 

Counting minor requirement as double-countable (appear in auto-fulfill list)
minor

Now requirements fulfilled by credit sum also appears:
credits
As a comparison, this is master:
bad-master


src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -109,48 +108,29 @@ export default Vue.extend({
...
+          requirementsThatAllowDoubleCounting.push(it.name);
+        } else {
+          relatedRequirements.push(it);
         }
-      }
+      });
+
+      this.requirementsThatAllowDoubleCounting = requirementsThatAllowDoubleCounting;
+      this.relatedRequirements = relatedRequirements;
+      this.selfCheckRequirements = selfCheckRequirements;
```
tcho6319, line 132: More general question about double-countable-reqs: Currently, do we apply the course we want to add to all the `requirementsThatAllowDoubleCounting`? Do we apply the course to any `relatedRequirements`?

tcho6319 (changes requested):
Thanks Sam! I left a couple clarification questions and requested a small change.

1. Just a small change: Can you fix the typo in the filename of requirement-fronend-computation.ts?
2. Currently, the "Liberal Studies Distribution: 6 courses" and "Liberal Studies Distribution: 18 credits" subrequirements seem to be double-countable in the Sidebar but in the Add Modal, I can only select one of the two. Is that ok?

SamChou19815:
1 is fixed
2 is a real problem. The temporary fix before we have compound requirement is to mark the 18-credit option to allow double counting. However, I plan to do it in another PR, or we do it when we have the compound requirement infra.

src/components/Modals/NewCourse/NewCourseModal.vue
SamChou19815: >  Currently, do we apply the course we want to add to all the `requirementsThatAllowDoubleCounting`?

Yes, we don't even need to store the choices. According to the graph algo, we first build a coarse graph in phase 1 where all possible edges are connected. The edge between a course and a requirement that allow double counting will never be removed in phase 3.

> That is, would the course be found under all those double-countable subrequirements in the Sidebar at the moment

Yes, since the edge exists in the graph, the `GroupedRequirementFulfillment` will also contain the course, so it will appear in side bar somewhere.

> Similarly, do we apply the course to any `relatedRequirements`

We will eventually, when the requirement choices are saved and integrated into the algo.

tcho6319: Got it ty!

tcho6319 (approved):
LGTM! Ty for answering my questions.

PR #304, opened 2021-02-19
Implement Selectable Requirements

benjamin-shen:
### Summary 

This pull request implements selectable requirements and is a step towards implementing self check. See [this sprint](https://www.notion.so/User-Choice-Save-User-Choices-b531699659f94b10a54302b946980eaf).

- [x] set up [new collection on Firebase](https://console.firebase.google.com/u/0/project/cornelldti-courseplan-dev/firestore/data~2Fuser-selectable-requirement-choices) and integrate it with Vuex
- [x] save user data when adding new course

### Test Plan 

Add a random course, select requirement, and observe that the data is saved on Firebase. The expected behavior (for now) for duplicate courses is an overwrite.

Example:
image
image
image
image


SamChou19815 (commented):
Changes looks good. The package-lock change seems to be caused by your using npm v7. I think it would be better done in another standalone PR to remind everyone we are switching.

SamChou19815 (approved):

src/store.ts
```diff
@@ -75,6 +77,18 @@ const store: TypedVuexStore = new TypedVuexStore({
...
+      selectableRequirementChoices: AppSelectableRequirementChoices
+    ) {
+      state.selectableRequirementChoices = selectableRequirementChoices;
+    },
+    setUserRequirementChoices(
+      state: VuexStoreState,
+      selectableRequirementChoices: AppSelectableRequirementChoices
+    ) {
+      state.selectableRequirementChoices = selectableRequirementChoices;
+    },
```
tcho6319, line 91: What is the purpose of `setUserRequirementChoices(...)`? I don't think this is necessary since we have `setSelectableRequirementChoices(...)`. @SamChou19815 can you also confirm?

tcho6319 (changes requested):
Overall LGTM! Thanks Ben. Excited to see the Firestore collection update with user choices :). I left a comment about one of the mutation methods, but once it's addressed, this PR should be good to go!

src/store.ts
SamChou19815: ah yes, this seems to be duplicated

benjamin-shen: lol whoops that was back before we named it selectable requirements and I didn't catch it. Thanks

tcho6319 (approved):
LGTM! Ty :)

PR #306, opened 2021-02-19
Remove resolved TODO in eslint config

SamChou19815:
### Summary 

In #300 we decided that we will not change this to error, so let's remove this TODO to avoid future confusion.

### Test Plan 

👀


willespencer (approved):
Good call!

PR #307, opened 2021-02-19
Remove all v-bind

SamChou19815:
### Summary 

v-bind will destructure the object to pass down many props implicitly. This can be very confusing because it's unclear by reading the code on what has been passed. As a result, deciding whether some props are unused or whether some props passed in are always true/false becomes harder.

Therefore, I decide to remove all the v-bind once and for all.

### Test Plan 

It will pass type checking, which shows that all the needed props are still passed down.

Also tested that drag and drop still works.

### Notes 

I left a few v-bind because dragula still depends on it. Now that we removed dragula, they are finally safe to remove.


tcho6319 (approved):
LGTM!

PR #308, opened 2021-02-19
Avoid width hardcoding and unnecessary override in Course.vue css

SamChou19815:
### Summary 

The css in `Course.vue` is unnecessarily long. We hardcoded a lot of width values and override a lot of things that are exactly the same. 

I solve the width hardcoding by using the flexbox's `flex: 1 1 auto`. It ensures that the element will use up all the available space, so that we don't need to hardcode a width for all compact x screen-size combinations.

Then I just removed those repeated overrides.

### Test Plan 

side-by-side comparisons against master
Screen Shot 2021-02-19 at 13 15 23
Screen Shot 2021-02-19 at 13 15 36
Screen Shot 2021-02-19 at 13 15 45
Screen Shot 2021-02-19 at 13 15 50

### DIFF
--- a/src/components/Course.vue
+++ b/src/components/Course.vue
@@ -179,7 +179,7 @@ export default Vue.extend({
     justify-content: center;
 
     &--active {
-      width: 19px;
+      width: 1.188rem;
     }
   }
 
@@ -282,7 +282,7 @@ export default Vue.extend({
     }
     &-color {
       &--active {
-        width: 19px;
+        width: 1.188rem;
       }
     }
 
--- a/src/components/Course.vue
+++ b/src/components/Course.vue
@@ -1,18 +1,12 @@
 <template>
   <div :class="{ 'course--min': compact, active: active }" class="course" @click="courseOnClick()">
-    <div
-      class="course-color"
-      :style="cssVars"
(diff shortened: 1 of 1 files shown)


src/components/Course.vue
```diff
@@ -188,15 +180,6 @@ export default Vue.extend({
 
     &--active {
       width: 19px;
```
tcho6319, line 182: Can we convert this to rem?

src/components/Course.vue
```diff
@@ -320,53 +280,9 @@ export default Vue.extend({
...
-      width: 1.25rem;
-      height: 5.625rem;
-      border-radius: 0.42rem 0 0 0.42rem;
-      background-color: var(--bg-color);
-      display: flex;
-      align-items: center;
-      justify-content: center;
-
       &--active {
         width: 19px;
```
tcho6319, line 285: Same here

tcho6319 (changes requested):
Thanks Sam! Just requested that we make margins in rem. Other than that, the PR should be good to go.

src/components/Course.vue
SamChou19815: @tcho6319 fixed

tcho6319 (approved):
LGTM. Ty

PR #309, opened 2021-02-19
[easy] Inline Dashboard tour window text

SamChou19815:
### Summary 

Inline the string props passed into tour-window, so that the intention is clearer, and avoid blouting the number of states in `Dashboard.vue`

### Test Plan 

👀


tcho6319 (approved):
LGTM! Ty

PR #310, opened 2021-02-19
Remove subRequirementProgress field when req is fulfilled by credits

SamChou19815:
### Summary 

For all the requirements that are fulfilled by credits, it's always the case that there is only one checker produces eligible courses, and we want to count all the course credits from that list. Therefore, it's always `any-can-count`. (There are several that uses `every-course-needed`, but the credit summation code for `any-can-count` and `every-course-needed` is the same anyways...)

Thus, it doesn't make sense to store the field, and leaving it there might be very confusing. This diff removes them.

### Test Plan 

Find a requirement that uses credits (e.g. engineering liberal arts 18 credits) and check its progress is still computed correctly.


src/requirements/data/majors/bme.ts
```diff
@@ -63,7 +63,7 @@ const biomedicalEngineeringRequirements: readonly CollegeOrMajorRequirement[] =
           ['BME 4190'],
           ['BME 5830', 'BME 5850', 'BTRY 4381', 'BTRY 4840', 'CHEM 4810', 'CHEME 5430', 'CS 4780']
         ),
-        counting: 'credits',
+        counting: 'courses',
```
SamChou19815, line 66: you can read the description above that it's very clear it should be courses

src/requirements/requirement-frontend-computation.ts
```diff
@@ -184,14 +183,9 @@ function computeFulfillmentStatisticsFromCourses(
...
-          subRequirementProgress === 'any-can-count'
-            ? coursesThatFulfilledSubRequirement
-                .map(course => course.credits)
-                .reduce((a, b) => a + b, 0)
-            : coursesThatFulfilledSubRequirement
-                .map(course => course.credits)
-                .reduce((a, b) => a + b);
+        minCountFulfilled += coursesThatFulfilledSubRequirement
+          .map(course => course.credits)
+          .reduce((a, b) => a + b, 0);
```
SamChou19815, line 188: the code is the same, so only keep one

src/requirements/data/majors/comm.ts
```diff
@@ -61,51 +53,50 @@ const commRequirements: readonly CollegeOrMajorRequirement[] = [
       'A student may elect to fulfill 3 of these credit hours by taking a third focus area introductory course. ' +
       'A maximum of 3 credits in either 4970 or 4990 (combined) can be counted toward the upper level major requirements. Refer to the Course and Time Roster for the most up-to-date offerings. ',
     source: 'https://communication.cals.cornell.edu/undergraduate-program/major-requirements/',
-    checker: includesWithSubRequirements(
-      ['COMM 2179'],
-      ['ANTHR 4330', 'COGST 4330', 'STS 4330'],
-      ['COML 2703', 'ENG 2703', 'MUSIC 2703', 'PMA 2703'],
```
SamChou19815, line 67: I only kept one, because i found out this list was all the cross listed courses.

src/requirements/data/majors/comm.ts
```diff
@@ -1,33 +1,26 @@
...
+      'COMM 1101',
+      'COMM 1300',
+      'COMM 2010',
+      'COMM 2310',
+      'COMM 2820'
     ),
-    subRequirementProgress: 'every-course-needed',
     fulfilledBy: 'credits',
     minCount: 15,
   },
```
tcho6319, line 18: I wonder if this could be `fulfilledBy: 'courses'`. Since it seems you would have to take every course they list.

src/requirements/data/majors/comm.ts
```diff
@@ -1,33 +1,26 @@
...
-      ['COMM 2200'],
-      ['COMM 2450'],
-      ['COMM 2760'],
-      ['COMM 2850']
-    ),
-    subRequirementProgress: 'any-can-count',
+    checker: includesWithSingleRequirement('COMM 2200', 'COMM 2450', 'COMM 2760', 'COMM 2850'),
     fulfilledBy: 'credits',
     minCount: 6,
   },
```
tcho6319, line 26: Similarly, can we do `fulfilledBy: 'courses'`? I can see why we chose credits but just wanted to hear your thoughts.

src/requirements/data/majors/comm.ts
SamChou19815: @tcho6319 I updated it, but I checked the website again and the requirement seems to change already. We need to make a pass over all requirements data before launch.

tcho6319 (changes requested):
minCount seems to be off for the Communication major's Core Courses subreq. Once that's fixed, the PR should be good to go!

src/requirements/data/majors/comm.ts
```diff
@@ -13,23 +13,18 @@ const commRequirements: readonly CollegeOrMajorRequirement[] = [
       ['COMM 2310'],
       ['COMM 2820']
     ),
+    fulfilledBy: 'courses',
     subRequirementProgress: 'every-course-needed',
-    fulfilledBy: 'credits',
     minCount: 15,
```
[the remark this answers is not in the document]
SamChou19815: fixed

tcho6319 (approved):
Ty! LGTM

PR #311, opened 2021-02-20
Implement wrong semester and double counting warning

SamChou19815:
### Summary 

This PR first setup infra to support multiple warnings in the course caution component and a base `CourseBaseTooltip.vue` to support info tooltip later.

Then I moved some derived course data into the Vuex store, so that relevant info for course warnings can be elegantly subscribed.

The final step is mostly straightforward: adding the relevant text and logic in `CourseCaution.vue`. The double counting warning doesn't have relevant text in design yet, so I just come up with my own words.

### Test Plan 

Duplicate warning as usual:
duplicate

Wrong semester warning:
wrong-semester

Multiple warnings in one place:
multiple

#### Missing Prereq

I didn't implement the missing prereq warning, since it's almost impossible to find out the prereq. Here is an example showing a prereq with arbitrary text:
tricky-prereq

#### Another course fulfills the same requirement

Here is the design that I have not implemented yet:
Screen Shot 2021-02-20 at 15 59 59

From my understanding, this is warning on a requirement being "over-fulfilled". I'm not sure how useful it is, and it might be too noisy if the user is just very passionate about something. It can become extremely noisy in some cases. Let's suppose the user takes 7 liberal studies class instead of 6, then all 7 courses will show this warning :(

### DIFF
--- a/src/components/Course/Course.vue
+++ b/src/components/Course/Course.vue
@@ -17,7 +17,7 @@
           <span v-if="!compact && semesterString" class="course-semesters">{{
             semesterString
           }}</span>
-          <course-caution v-if="cautionString" :compact="compact" :cautionString="cautionString" />
+          <course-caution v-if="!compact" :course="courseObj" />
         </div>
       </div>
     </div>
@@ -38,7 +38,7 @@
 <script lang="ts">
 import Vue, { PropType } from 'vue';
 import CourseMenu from '@/components/Modals/CourseMenu.vue';
-import CourseCaution from '@/components/CourseCaution.vue';
+import CourseCaution from '@/components/Course/CourseCaution.vue';
 import {
   addCourseToBottomBar,
   reportCourseColorChange,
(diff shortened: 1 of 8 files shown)


src/components/Course/Course.vue
```diff
@@ -67,9 +70,20 @@ export default Vue.extend({
     };
   },
   computed: {
-    cautionString(): string | null {
-      if (this.duplicatedCourseCodeList == null) return null;
-      return this.duplicatedCourseCodeList.includes(this.courseObj.code) ? 'Duplicate' : null;
+    courseCautions(): CourseCautions {
+      return getCourseCautions(this.courseObj);
+    },
```
tcho6319, line 75: Can we not move `getCourseCautions(...)` to Course.vue? Why did you keep it in CourseCaution.vue?

SamChou19815: @tcho6319  Now I managed to keep it within the CourseCaution completely. Now you only need to pass in the course object and it can derive all the warnings.

tcho6319 (approved):
LGTM. Thanks Sam. I'm good with approving this PR. I think you bring up good points re: the warning for courses applying to the same req. I think that could clutter the tooltip / every course might end up having some warning. @einc what are your thoughts?

PR #312, opened 2021-02-21
Initial Self-Check frontend implementation

willespencer:
### Summary 

Started the frontend implementation of the self-check requirements on the sidebar as per this [notion](https://www.notion.so/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=7a3e7595daa74cc992487adc2d679e31) item.

- [x] Added a dropdown to self-checks populated with courses on the user's schedule
- [x] Selecting a course displays a completed requirement
- [x] Resetting a requirement has it disappear from the plan and from self check
- [x] Added a dropdown item to add courses from a new condensed add modal
- [x] Removed semester select from the old add modal entirely
- [x] Factored out a bunch of shared functions

image

Remaining TODOs in the sprint over the next 2 weeks.

- [ ] Update dropdown styling. I copied the toggleable requirements dropdown but both could use an improvement to look more like the Figma designs. Also need to add some more text ("Add Course") and margins.
- [ ] Only show courses that are not mapped to requirements in the dropdown
- [ ] Map courses to self-check requirements so they actually fulfill them
- [ ] Store requirement choices on firebase so they last after refresh (see #304)
- [ ] Limit the self-check dropdown from appearing after a course/credit limit for self-checks, need data for this first @einc 
- [ ] Resolve codebase changes, see questions in Notes section

### Test Plan 

1. Add an existing course to a self-check requirement, ensure that it shows up as completed
2. Add a new course from the self-check add modal
3. Reset both types of courses and ensure they are removed from the plan and sidebar
4. Ensure other subreqs (like toggleable requirements) still work as expected
5. Ensure adding from the original add modal still works as expected

### Notes 

Have a few changes I made to the codebase that I am unsure about...
- I duplicated a lot of adding course code to requirements sidebar files. I noticed later that there was already some unused code to call the add modal (the $parent.$parent etc). Should I use that existing $parent method, keep the code I duplicated and remove the unused code, or try to make some of the code I copied global so it is not duplicated anywhere?
- Design question: If either of the last two options for the above is picked, then I made some changes to the existing toggleable dropdown styling to not show up above the sidebar to get the add modal to work properly. Can I leave it like it is now, or should I try and find a way to change it back.
- I also duplicated some dropdown code & styling from toggleable requirements. I decided not to try to make it more accessible or anything since the code already existed, and we talked about maybe trying to make a dropdown component. Is this ok or should I try something different?
- @SamChou19815 I noticed that courses from semesters are a different type (FirestoreSemesterCourse) vs the ones mapped to requirements (CourseTaken). I ended up updating CompletedSubReqCourse to handle both types, but I was not sure if I was doing something wrong or should somehow be using the same type as before.


src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -109,6 +110,10 @@ export default Vue.extend({
   methods: {
     selectCourse(result: MatchingCourseSearchResult) {
       this.selectedCourse = result;
+      if (this.isSelfCheck) {
+        this.addCourse();
+        return;
+      }
```
SamChou19815, line 116: This is a smart hack, but I think it's a better idea to create a specialized new course modal (e.g. `<add-course-for-self-check-modal />`) that just let you choose courses and semester. 

This can be easily achieved by copy-pasting all the code in `NewCourseModal` and then remove the code you don't need (in this case it's mostly the requirement editor code). This approach will create some duplicated code, but I think it's worth it, because then the `NewCourseModal` don't need to worry about editing semester at all.

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -0,0 +1,260 @@
...
+      //   this.$refs.modal.courseIsAddable = true;
+      //   this.courses.forEach(course => {
+      //     if (course.code === key) {
+      //       // @ts-ignore
+      //       this.$refs.modal.courseIsAddable = false;
+      //       this.$emit('open-caution-modal');
+      //     }
+      //   });
+      // }
+    },
```
SamChou19815, line 118: This code should be removed, and probably also the place where you copied it. I think it is pretty dead. The new course warning implementation doesn't require this code.

SamChou19815 (changes requested):
> I duplicated a lot of adding course code to requirements sidebar files. I noticed later that there was already some unused code to call the add modal (the $parent.$parent etc). Should I use that existing $parent method, keep the code I duplicated and remove the unused code, or try to make some of the code I copied global so it is not duplicated anywhere?

I believe the `$parent` code in `IncompleteSubReqCourse` should be removed now. In your PR you ensured that that code will never be called, so I think it's good to go now.

 > I also duplicated some dropdown code & styling from toggleable requirements. I decided not to try to make it more accessible or anything since the code already existed, and we talked about maybe trying to make a dropdown component. Is this ok or should I try something different?

Looks good to me.

> I noticed that courses from semesters are a different type (FirestoreSemesterCourse) vs the ones mapped to requirements (CourseTaken). I ended up updating CompletedSubReqCourse to handle both types, but I was not sure if I was doing something wrong or should somehow be using the same type as before.

Commented in review

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -0,0 +1,260 @@
...
+      }
+      if (a.year < b.year) {
+        return 1;
+      }
+      if (SeasonsEnum[a.type] < SeasonsEnum[b.type]) {
+        return 1;
+      }
+      return -1;
+    },
+  },
```
tcho6319, line 160: Maybe we can put the duplicated functions in a utils file. I've also seen `SeasonsEnum` several times in the codebase.

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -0,0 +1,260 @@
...
+    height: 100%;
+  }
+  &-arrow {
+    border-left: 6.24px solid transparent;
+    border-right: 6.24px solid transparent;
+    border-top: 6.24px solid $inactiveGray;
+    background: transparent;
+    margin-right: 8.7px;
+    margin-left: 5px;
+    margin-top: 5px;
```
tcho6319, line 209: Can we have margins in rem?

tcho6319 (changes requested):
Excited for the Self-Check! Overall LGTM. I agree with Sam's comments and made some additional suggestions. The main change that I would like is for there to be a new component for the Self Check Add Modal. 

Also, @hahnbeelee  will be working on an accessible dropdown modal next sprint so this PR can just keep the duplicated code for now.

src/components/Requirements/IncompleteSelfCheck.vue
willespencer: It looks like these exact pixel numbers are used within 4 different dropdowns currently. I agree they should be in rem, but maybe we should update them if/when we factor out dropdown?

src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -92,9 +61,6 @@ export default Vue.extend({
       year: 0,
```
SamChou19815, line 61: they can also be removed now, and then Semester.vue no longer needs to handle addCourse emit based on whether season and year having dummy value!

SamChou19815 (approved):
Looks reasonable, just one nit

src/components/Modals/NewCourse/NewSelfCheckCourseModal.vue
```diff
@@ -0,0 +1,152 @@
...
+    color: $lightPlaceholderGray;
+    margin-bottom: 6px;
+  }
+}
+
+.content-course {
+  width: 27.75rem;
+}
+
+@media only screen and (max-width: 600px) {
```
tcho6319, line 147: Can you replace 600px with the breakpoint variable? I believe it's in `src/assets/scss/_variables.scss`

tcho6319 (changes requested):
Thanks Will! Left a small nit and then wanted to point out a small styling discrepancy (below). Once these are addressed, this PR should be good to go!

The height of the Sidebar doesn't seem to be the entire height of the screen. I first show the staging site for this PR, and then master staging site in my screen recording.
https://user-images.githubusercontent.com/54298311/108745856-9d59de00-7509-11eb-8504-42caa2640cf8.mov

einc:
for the test plan, if I add a class for self check, then remove from the schedule, is it supposed to get cleared? (or is this after storing to firebase i wasn't sure)

https://user-images.githubusercontent.com/14082828/108754921-94223e80-7514-11eb-9770-a6d05d5a9840.mov

SamChou19815:
> for the test plan, if I add a class for self check, then remove from the schedule, is it supposed to get cleared? (or is this after storing to firebase i wasn't sure)
> 
>  will.mov

I think it's very hard to implement with the current structure without global data, but it will become much more straightforward after using the global data, so it's better to save it to a future diff.

willespencer:
> Thanks Will! Left a small nit and then wanted to point out a small styling discrepancy (below). Once these are addressed, this PR should be good to go!
> 
> The height of the Sidebar doesn't seem to be the entire height of the screen. I first show the staging site for this PR, and then master staging site in my screen recording.
> 
>  Screen.Recording.2021-02-22.at.12.28.41.PM.mov

Thanks for pointing this out! I just fixed this in my newest commit, along with an issue I found where the screen was sometimes horizontal scrollable and the rightmost edge did not have the background color (which also exists on a less sever extent on the staging site). Can you make sure both seem fixed to you, since I was messing around with a lot of different CSS props?

tcho6319 (approved):
LGTM! Thanks Will :D

tcho6319 (approved):
Oops I missed this the second time around lol. LGTM!

PR #316, opened 2021-02-21
[easy] Fix linter warnings and strictify linter config

SamChou19815:
### Summary 

Deleted some dead code that are guaranteed to be useless, since the correct course warning code has been merged in.

As a result, we removed all the `@ts-ignore`, so I strictify the level to `error`. In the future, every ts error suppression must be `@ts-expect-error: your reason why this is needed`.

### Test Plan 

New course modal can still open and work.

### DIFF
--- a/src/components/Modals/NewCourse/NewCourseModal.vue
+++ b/src/components/Modals/NewCourse/NewCourseModal.vue
@@ -88,7 +88,6 @@ export default Vue.extend({
       selfCheckRequirements: [] as readonly RequirementWithIDSourceType[],
       editMode: false,
       courseSelectorKey: 0,
-      courseIsAddable: true,
       season: '' as FirestoreSemesterType,
       year: 0,
     };
@@ -145,10 +144,6 @@ export default Vue.extend({
       this.reset();
       this.$emit('close-course-modal');
     },
-    // Note: Currently not used
-    checkCourseDuplicate(key: string) {
-      this.$emit('check-course-duplicate', key);
-    },
     addItem() {
       if (this.editMode) {
         this.editMode = false;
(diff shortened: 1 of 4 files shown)


src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -169,10 +164,8 @@ export default Vue.extend({
           resultJSON.data.classes.forEach((resultJSONclass: CornellCourseRosterCourse) => {
             if (resultJSONclass.catalogNbr === number) {
               const course = { ...resultJSONclass, roster };
-              if (this.courseIsAddable) {
-                this.$emit('add-course', course, this.season, this.year);
-                this.chooseSelectableRequirementOption(course.crseId.toString(), requirementID);
-              }
+              this.$emit('add-course', course, this.season, this.year);
+              this.chooseSelectableRequirementOption(course.crseId.toString(), requirementID);
```
SamChou19815, line 168: ping @willespencer, since you copy pasted some code from here.

src/requirements/requirement-frontend-computation.ts
```diff
@@ -225,8 +225,7 @@ function filterAndPartitionCoursesThatFulfillRequirement(
 function computeFulfillmentCoursesAndStatistics(
   requirement: RequirementWithIDSourceType,
   coursesTaken: readonly CourseTaken[],
-  toggleableRequirementChoices: AppToggleableRequirementChoices,
-  selectableRequirementChoices: AppSelectableRequirementChoices
+  toggleableRequirementChoices: AppToggleableRequirementChoices
```
SamChou19815, line 228: this function doesn't need `selectableRequirementChoices`.

src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -83,7 +83,6 @@ export default Vue.extend({
       selfCheckRequirements: [] as readonly RequirementWithIDSourceType[],
       editMode: false,
       courseSelectorKey: 0,
-      courseIsAddable: true,
```
SamChou19815, line 86: The only places where `courseIsAddable` is ever set to false are the deleted uncalled code. Now that they are gone, we know it's always true.

hahnbeelee (approved):
LGTM

PR #317, opened 2021-02-21
[RFC] Generate readable requirement json

SamChou19815:
### Summary 

Right now, the generated requirement json is collapsed into a single line. It is almost completely unreadable. This is bad for code review, because we don't know what's changed there. Now the only way I can convince you that something related to requirement data is correct without asking you to run code is by arguing that a change produces no changes in generated requirement json, which is not very common.

By producing a human readable json, we can see the diff in github review page, so everyone can have a better idea on what is changed.

There might be some concerns of creating such uncompressed json will cause the user to get larger bundle file, but it's not a problem, because webpack will bundle everything compressed anyways.

### Test Plan 

A screenshot showing the bundled js file that proves my claim:

Screen Shot 2021-02-21 at 15 31 02


.gitattributes
```diff
@@ -1,2 +1,3 @@
 # Auto detect text files and perform LF normalization
 * text=auto
+src/requirements/decorated-requirements.json linguist-generated=true
```
SamChou19815, line 3: mark this big file as generated so that github doesn't classify our repo as a json repo lol

tcho6319 (approved):
I'd be ok with this especially if the json would be compressed anyway for the user. A more readable json would also give devs more confidence that their changes were intentional.

PR #320, opened 2021-02-22
Adjust liberal arts requirement data for 2020 change

SamChou19815:
### Summary 

There was a big liberal arts changes in arts and engineering. Although not all colleges are impacted by this policy, they do impact the course json we receive from roster API. This PR fixes all the related issues due to this change.

Reasons for changes will be explained in PR code comments directly.

Depend on #317.

### Test Plan 

👀


src/requirements/data/colleges/as.ts
```diff
@@ -146,11 +149,11 @@ const casRequirements: readonly CollegeOrMajorRequirement[] = [
...
-      (course: Course): boolean => course.catalogDistr?.includes('(CA-AS)') ?? false,
-      (course: Course): boolean => course.catalogDistr?.includes('(HA-AS)') ?? false,
-      (course: Course): boolean => course.catalogDistr?.includes('(KCM-AS)') ?? false,
-      (course: Course): boolean => course.catalogDistr?.includes('(LA-AS)') ?? false,
-      (course: Course): boolean => course.catalogDistr?.includes('(SBA-AS)') ?? false,
+      (course: Course): boolean => course.catalogDistr?.includes('CA-AS') ?? false,
+      (course: Course): boolean => course.catalogDistr?.includes('HA-AS') ?? false,
+      (course: Course): boolean => course.catalogDistr?.includes('KCM-AS') ?? false,
+      (course: Course): boolean => course.catalogDistr?.includes('LA-AS') ?? false,
+      (course: Course): boolean => course.catalogDistr?.includes('SBA-AS') ?? false,
```
SamChou19815, line 156: Due to the additional of new categories, there might be multiple items between the parenthesis.
e.g. [ASIAN 2212](https://classes.cornell.edu/browse/roster/SP21/class/ASIAN/2212) originally only has `(CA-AS)`, now it's `(CA-AS, ALC-AS, HST-AS)`. Therefore, we should not include parenthesis in the matching. Similar changes are made elsewhere.

src/requirements/data/colleges/en.ts
```diff
@@ -5,6 +5,23 @@ import {
...
+  'KCM',
+  'SBA',
+  'FL',
+  'CE',
+  'ALC',
+  'SCD',
+  'HST',
+  'ETM',
+  'SSC',
+  'GLC',
```
SamChou19815, line 22: Added these new categories. It turns out that 2019 and 2020 requirements for engineering is exactly the same...
https://www.engineering.cornell.edu/students/undergraduate-students/advising/liberal-studies/liberal-studies-policy-students-entering
https://www.engineering.cornell.edu/students/undergraduate-students/advising/liberal-studies/liberal-studies-policy-students-entering-0
not sure why they made two separate pages...

src/requirements/data/colleges/en.ts
```diff
@@ -80,33 +97,47 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
...
+    source:
+      'https://www.engineering.cornell.edu/students/undergraduate-students/advising/liberal-studies',
+    checker: engineeringLiberalArtsDistributions.map(distribution => (course: Course): boolean =>
+      course.catalogDistr?.includes(distribution) ?? false
+    ),
+    subRequirementProgress: 'every-course-needed',
+    fulfilledBy: 'courses',
+    allowCourseDoubleCounting: true,
+    minCount: 3,
+  },
```
SamChou19815, line 128: also added this 3 category checker

src/requirements/data/colleges/as.ts
```diff
@@ -10,7 +10,10 @@ const casRequirements: readonly CollegeOrMajorRequirement[] = [
       'they can also take all their credits in Arts & Sciences and accumulate more than 120. ' +
       'Note: AP, IB, and A-Level credits count toward the 120 total credits but not toward the 100 A&S credits.',
     source: 'https://as.cornell.edu/degree-requirements',
-    checker: (course: Course): boolean => course.catalogDistr?.includes('-AS') ?? false,
+    checker: (course: Course): boolean =>
+      course.acadGroup.includes('AS') ||
+      course.subject === 'CS' ||
+      (course.catalogDistr?.includes('-AS') ?? false),
```
SamChou19815, line 16: It was originally `course.acadGroup.includes('AS')`, but it was changed to the current code in #115. I don't think we should remove the `acadGroup` check. I also special cased for CS, since CS is both in arts and engineering. Not sure whether it's 100% correct...

src/requirements/data/colleges/en.ts
tcho6319: To confirm: This checker doesn't check if the courses fall under 3 distinct categories right?

tcho6319: Just wanted to point out that it seems a single course that includes multiple distribution categories may "fulfill" this requirement. Is that ok? I believe that a post-launch goal is to be able to have category filters where the user can select a certain category (e.g. HA) for a course. So this 3 category subreq coupled with the other liberal arts subreqs may be ok for launch.

https://user-images.githubusercontent.com/54298311/108725903-d8054b80-74f4-11eb-80bc-9f045f72731f.mov

tcho6319 (commented):
Thanks Sam. Left a comment about the 3 category checker, but I would also get @einc's approval on this. If she's ok with it, then I'd be good to approve!

src/requirements/data/colleges/en.ts
einc: also why are there add courses for up to 14? 
Is this because of the 13 categories?

SamChou19815: Yes, it's mostly for 3 categories. It's 14 instead of 13 due to LA vs LAD, but I can try to condense them into one to make it more intuitive.

einc: maybe for now, we could just count the first category a class fulfills for this req?
if we need to keep track of 3 separate categories, would this require knowing which categories the user already has in their schedule, checking if its new, and if so counting it?

einc (approved):
Yes, looks good to me too and agree with Theresa's comment!

PR #321, opened 2021-02-22
Simplify firestore course by avoiding storing data only for bottom bar

SamChou19815:
### Summary 

Right now our `FirestoreSemesterCourse` stores a lot of information. A lot of information is only needed for bottom bar. However, the bottom bar requires fetching data from CUReviews, so some data will be unavailable immediately anyways, so pre-store those additional data doesn't make sense.

Removing those data for just bottom bar not only reduces our db storage, but also reduces the amount of course data we need to store as static course data. In the next diff, I will directly show the benefit of cutting down this large chunk of data.

### Test Plan 

Add a new course, and click on the course to open the bottombar. Check that the data can still be displayed correctly.


src/components/BottomBar/BottomBarState.ts
```diff
@@ -24,6 +29,21 @@ export const reportCourseColorChange = (courseUniqueID: number, color: string):
...
+const getDetailedInformationForBottomBar = async (
+  roster: string,
+  subject: string,
+  number: string
+) => {
+  const courses: readonly CornellCourseRosterCourse[] = (
+    await fetch(
+      `https://classes.cornell.edu/api/2.0/search/classes.json?roster=${roster}&subject=${subject}`
+    ).then(response => response.json())
+  ).data.classes;
```
SamChou19815, line 41: I decide to fetch the data from cornell roster ondemand, because then we don't need to store those data on our own.

src/components/Course/Course.vue
```diff
@@ -80,15 +80,6 @@ export default Vue.extend({
...
     },
-    review(): string {
-      const [subject, number] = this.courseObj.code;
-      return `https://www.cureviews.org/course/${subject}/${number}`;
-    },
-
-    roster(): string {
-      const [subject, number] = this.courseObj.code;
-      return `https://classes.cornell.edu/browse/roster/FA18/class/${subject}/${number}`;
-    },
```
SamChou19815, line 91: unused

src/assets/courses/rosters.json
```diff
@@ -1 +1 @@
-["FA14","WI15","SP15","SU15","FA15","WI16","SP16","SU16","FA16","WI17","SP17","SU17","FA17","WI18","SP18","SU18","FA18","WI19","SP19","SU19","FA19","WI20","SP20","SU20","FA20","WI21","SP21","SU21"]
\ No newline at end of file
+["WI17","SP17","SU17","FA17","WI18","SP18","SU18","FA18","WI19","SP19","SU19","FA19","WI20","SP20","SU20","FA20","WI21","SP21","SU21"]
```
tcho6319, line 1: Can we add a newline at the end of this file?

src/components/BottomBar/BottomBarState.ts
tcho6319: Do you think we would ever run the risk of hitting the API too frequently once we launch? I want to make sure that this would be ok in the long term.

tcho6319 (changes requested):
Thanks Sam! I requested a small change for newline and left a comment about using the Cornell Course Roster API.

src/components/BottomBar/BottomBarState.ts
SamChou19815: No, this one use is the exactly the one we hit rarely. Bottom bar isn't open all the time, and course data is not refetched all the time.

src/assets/courses/rosters.json
SamChou19815: @tcho6319 new line added

tcho6319 (approved):
LGTM ty Sam!

PR #322, opened 2021-02-22
Generate full-courses.json

SamChou19815:
### Summary 

Depends on #321.

According to our meeting on Sunday, it's better to use only the data from the last offering, because it reduces potential user confusion without losing much accuracy.

This PR computes a new list of courses that is the same as all the old filtered courses, but only keep the data from the last offering. Currently, this json is 3.5M when condensed on a single line. However, it's important to note that it's not the raw size that matters, it's the compressed size. I ran `npm run build` to show that the final bundle, when gzipped, still has reasonable size:

Screen Shot 2021-02-22 at 17 35 40

### Test Plan 

It shouldn't change any prod behavior. So see that everything still passes on CI.

### Notes

In the next diff, I will use the courses with latest offering data to compute requirements, and use the new `full-courses.json` to get new information instead of `FetchCourses` function!


src/requirements/courses-json-generator.ts
```diff
@@ -10,17 +11,26 @@ export type CourseJson = {
...
+  courseCodeToDataAndOfferedMap.forEach(([roster, course]) => {
+    const existingArray = json[course.crseId] || [];
+    existingArray.push({ ...course, roster });
+    json[course.crseId] = existingArray;
+  });
+
+  return json;
+})();
+
 writeFileSync('src/assets/courses/courses.json', JSON.stringify(courseJson));
```
tcho6319, line 35: Is courseJson used for our course selector dropdown?

tcho6319 (commented):
LGTM! Ty Sam :D I left a clarification question, but I support these changes. `FetchCourses()` by default "fetches" course data from the most recent course offering anyway, so `full-courses.json` seems to be a more lightweight solution! I am going to leave my review as a comment for now, but if @einc approves, I also approve.

src/requirements/courses-json-generator.ts
SamChou19815: Yes, but with full courses it might be needed anymore. I just keep it here to make the PR more focused

einc (approved):

PR #323, opened 2021-02-23
Make requirement computation use only course data with latest offering

SamChou19815:
### Summary 

Depends on #322.

This PR implements my proposal to use only the latest course data to compute requirement fulfillment. Concretely, this means use the courses with the latest offering generated in #322, and completely disregard the semester a course is in.

In addition, using the `full-courses.json` generated in #322, we can avoid fetching data using firebase functions, since `full-courses.json` is small enough to be included in the bundle. Therefore, I cleaned up a lot of code related to data fetching, and make the flow of incomplete-sub-req courses and see all courses much simpler. (This may have auto-fixed the see all courses bug mentioned on notion, but I have not verified it yet.)

### Test Plan 

Requirement computation is consistent, and incomplete requirement courses can still be displayed :
Screen Shot 2021-02-23 at 00 26 52

Verify that dragging requirement courses into semester still works.


functions/index.js
```diff
@@ -95,244 +88,3 @@ exports.TrackUsers = functions.https.onRequest(async (req, res) => {
...
-    // Fetch any course objects with the provided crseIds for this roster
-    crseIds = cleanCrseIds(crseIds, roster);
-    const fetchedCoursesWithCrseIds = fetchCoursesWithCrseIds(roster, crseIds, fetchedCoursesSoFar);
-    fetchedCourses = fetchedCourses.concat(fetchedCoursesWithCrseIds);
-  });
-
-  return {
-    courses: fetchedCourses,
-  };
-});
```
SamChou19815, line 338: we no longer need fetch courses! bye!

src/components/Requirements/SubRequirement.vue
```diff
@@ -194,8 +184,39 @@ export default Vue.extend({
...
+            });
+          }
+        } else {
+          slots.push({
+            isCompleted: false,
+            courses: generateSubReqIncompleteCourses(allTakenCourseIds, subReqEligibleCourses[i]),
+          });
+        }
+      });
+      return slots;
```
SamChou19815, line 219: we no longer need to batch fetch data, so now the resulting flow is much cleaner!

src/requirement-types.d.ts
```diff
@@ -54,10 +49,9 @@ type RequirementFulfillmentInformation<T = Record<string, unknown>> =
     };
 
 type DecoratedCollegeOrMajorRequirement = RequirementCommon &
-  RequirementFulfillmentInformation<{ readonly courses: readonly EligibleCourses[] }>;
+  RequirementFulfillmentInformation<{ readonly courses: readonly (readonly number[])[] }>;
 
 type CourseTaken = {
-  readonly roster: string;
```
tcho6319, line 60: To confirm: We don't need the `roster` field for `CourseTaken` type because we are getting courses from most latest offering right?

src/global-firestore-data.ts
```diff
@@ -91,8 +91,11 @@ export const chooseSelectableRequirementOption = (
     .set(selectableRequirementChoices);
 };
 
-const getSubjectColor = (subjectColors: Record<string, string>, subject: string): string => {
-  if (subjectColors[subject]) return subjectColors[subject];
+const getSubjectColor = (
+  subjectColors: Record<string, string>,
+  subject: string
+): readonly [string, boolean] => {
+  if (subjectColors[subject]) return [subjectColors[subject], false];
```
tcho6319, line 98: Nit: Can you write a // comment that explains what the boolean means? I know that it indicates whether or not the color was changed, but a comment in the code might be clearer for other devs. Your call.

src/components/Requirements/SubRequirement.vue
```diff
@@ -138,46 +137,47 @@ type CompletedSubReqCourseSlot = {
...
+  eligibleCourseIds: readonly number[]
+): readonly FirestoreSemesterCourse[] => {
+  const rosterCourses = eligibleCourseIds
+    .filter(courseID => !allTakenCourseIds.has(courseID))
+    .flatMap(courseID => fullCoursesJson[courseID] || []);
+  const subjectColors = allocateSubjectColors(new Set(rosterCourses.map(it => it.subject)));
+  return rosterCourses.map(rosterCourse =>
+    cornellCourseRosterCourseToFirebaseSemesterCourseWithCustomIDAndColor(
+      rosterCourse,
+      -1,
```
tcho6319, line 162: Is the -1 here a dummy unique id for a course in the incomplete sub req course slot?

tcho6319 (approved):
LGTM! Thanks Sam. I just left several clarification comments and a small nit.

functions/index.js
```diff
@@ -47,20 +40,20 @@ function isOld(semester) {
...
-  var semester = [];
-  var oldSemester = [];
-  var newSemester = [];
-  var semesterCount = 0;
+  const arr = [];
+  let count = 0;
+  const semester = [];
+  const oldSemester = [];
+  const newSemester = [];
+  const semesterCount = 0;
```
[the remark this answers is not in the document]
SamChou19815: We want to probably keep the current deployment for a while so that FetchCourses still works for devs who haven't merged in latest master.

src/requirement-types.d.ts
```diff
@@ -54,10 +49,9 @@ type RequirementFulfillmentInformation<T = Record<string, unknown>> =
     };
 
 type DecoratedCollegeOrMajorRequirement = RequirementCommon &
-  RequirementFulfillmentInformation<{ readonly courses: readonly EligibleCourses[] }>;
+  RequirementFulfillmentInformation<{ readonly courses: readonly (readonly number[])[] }>;
```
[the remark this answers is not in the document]
SamChou19815: yes

src/requirement-types.d.ts
SamChou19815: yes, so the roster doesn't do anything to the requirement graph anymore, so it's better to remove it.

src/components/Requirements/SubRequirement.vue
SamChou19815: yes, you can see in the clone function I added, I replace the id with a real allocated id when the course is dropped into a semester.

PR #324, opened 2021-02-24
Integrate user's requirement choice into requirement graph

SamChou19815:
### Summary 

This PR integrates user's requirement choice into the graph by implementing the `userChoiceOnDoubleCountingElimiation` function (I will fix the typo of `elimination`  in another PR). This is mostly straightforward since I just need to turn the object into a list of tuples.

I also did a related factoring of including `uniqueID` into the `CourseTaken` type, so that it can be uniquely traced back to the original course. This is not directly required to make this integration happen, but it's important for future work. Right now our requirement graph just thinks that two courses with the same course id as one single course, but this is inaccurate. e.g. You can take two `HIST 1200`, each accounting for one FWS slot. Therefore, it's important for the `CourseTaken` to keep track of both `courseID` and `uniqueID`, one for requirement matching and the other for differentiation. Although I didn't change the requirement algorithm in this PR, it will be the focus of another PR.

### Test Plan 

Delete some of your courses that show double counted, readd it by selecting some requirement, and see the warning goes away.

### Notes

Note that the courses dragged from the requirement bar can still produce the warnings. That's because they don't go through the process of choosing requirement. We don't really need to choose a requirement for it in UI, because the chosen requirement is already implied by the source of drag and drop. However, that's a thing we haven't done it, and it needs to be done sometime before launch.

### DIFF
--- a/src/components/Course/CourseCaution.vue
+++ b/src/components/Course/CourseCaution.vue
@@ -44,7 +44,7 @@ const getCourseCautions = (course: FirestoreSemesterCourse): CourseCautions => {
   if (illegallyDoubleCountedCourseIDs.has(course.crseId)) {
     illegallyDoubleCountedCourseIDs.forEach(courseId => {
       doubleCountingRequirementWarning = requirementFulfillmentGraph
-        .getConnectedRequirementsFromCourse({ courseId, subject: '', number: '', credits: 0 })
+        .getConnectedRequirementsFromCourse({ courseId, uniqueId: -1, code: '', credits: 0 })
         .filter(id => !userRequirementsMap[id].allowCourseDoubleCounting)
         .map(id => userRequirementsMap[id].name);
     });
--- a/src/components/Requirements/CompletedSubReqCourse.vue
+++ b/src/components/Requirements/CompletedSubReqCourse.vue
@@ -19,8 +19,7 @@
       <div class="completed-reqCourses-course-object-wrapper">
         <req-course
           :color="courseInfoAndSemesterLabel.color"
-          :subject="courseTaken.subject"
-          :number="courseTaken.number"
+          :courseCode="courseTaken.code"
           :compact="true"
           :isCompletedReqCourse="true"
           class="completed-reqCourses-course-object"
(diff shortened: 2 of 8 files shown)


src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -73,22 +68,16 @@ export default Vue.extend({
...
-        if (filteredSemesterCourses.length > 0) {
-          return {
-            semesterLabel: `${semester.type} ${semester.year}`,
-            uniqueID: filteredSemesterCourses[0].uniqueID,
-            color: filteredSemesterCourses[0].color,
-          };
-        }
-      }
-      return { semesterLabel: `${getCurrentSeason()} ${getCurrentYear()}`, uniqueID: 0, color: '' };
+      return { semesterLabel, uniqueID: this.courseTaken.uniqueId, color: courseColor };
```
tcho6319, line 80: Is there a need to return `uniqueID` here? I don't see `courseInfoAndSemesterLabel.uniqueID` being accessed.

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -73,22 +68,16 @@ export default Vue.extend({
         return { semesterLabel: 'Transfer Credits', color: transferCreditColor, uniqueID: 0 };
```
tcho6319, line 68: Should `uniqueID` be -1 since `this.courseTaken.uniqueId === -1` for transfer credits? What's the distinction between `uniqueId` and `uniqueID`?

src/components/Course/CourseCaution.vue
```diff
@@ -44,7 +44,7 @@ const getCourseCautions = (course: FirestoreSemesterCourse): CourseCautions => {
   if (illegallyDoubleCountedCourseIDs.has(course.crseId)) {
     illegallyDoubleCountedCourseIDs.forEach(courseId => {
       doubleCountingRequirementWarning = requirementFulfillmentGraph
-        .getConnectedRequirementsFromCourse({ courseId, subject: '', number: '', credits: 0 })
+        .getConnectedRequirementsFromCourse({ courseId, uniqueId: -1, code: '', credits: 0 })
```
tcho6319, line 47: Is the -1  for `uniqueId` intended to be a dummy value here?

tcho6319 (commented):
Overall LGTM! I just left some comments but once they're addressed, this PR is g2g!

src/components/Course/CourseCaution.vue
SamChou19815: Yes

tcho6319 (approved):
LGTM!

PR #327, opened 2021-02-25
Make requirement graph aware of courseId and uniqueId

SamChou19815:
### Summary 

Depends on #326.

In the past, our requirement algorithm is essentially a bipartite graph between requirement ID and course ID. This works pretty well in most cases, but not all. It's important to note that we were assuming course ID uniquely denotes a course taken by the user. We know this is not true, since cross-listed courses have the same ID, but it doesn't cause any issues most of the time.

However, it does cause issues sometimes. For example, the FWS requirement can be fulfilled by HIST 1200, and HIST 1200 has multiple sections, each denoting a different actual class. You can take `HIST 1200 - SEC 201` in the first semester and `HIST 1200 - SEC 202` in the second semester, and then the requirement should just be fulfilled without issue. It is not the case right now, because our requirement graph thinks that they are the same course, and of course, one course can only be used for one slot.

This PR changes this situation by making the requirement graph building algorithm aware of the distinction between `courseId` and `uniqueId`. The requirement graph should be a graph between requirements and **unique ID**, because it's legit to make one class fulfill requirement A and requirement B. `courseId` is still needed, because we need `courseId` to match against eligible courses in requirements json.

After this PR, each course, even repeated ones, can choose its own requirement. Therefore, I also changed the double counted course courseId set to be unique ID set.

The changes in production code itself is mostly straightforward. A lot of changes are coming from the tests.

### Test Plan 

Normal requirement computation is not changed:
local:
local
master:
master

Now duplicate courses can be used to count towards different requirements:
(Example: Two different FWS both under HIST 1200)
Screen Shot 2021-02-25 at 16 11 33

### DIFF
--- a/src/requirements/requirement-graph-builder.ts
+++ b/src/requirements/requirement-graph-builder.ts
@@ -1,7 +1,13 @@
-import RequirementFulfillmentGraph from './requirement-graph';
-import { HashMap } from './util/collections';
+import RequirementFulfillmentGraph, { CourseWithUniqueId } from './requirement-graph';
 
-type BuildRequirementFulfillmentGraphParameters<Requirement extends string, Course> = {
+interface CourseForRequirementGraph extends CourseWithUniqueId {
+  readonly courseId: number;
+}
+
+type BuildRequirementFulfillmentGraphParameters<
+  Requirement extends string,
+  Course extends CourseForRequirementGraph
+> = {
   /**
    * A list of applicable requirements in the system. e.g. if the user is CS major
    * in COE, then the list should contain all university requirements, COE requirements, and CS major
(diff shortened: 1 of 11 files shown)


src/requirements/requirement-graph-builder.ts
```diff
@@ -12,86 +18,97 @@ type BuildRequirementFulfillmentGraphParameters<Requirement extends string, Cour
   readonly userCourses: readonly Course[];
   /**
    * Some requirements might have several different ways to fulfill them. This is a map from
-   * toggleable requirements to a list of courses that could be applied to the requirement under
+   * toggleable requirements to a list of course IDs that could be applied to the requirement under
```
tcho6319, line 21: This is the crseId (that crosslisted courses share) and not unique ID right?

src/requirements/requirement-graph-builder.ts
```diff
@@ -12,86 +18,97 @@ type BuildRequirementFulfillmentGraphParameters<Requirement extends string, Cour
...
+  readonly userChoiceOnDoubleCountingElimination: Readonly<Record<number, Requirement>>;
   /**
    * Naively give a list of courses that can satisfy a requirement. Most of the time this function
    * should just return the pre-computed course list. For requirements have multiple fulfillment
    * strategies, it will return the union of all pre-computed course list.
    */
   readonly getAllCoursesThatCanPotentiallySatisfyRequirement: (
     requirement: Requirement
-  ) => readonly Course[];
+  ) => readonly number[];
```
tcho6319, line 40: can we update the spec so that it says we're getting a list of the course id (crseId from cornell course roster api that crosslisted courses share)

tcho6319 (changes requested):
LGTM Sam! Requested a small change for documentation. Now that we use `courseId`, `uniqueId`, and `course`, I think we should make sure both the code and the documentation distinguish among these three clearly.

src/requirements/requirement-graph-builder.ts
SamChou19815: Yes, because it's used to match to eligible courses.

tcho6319 (approved):
Ty Sam!

PR #328, opened 2021-02-26
Bootstrap requirement with warnings

SamChou19815:
### Summary 

Setup an example requirement with warnings data. Currently the checker is very loose. We could do better but I decide to do it in a future PR, so that this PR is easily reviewbale and @einc can start adding more data without waiting on my work.

### Test Plan 

No impact on other parts of the app (besides the technical elective one) since the new field is just ignored.


src/requirement-types.d.ts
```diff
@@ -7,6 +7,11 @@ type RequirementCommon = {
   readonly source: string;
   /** If this is set to true, then an edge to the course doesn't count towards double counting. */
   readonly allowCourseDoubleCounting?: true;
```
einc, line 9: unrelated to your PR, but I just noticed this. why is the type of allowCourseDoubleCounting true?

einc (approved):
Looks good, thank you Sam!

tcho6319 (approved):
LGTM

src/requirement-types.d.ts
SamChou19815: It's an optional field. (See the ?), so when it doesn't exist it just means false

einc: ah got it, ty!!

PR #329, opened 2021-02-27
[easy] Treat requirement with warnings as self-check

SamChou19815:
### Summary 

We can easily treat requirement with warnings as self-check by checking whether the `checkerWarning` field exist on requirement. If it exists, we treat it as self-check.

### Test Plan 

In the add modal, `technical elective` is still considered as self-check.
Screen Shot 2021-02-27 at 16 36 36

In the sidebar, `technical elective` is still considered as self-check with dropdown, but has correct number of courses shown.
Screen Shot 2021-02-27 at 16 36 42

### Notes 

The warning text is not shown yet because I don't see the design.

### DIFF
--- a/src/components/Requirements/SubRequirement.vue
+++ b/src/components/Requirements/SubRequirement.vue
@@ -80,7 +80,11 @@
         </div>
       </div>
       <div
-        v-if="displayDescription && subReq.requirement.fulfilledBy !== 'self-check'"
+        v-if="
+          displayDescription &&
+          subReq.requirement.fulfilledBy !== 'self-check' &&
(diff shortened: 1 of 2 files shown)


einc (approved):
Looks good, ty Sam! Here's the link for the [warning design](https://www.figma.com/file/81R3CvQY1EgxFZSxtEJNrj/Ein-Working-File?node-id=540%3A770) on Req Bar
Also added the svg for the emoji [here](https://drive.google.com/drive/folders/1EbYrNnk0Bb_ynel0x_DT6AEYoNCH9wtM?usp=sharing)

PR #330, opened 2021-02-27
See all fixes

willespencer:
### Summary 

This pull request fixes a couple bugs with see all functionality on the requirements menu, reverts some previously introduced requirements menu bugs, and adds pagination to the see all menu.

- [x] Limit see all courses to only show courses from that specific self check slot, and not courses from other slots for the same requirement
- [x] Add functionality to shift between 24 courses at a time in the see all menu with new previous and next buttons
- [x] Fix mobile reqs menu styling
- [x]  Make courses in see all menu show up as smaller course versions at smaller breakpoints
- [x] Other styling fixes (reqs continues to the bottom of the screen, reqs dropdowns show above bar, course cards have draggable cursor, course cards triple dots sometimes displayed incorrectly)

Notion item: https://www.notion.so/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=127d9034dcab4f2fb369f0a8bc874d52 (first two already fixed by #323, last two are duplicate items)

Example of the first item: The 4th BME concentration requirement see all now does not show Chem 1570 or other courses it should not from the first 3 slots.
image
image

Remaining TODOs:

- [ ] Mobile see all menu is not ideal. Courses are still draggable (with the new grabbing cursor), and on high breakpoints, look off as the entire menu is left oriented instead of center-oriented. Don't think this is worth spending time on though until/if we decide how we want courses to work on this mobile menu.

### Test Plan 

1. Confirm clicking "see all" on a requirement with multiple req slots only displays courses that fulfill that specific slot (in contrast to the dev site that shows courses from all slots). An example would be for the Biological Sciences major in A&S, the "College Mathematics" requirement should not show MATH 1106 or MATH 1110.
2. Confirm that clicking on see all for a requirement with more than 24 courses lets you switch between previous and next pages. An example would be for the "DEA Thematic Courses" req in DEA in Human Ecology as it has 32 courses that can fulfill it, and thus 2 pages. Previous and Next buttons should be disabled when prev/next pages do not exist
3. Confirm requirements with much more requirements also let you cycle through a ton of pages (such as A&S or CALS credits)
4. Mobile requirements menu works as before, including see all, and adding a new course from self check modal
5. The requirements menu continues to the bottom of the screen when the semester view is both smaller and larger than the screen height (add and delete semesters/courses to make this happen)
6. Ensure that on smaller breakpoints, the see all courses are smaller in width than the ones in the semester plan
7.  Identified a new bug on master where courses the triple dot coloring shrunk in width when the info box was filled (Fall, Winter, Spring, Summer) on smaller breakpoints. Confirm this does not happen anymore.
image

### Notes 

- [x] (OUTDATED) Do we want this 24 course at a time functionality in see all? It seems better than the current functionality where only the first 24 courses are shown ever, but it is not that intuitive, especially with the wording "see all." Showing all of them at once however would be problematic for very broad requirements such as "CALS Credits" that have up to 2500 possible courses.
- [x] Should these new previous/next buttons show up for subreqs with only 1 page? See the above screenshot.
- [x] When a course is dragged from the see all menu to the plan, it shows back up in the see all menu after dropped, and does not disappear until the menu is reopened. Is this functionality ok?

### DIFF
--- a/src/components/Requirements/IncompleteSelfCheck.vue
+++ b/src/components/Requirements/IncompleteSelfCheck.vue
@@ -146,6 +146,8 @@ export default Vue.extend({
   }
   &-content {
     z-index: 2;
+    position: absolute;
+    width: 88%;
     background: $white;
     box-shadow: -4px 4px 10px rgba(0, 0, 0, 0.25);
     border-radius: 7px;
--- a/src/components/Requirements/IncompleteSelfCheck.vue
+++ b/src/components/Requirements/IncompleteSelfCheck.vue
@@ -116,6 +116,10 @@ export default Vue.extend({
     &:not(:first-child) {
       margin-top: 0.5rem;
     }
+
+    &-wrapper {
+      position: relative;
+    }
   }
   &-placeholder {
     height: 100%;
@@ -147,7 +151,7 @@ export default Vue.extend({
   &-content {
     z-index: 2;
     position: absolute;
-    width: 88%;
+    width: 100%;
     background: $white;
     box-shadow: -4px 4px 10px rgba(0, 0, 0, 0.25);
     border-radius: 7px;
--- a/src/components/Requirements/Requirements.vue
+++ b/src/components/Requirements/Requirements.vue
@@ -81,7 +81,7 @@ Vue.use(VueCollapse);
 
 export type ShowAllCourses = {
   readonly name: string;
-  readonly courses: FirestoreSemesterCourse[];
+  readonly courses: CourseTaken[];
 };
 
 type Data = {
@@ -169,12 +169,10 @@ export default Vue.extend({
     },
     onShowAllCourses(showAllCourses: {
       requirementName: string;
-      subReqCoursesArray: SubReqCourseSlot[];
+      subReqCoursesArray: CourseTaken[];
     }) {
       this.shouldShowAllCourses = true;
-      const allPotentialCourses = showAllCourses.subReqCoursesArray
-        .flatMap(slot => (slot.isCompleted ? [] : slot.courses))
-        .slice(0, 24);
+      const allPotentialCourses = showAllCourses.subReqCoursesArray.slice(0, 24); // limited to 24 courses
       const lastCourse = allPotentialCourses[allPotentialCourses.length - 1];
       this.lastLoadedShowAllCourseId = lastCourse.crseId;
       this.showAllCourses = {
--- a/src/components/Requirements/Requirements.vue
+++ b/src/components/Requirements/Requirements.vue
@@ -80,7 +80,7 @@ Vue.use(VueCollapse);
 
 export type ShowAllCourses = {
   readonly name: string;
-  readonly courses: CourseTaken[];
+  readonly courses: FirestoreSemesterCourse[];
 };
 
 type Data = {
@@ -168,7 +168,7 @@ export default Vue.extend({
     },
     onShowAllCourses(showAllCourses: {
       requirementName: string;
-      subReqCoursesArray: CourseTaken[];
+      subReqCoursesArray: FirestoreSemesterCourse[];
     }) {
       this.shouldShowAllCourses = true;
 
--- a/src/components/Requirements/Requirements.vue
+++ b/src/components/Requirements/Requirements.vue
@@ -90,7 +90,7 @@ Vue.use(VueCollapse);
 
 export type ShowAllCourses = {
   readonly name: string;
-  readonly shownCourses: FirestoreSemesterCourse[];
+  shownCourses: FirestoreSemesterCourse[];
   readonly allCourses: FirestoreSemesterCourse[];
 };
 
@@ -154,7 +154,7 @@ export default Vue.extend({
       return Math.ceil(this.showAllCourses.allCourses.length / this.maxSeeAllCoursesPerPage);
     },
     hasNextPage(): boolean {
-      return this.showAllPage * this.maxSeeAllCoursesPerPage < this.numPages;
+      return this.showAllPage + 1 < this.numPages;
     },
     hasPrevPage(): boolean {
       return this.showAllPage > 0;
--- a/src/components/Requirements/Requirements.vue
+++ b/src/components/Requirements/Requirements.vue
@@ -69,7 +69,6 @@ import introJs from 'intro.js';
 
 import Course from '@/components/Course/Course.vue';
 import RequirementView from '@/components/Requirements/RequirementView.vue';
-import { SubReqCourseSlot } from '@/components/Requirements/SubRequirement.vue';
 import DropDownArrow from '@/components/DropDownArrow.vue';
 // emoji for clipboard
 import clipboard from '@/assets/images/clipboard.svg';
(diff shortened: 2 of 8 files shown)


src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -146,6 +146,8 @@ export default Vue.extend({
   }
   &-content {
     z-index: 2;
+    position: absolute;
+    width: 88%;
```
SamChou19815, line 150: 🤔

What's the 12% percent? Maybe we should use flexbox?

src/components/Requirements/SubRequirement.vue
```diff
@@ -429,6 +429,8 @@ button.view {
     }
     &-content {
       z-index: 2;
+      position: absolute;
+      width: 80%;
```
SamChou19815, line 433: 🤔

src/components/Requirements/IncompleteSelfCheck.vue
willespencer: This magic number already existed in the codebase (I actually removed this line in #312 when I was struggling with the requirements styling), so this PR just added it back when I realized I could fix the issues in a better way that also worked for mobile. I can still look into fixing it with flex or something if you think that is a good idea.

SamChou19815 (approved):

src/components/Requirements/Requirements.vue
```diff
@@ -169,14 +168,23 @@ export default Vue.extend({
...
+        this.lastLoadedShowAllCourseId + 24
+      );
+
+      // next time see all is open, show next 24 courses (unless already reached the end)
+      if (this.lastLoadedShowAllCourseId + 24 > showAllCourses.subReqCoursesArray.length) {
+        this.lastLoadedShowAllCourseId = 0;
+      } else {
+        this.lastLoadedShowAllCourseId += 24;
+      }
+
```
tcho6319, line 187: I'm not really a fan of showing 24 different courses every time the user clicks See All. I think it could be a little confusing. Could we possibly re-introduce pagination? I think a long term goal for courses with lots of See All courses (e.g. CALS) would be to have an add course button that opens up the add modal where the only options are the See All courses. So for example, if I click an add course button for CALS credits, when I start typing a course to add in the add modal, the dropdown suggestions are the See All courses. @einc would you agree?

tcho6319 (changes requested):
Thanks for the fixes Will! I requested a change for the See All course card sizes and left a comment re: the 24 courses. I'll bump this PR to @einc as well. 

1. The See All course card sizes look a little too large at smaller width breakpoints, can we adjust them to be the smaller card size that courses are on mobile? @einc does that sound ok?
Screen Shot 2021-02-28 at 1 37 52 PM

benjamin-shen:
Hey Will, I noticed 2 more potential fixes related to this PR:
1. Could you change the cursor to pointer when hovering over the course cards? Right now I can't tell that they're draggable.
2. When I drag a course card to the semester view, I can't "cancel" the drag by dragging it back to the sidebar. Is this an easy fix? If not, maybe it can go in another PR.

Thanks :D

src/components/Requirements/Requirements.vue
einc: +1 on showing 24 different courses. is pagination possible/easy to implement? this is a super rough design and ill look into it after todays meeting but would something like this be do-able by launch?
Screen Shot 2021-02-28 at 5 29 15 PM

tcho6319: idk if it would be possible to do it like before but could we do something like once you scroll to a certain point, more courses get loaded. And then this batching stops once you get to the end?

SamChou19815: @tcho6319 I think this approach still needs to render increasing number of cards. It's however possible to do virtual list rendering: calculating how many cards are visible and only render those visible and a few cards around them. However, this is much more involved, and the pagination solution is much simpler.

src/components/Requirements/Requirements.vue
```diff
@@ -169,22 +194,47 @@ export default Vue.extend({
...
+        this.showAllPage * this.maxSeeAllCoursesPerPage,
+        (this.showAllPage + 1) * this.maxSeeAllCoursesPerPage
+      );
+      return allPotentialCourses;
+    },
     backFromSeeAll() {
       this.shouldShowAllCourses = false;
-      this.showAllCourses = { name: '', courses: [] };
+      this.showAllCourses = { name: '', shownCourses: [], allCourses: [] };
     },
```
tcho6319, line 238: In `backFromSeeAll()` can we also reset `this.showAllPage = 0`? Currently, if I flip to the 2nd page for the See All panel of a requirement, if I click on the See All for another requirement, I automatically get directed to the 2nd page.

https://user-images.githubusercontent.com/54298311/109899264-98143600-7c63-11eb-9db7-2d394b0feabb.mov

This could be bad in the case where the page (i.e. 2nd page) doesn't exist in the See All panel of a req:
Screen Shot 2021-03-03 at 9 01 52 PM

src/components/Requirements/Requirements.vue
```diff
@@ -37,14 +36,25 @@
       </div>
       <div class="see-all-padding-x py-3">
         <h1 class="title">{{ showAllCourses.name }}</h1>
+        <div class="see-all-pages">
+          <span class="see-all-pageCount">{{ pageText }}</span>
+          <div class="see-all-buttonWrapper">
+            <button class="see-all-button" :class="{ 'see-all-button--disabled': !hasPrevPage }">
+              <span class="see-all-button-text" @click="prevPage()">PREV</span>
+            </button>
```
tcho6319, line 44: Thank you for doing buttons instead of divs haha. Can you move `@click` to the `button` tag? Atm, I'm able to tab but I can't press `Enter` key to click the PREV or NEXT buttons.

tcho6319 (changes requested):
Tysm Will! This looks great. Thank you for having a thorough summary as well! I have some additional suggestions, but once those are addressed, I'd be good with approving this PR.

> Mobile see all menu is not ideal. Courses are still draggable (with the new grabbing cursor), and on high breakpoints, look off as the entire menu is left oriented instead of center-oriented. Don't think this is worth spending time on though until/if we decide how we want courses to work on this mobile menu.

Yes, this is more the work after Yuxuan's launch mobile designs are finalized. Good to note, but not in the scope of this PR.
 
> Should these new previous/next buttons show up for subreqs with only 1 page? See the above screenshot.

I personally feel like the previous/next buttons (and even the "Page 1/1") are unnecessary for subreqs with only 1 page. @einc what do you think? [Yes I think so too -ein]

> When a course is dragged from the see all menu to the plan, it shows back up in the see all menu after dropped, and does not disappear until the menu is reopened. Is this functionality ok?

It could be a fix we look into post-launch but I think it should be fine for launch.

einc:
running into this bug where if i have two requirements uncollapsed I think the page counter carries over to the next requirement. i was on page 20 of see all and then when i went to a different requirement's see all it was on page 20, even though there are only 6 pages
https://user-images.githubusercontent.com/14082828/109906377-62c11580-7c6e-11eb-810c-e58100c2871f.mov

src/components/Requirements/Requirements.vue
willespencer: Shoot did not even realize I put them on the spans, thank you 😄

tcho6319 (approved):
LGTM :D

PR #333, opened 2021-02-28
Front end onboarding issues

jason-tung:
### Summary 

Fix inconsistent padding outside of each onboarding page and added colors to the asterisks under onboarding

https://www.notion.so/Front-End-Onboarding-Issues-9e0328bb8fc348b99a374eb9622a6a24

### Test Plan 
Only changes should be color of asterisks is changed to emGreen for onboarding pages and all onboarding pages should have consistent padding between the border of the inside boxes and the edge of the modal (the red length)
image

### DIFF
--- a/src/components/Modals/Onboarding/Onboarding.scss
+++ b/src/components/Modals/Onboarding/Onboarding.scss
@@ -463,6 +463,10 @@ input {
   &--hidden {
     display: none;
   }
+
+  &-required-star {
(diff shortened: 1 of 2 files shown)


SamChou19815 (commented):
Test plan should give a screenshot or say what need to be tested. 👀 are only reserved for very simple changes

SamChou19815 (approved):

PR #334, opened 2021-03-01
No duped majors, No duped tests

jason-tung:
### Summary 

Users cannot add duplicate majors and minors in the frontend anymore
Deduped exam names in onboarding

https://www.notion.so/don-t-let-users-add-duplicate-majors-and-minors-d5ef59eb63634e6c886fbe281c43407c
https://www.notion.so/De-dupe-exam-names-in-Onboarding-d657f675b5be4e7a85ef357ea088144c

### Test Plan 

Try to add duplicate majors and notice you cannot do so.
image

Notice certain exams like stats are not showing up twice anymore.
image

### Notes
The duplicate majors fix does not apply to AP tests, which use a different kind of dropdown? Not sure if that needs to be addressed as well.

### DIFF
--- a/src/components/Modals/Onboarding/OnboardingBasicMultiDropdown.vue
+++ b/src/components/Modals/Onboarding/OnboardingBasicMultiDropdown.vue
@@ -45,11 +45,14 @@ export default Vue.extend({
     },
     getSelectableOptions(choice: string) {
       const selectableOptions: Record<string, string> = {};
+      // copy availableChoices but don't include ones that are already selected
       for (const key of Object.keys(this.availableChoices)) {
+        // don't include selected ones
         if (!this.dropdownChoices.includes(key)) {
           selectableOptions[key] = this.availableChoices[key];
         }
       }
+      // add the current selection associated with this input into the availableChoices
       if (choice !== '') {
         selectableOptions[choice] = this.availableChoices[choice];
       }
--- a/src/components/Modals/Onboarding/OnboardingBasicMultiDropdown.vue
+++ b/src/components/Modals/Onboarding/OnboardingBasicMultiDropdown.vue
@@ -50,7 +50,9 @@ export default Vue.extend({
           selectableOptions[key] = this.availableChoices[key];
         }
       }
-      selectableOptions[choice] = this.availableChoices[choice];
+      if (choice !== '') {
+        selectableOptions[choice] = this.availableChoices[choice];
+      }
       return selectableOptions;
     },
   },
--- a/src/components/Modals/Onboarding/OnboardingBasicMultiDropdown.vue
+++ b/src/components/Modals/Onboarding/OnboardingBasicMultiDropdown.vue
@@ -3,7 +3,7 @@
     <onboarding-basic-single-dropdown
       v-for="(choice, index) in dropdownChoices"
       :key="choice"
-      :availableChoices="availableChoices"
+      :availableChoices="getRemainingChoicesAfterChoice(choice)"
       :choice="choice"
       :cannotBeRemoved="dropdownChoices.length === 1 && choice === ''"
       @on-select="choice => onSelect(choice, index)"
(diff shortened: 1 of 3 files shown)


src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -192,6 +192,13 @@ type Data = {
 
 const scoresAP = [1, 2, 3, 4, 5];
 const scoresIB = [1, 2, 3, 4, 5, 6, 7];
+const existingAP: Record<string, boolean> = {};
+// filter duplicate exam names and ones already selected
+reqsData.AP = reqsData.AP.filter(ap => {
+  const inExisting = ap.name in existingAP;
+  existingAP[ap.name] = true;
+  return !inExisting;
+});
```
SamChou19815, line 201: Why don't you do the same for IB?

src/components/Modals/Onboarding/OnboardingBasicMultiDropdown.vue
```diff
@@ -43,6 +43,18 @@ export default Vue.extend({
...
+      for (const key of Object.keys(this.availableChoices)) {
+        if (!this.dropdownChoices.includes(key)) {
+          selectableOptions[key] = this.availableChoices[key];
+        }
+      }
+      if (choice !== '') {
+        selectableOptions[choice] = this.availableChoices[choice];
+      }
+      return selectableOptions;
+    },
```
SamChou19815, line 57: can you explain in comments what are you doing here

src/components/Modals/Onboarding/OnboardingTransfer.vue
jason-tung: There are no duplicate exams for IB but I can do the check anyways

SamChou19815 (approved):
LGTM

src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -192,6 +192,20 @@ type Data = {
 
 const scoresAP = [1, 2, 3, 4, 5];
 const scoresIB = [1, 2, 3, 4, 5, 6, 7];
+const existingAP: Record<string, boolean> = {};
```
tcho6319, line 195: Do you think making `existingAP` and `existingIB` arrays could simplify the filtering? So you would check whether the name of the exam is in the array. Is there a benefit to using `Record` here?

tcho6319 (changes requested):
Thanks Jason! LGTM. 

> The duplicate majors fix does not apply to AP tests, which use a different kind of dropdown? Not sure if that needs to be addressed as well.

Since we're assuming that users can only submit one score for each AP/IB exam they take, can we also dedupe AP/IB exams based on what the user has already inputted into onboarding? So for example, I couldn't input AP Bio multiple times.
Screen Shot 2021-03-01 at 11 21 27 AM

Once this change and my comment are addressed, this PR should be g2g.

src/components/Modals/Onboarding/OnboardingTransfer.vue
jason-tung: time complexity

src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -329,6 +343,22 @@ export default Vue.extend({
           });
         });
     },
+    getSelectableOptions(exams: FirestoreAPIBExam[], choice: string) {
+      const selectedExams = exams.map(exam => exam.subject);
+      const selectableOptions: string[] = [];
+      // copy subjectsAP but don't include ones that are already selected
+      for (const subject of subjectsAP) {
```
tcho6319, line 350: Since we only iterate over `subjectsAP` and not `subjectsIB`, the options for the IB exams are the options for AP exams. Left is this PR's staging site and right is master.
Screen Shot 2021-03-02 at 5 56 22 PM

tcho6319 (changes requested):
Almost there! I found a small bug where the dropdown for IB exams show the AP exams.

src/components/Modals/Onboarding/OnboardingTransfer.vue
jason-tung: added a parameter for this -- thanks for pointing this out.

tcho6319 (approved):
LGTM. Ty Jason!

PR #337, opened 2021-03-02
A&S Updates + Preparing to add updated A&S Requirements 

einc:
### Summary 

- Updated requirements for College of Arts and Science (mainly so that the distribution requirements for A&S can be now be double-counted).

- @hahnbeelee hiii our resident a&s major. If you could take a close look at the A&S requirements and make sure they are correct that would be great! 

- Also, made a few changes to prepare to change the current A&S requirements to A&S "before Fall 2020", in order to make an A&S "Fall 2020 and after" requirement as our quick fix for the [changing requirements in A&S](https://as.cornell.edu/education/degree-requirements). Made changes with the naming and its impact on Exam Credit. @SamChou19815 Is there anything I missed? I'll add the new A&S requirements in the next PR!

### Test Plan 

Add courses as a major in A&S. Check that distribution requirements are double-counted and nothing else is broken
Example: CS 1110 counts for MQR-AS (A&S req) AND Intro Programming (CS major Req)
Screen Shot 2021-03-02 at 12 56 35 PM

### Notes 
Noticed this bug for the Distribution Requirement where you have to take 5 courses from at least 4 categories. This means that 2 courses could come from the same category. 
Screen Shot 2021-03-02 at 12 52 23 PM
However, when I add two courses from the same category (ex: ANTHR 1400 and ASIAN 2208), it now looks like there are 6 (instead of 5) courses that need to be fulfilled. 
Screen Shot 2021-03-02 at 12 54 26 PM
By adding courses from different categories, you can eventually fulfill this requirement. 
This is a post-launch fix but hopefully eventually we can show the category name ("Add CA-AS") instead of "Add course 1"


SamChou19815 (approved):
Changes look good to me. Haven't checked the accuracy of requirement yet, but I believe @hahnbeelee might be better at this.

hahnbeelee (changes requested):
I think MSQ and MSQ/PSB are a OR relation. I can't tell if this is implemented? On the frontend it seems like that isn't shown? But that could just be my job lol. As long as there's a way for me to distinguish the fact that it's a OR relation it should be good.

einc:
> I think MSQ and MSQ/PSB are a OR relation. I can't tell if this is implemented? On the frontend it seems like that isn't shown? But that could just be my job lol. As long as there's a way for me to distinguish the fact that it's a OR relation it should be good.

Screen Shot 2021-03-03 at 4 49 15 PM
i think you have to take a MQR course _and_ another MRQ/PSB course. So 2 courses total. Does that seem right? But there's currently a problem where one MQR course counts for both requirements rn. I'm temporarily going to make the MQR/PBS not double countable for launch. We'll have to fix this eventually because ppl will prob want to double count the MQR/PBS requirement with major requirements.

hahnbeelee:
> > I think MSQ and MSQ/PSB are a OR relation. I can't tell if this is implemented? On the frontend it seems like that isn't shown? But that could just be my job lol. As long as there's a way for me to distinguish the fact that it's a OR relation it should be good.
> 
> Screen Shot 2021-03-03 at 4 49 15 PM
> 
> i think you have to take a MQR course and another MRQ/PSB course. So 2 courses total. Does that seem right? But there's currently a problem where one MQR course counts for both requirements rn. I'm temporarily going to make the MQR/PBS not double countable for launch. We'll have to fix this eventually because ppl will prob want to double count the MQR/PBS requirement with major requirements.
> oh shoot i misread your question. I'm not sure if we can tell MSQ and MSQ/PSB are OR relations. @SamChou19815 Do you know?

Yes that is 100% correct! So, one class can't count for both.

hahnbeelee:
image
Just for clarification - was your most recent commit supposed to make any changes here? Am I supposed to not be seeing one of "Math" or "PBS or MQR"?

Sorry if I'm asking a lot of questions. I'm just genuinely confused/curious. Because I know that I can do some logical rendering on the frontend, but I want to know what logic is available for me from the "backend"

hahnbeelee (approved):
I shall just approve for noww. My questions aren't really in regards to the PR specifically. just general questions for my understanding!

src/requirements/data/colleges/asPreFA2020.ts
```diff
@@ -0,0 +1,201 @@
+import { Course, CollegeOrMajorRequirement } from '../../types';
+import { courseIsFWS, includesWithSingleRequirement } from '../checkers-common';
+
+const casPreFA2020Requirements: readonly CollegeOrMajorRequirement[] = [
+  {
+    name: 'A&S Credits',
+    description:
+      '100 credits in Arts & Sciences are required.' +
```
tcho6319, line 8: Nit: I would add a space at the end so that it's  '100 credits in Arts & Sciences are required. '

SamChou19815 (approved):
@einc @hahnbeelee The current changes look good to me.

Right now the allow double counting mechanism is still not good enough. I initially designed this for the engineering model where a requirement either allow or disallow double count. But the A&S one seems to be like a course can fulfill some major requirement and A&S requirement, but it can't fulfill two A&S requirement at the same time.

Right now the hack seems to work, but post-launch we should think carefully about whether we can find more principled approach.

tcho6319 (commented):
Just a small nit, but once that's addressed, I think this PR is g2g.

> i think you have to take a MQR course and another MRQ/PSB course. So 2 courses total. Does that seem right? But there's currently a problem where one MQR course counts for both requirements rn. I'm temporarily going to make the MQR/PBS not double countable for launch. We'll have to fix this eventually because ppl will prob want to double count the MQR/PBS requirement with major requirements.

Yes, I think this is a good temporary fix. It's better to give a double counting warning than to have the user mistakenly think that 1 course fulfills both MQR-AS and MQR/PSB reqs.

PR #338, opened 2021-03-02
Change mechanism of disallowing dropping to fix drop cancel issue

SamChou19815:
## Summary

We don't need to rely on the move function's return value. Instead, we can use the [`group` option](https://github.com/SortableJS/Sortable#group-option) to disallow items being put into the list.

Surprisingly, this change also partially fixes the issue of [not dropping the req course into the semester still causing it to be added](https://www.notion.so/Cancel-Drag-Req-Course-8f5d03f44cb14cdf8576e13e19f3e21a). However, **it's not a full fix**, since you still need to drop the req course into the original req course slot. I don't believe a full fix is possible since it seems to be the fundamental assumption of the library that the card will stuck in the previous eligible slot until it finds a new slot.

## Test Plan

- Try to drop semester class into req courses and see it fail.
- Try to drag a req course into semester but not drop, then drag it back to the req course slot and drop. No change in semester data.

### DIFF
--- a/src/components/Requirements/IncompleteSubReqCourse.vue
+++ b/src/components/Requirements/IncompleteSubReqCourse.vue
@@ -15,9 +15,7 @@
       <draggable
         v-if="courses.length > 0"
         class="draggable-requirements-courses"
-        group="draggable-semester-courses"
-        :move="onDraggedCourseMove"
-        data-not-droppable="true"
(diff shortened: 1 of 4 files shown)


tcho6319 (approved):
LGTM Sam ty! I'll look into this more post-launch but I think that this partial solution is good for our launch.

PR #339, opened 2021-03-02
Fix incorrect double counting warning issue

SamChou19815:
### Summary 

Recently I found that I'm going some weird double counting warnings like:

error

Digging deep into it, it appears that the current code (written by me 😅) is horribly wrong. It will cause us to always fetch the double counted requirement of the last `illegallyDoubleCountedCourseUniqueIDs`. Now I fixed it by fetching the relevant one from the course id.

### Test Plan 

Now fixed: (it shows the warning, because it's added from drop from req courses, which doesn't remember req choice yet)
fixed


einc (approved):
tested and looks good, ty sam for catching this!

PR #340, opened 2021-03-04
[easy] Update course caution icon

SamChou19815:
### Summary 

Mostly just change the color to be consistent with warning color.

### Test Plan 

Screen Shot 2021-03-03 at 23 35 39


src/components/Course/CourseBaseTooltip.vue
```diff
@@ -1,11 +1,7 @@
 <template>
   <div class="course-tooltip">
     <div class="course-iconWrapper course-iconWrapper--caution">
-      <img
-        class="course-icon course-icon--caution"
-        src="@/assets/images/caution.svg"
-        alt="caution sign"
-      />
+      <img class="course-icon course-icon--caution" :src="icon" :alt="alt" />
```
SamChou19815, line 4: if you expand below, you will see that it's using the correct value. I defined the computed properties but forget to use it 😄

einc (approved):
ty for updating this!

PR #342, opened 2021-03-04
Refactor firestore modifiers and gtag events

benjamin-shen:
### Summary 

This pull request cleans up some of the data modifier code after the VueX migration.

- [x] refactor firestore modifiers to be directly called from `global-firestore-data.ts` (opposed to having convoluted and repeated logic in the components)
- [x] refactor firestore-related gtag events from the semester components to `global-firestore-data.ts`
- [x] refactor all gtag events; they are now handled in `gtag.ts`

Remaining TODOs:

- [ ] Test that gtag still works properly

### Test Plan 

Make sure add course, delete course, add semester, delete semester still work properly. Now that the requirements bar listens to firestore changes, the important thing to test is that the `user-semesters` data is being stored properly in firestore.

### Notes 

When using or working with firestore modifiers (or any global functions), we should try to move the function call closer to the source of the trigger. This is to try to clean up the long $emit chains when possible.

### DIFF
--- a/src/global-firestore-data.ts
+++ b/src/global-firestore-data.ts
@@ -73,14 +73,14 @@ export const addSemester = (
   gtag?: GTag,
   courses: readonly FirestoreSemesterCourse[] = []
 ): void => {
-  GTagEvent(gtag, "add-semester");
+  GTagEvent(gtag, 'add-semester');
   editSemesters(oldSemesters =>
     [...oldSemesters, createSemester(type, year, courses)].sort(compareFirestoreSemesters)
   );
 };
 
 export const deleteSemester = (type: FirestoreSemesterType, year: number, gtag?: GTag): void => {
-  GTagEvent(gtag, "add-semester");
+  GTagEvent(gtag, 'add-semester');
   editSemesters(oldSemesters =>
     oldSemesters.filter(semester => semester.type !== type || semester.year !== year)
   );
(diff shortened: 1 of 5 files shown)


src/global-firestore-data.ts
```diff
@@ -44,21 +37,62 @@ export const compareFirestoreSemesters = (a: FirestoreSemester, b: FirestoreSeme
...
+  courses: readonly FirestoreSemesterCourse[] = []
+): void => {
+  GTagEvent(gtag, 'add-semester');
+  editSemesters(oldSemesters =>
+    [...oldSemesters, createSemester(type, year, courses)].sort(compareFirestoreSemesters)
+  );
+};
+
+export const deleteSemester = (type: FirestoreSemesterType, year: number, gtag?: GTag): void => {
+  GTagEvent(gtag, 'add-semester');
```
SamChou19815, line 83: typo?

SamChou19815 (approved):
Changes LGTM, with one potential issue

tcho6319 (changes requested):
Tysm Ben! LGTM. I would fix the typo that Sam pointed out but other than that, this PR should be g2g

src/global-firestore-data.ts
benjamin-shen: Thanks lol good catch

SamChou19815 (approved):

tcho6319 (approved):

Issue #343, opened 2021-03-04
[RFC] Fine-grained sub-requirement progress calculation


SamChou19815:
## Background

Right now, we have a very coarse strategy of computing the progress of a requirement. For example, consider the CS core requirement. We have five slots, with these eligible courses:

```json
[
  ["CS 2800", "CS 2802"],
  ["CS 3110"],
  ["CS 3410", "CS 3420"],
  ["CS 4410"],
  ["CS 4820"]
]
```

How do we know that the requirement has been fulfilled? Easy: we have a field for each requirement called `minCount`. In this case, `minCount === 5`. We assign user's courses to each slot, and then count the number of slots that have courses assigned to it. Thus, if the user's courses are `CS 2800, CS 2802, CS 3110`, then the first two slots will be filled, and the progress will be `2/5`.

This seems to work pretty well, until we encountered some other requirements like CS electives. CS electives allows any upper CS classes, and you need 3 courses in total. If you try to describe it in terms of slot, you will get something like

```json
{
  "slots": [["CS 3152", "CS 4110", "CS 4120", "..."]],
  "minCount": 3
}
```

Now we encounter a big problem here: if we use the same algorithm as above, then no matter how many classes have taken by the user in this slots, the fulfillment progress will always be at most `1/3`. Fortunately, we managed to solve this problem a while ago, by introducing a concept called `subRequirementProgress`. When it's `every-course-needed`, then we count the number of slots fulfilled. When it's `any-can-count`, we count the total number of courses, regardless how they are assigned in slots.

As an examplem suppose you have taken `CS 3152, CS 4110`, your progress will be `2/3` if `subRequirementProgress === 'any-can-count'` and `1/3` if `subRequirementProgress === 'every-course-needed'`.

This again works pretty well, until there are more requirements on the slots.

## Issue with fine-grained sub-requirements from Cornell

Let's take a look at an [concentration requirement of InfoSci](https://infosci.cornell.edu/undergraduate/info-sci-majors/ba-information-science-college-arts-sciences/degree-requirements-0-0). You can think of it as a requirement with 3 slots (sub-requirements), and slot A requires 2 courses and slot B and C require one course each.

Let's think how we can express this requirement using the current infra we have. We can first try using `subRequirementProgress === 'every-course-needed'`, then we will have:

```json
{
  "slots": [
    ["course from group A", "course from group A", "course from group A"],
    ["course from group B", "course from group B", "course from group B"],
[...]

PR #344, opened 2021-03-04
Record and compute sub-requirement progress in a fine-grained way

SamChou19815:
### Summary 

Implemented the proposal in #343. Check that issue for the discussion on design decisions.

### Test Plan 

Added a test to ensure `perSlotMinCount.length === checkers.length` for all requirements,

Example showing A&S's PSB vs MQR requirement is correctly handled:
Screen Shot 2021-03-04 at 15 58 56

Engineering requirement is still correctly computed:
consistent

### Notes

Info-sci concentrations are not done in this PR to avoid getting this PR too large.


hahnbeelee:
✅ functionality of PBS/MSQ working as desired
- your residential A&S person

hahnbeelee (changes requested):
wait jk ignore my request for change
still in process of reviewing

hahnbeelee:
OK - so I'm seeing why I was confused. So it seems as though on the add modal it shows that it'll satisfy PBS & MQR but in reality it doesn't end up affecting the sidebar (as desired).
image

We should make it so that it doesn't show that it fulfills MQR or PBS on the add modal. How? Is there a way to check through the frontend that it will eventually not fulfill the requirement? Or is there a way to ensure that it won't show up at all from some calculation in the requirements logic?

hahnbeelee:
So, it seems as though now there is a hierarchy of requirement > sub-requirement > slot

Can you add definitions to all of these in the READ-ME in a future PR?

src/requirements/data/minors/dbme.ts
```diff
@@ -38,23 +38,23 @@ const dbmeMinorRequirements: readonly CollegeOrMajorRequirement[] = [
...
+    perSlotMinCount: [3],
   },
   {
     name: 'Capstone course',
     description: 'AEM 4660',
     source: 'https://business.cornell.edu/programs/undergraduate/minors/business-engineers/',
     checker: includesWithSingleRequirement('AEM 4660'),
     fulfilledBy: 'credits',
-    minCount: 1.5,
+    perSlotMinCount: [1.5],
```
hahnbeelee, line 57: you can have half a credit??

hahnbeelee:
> OK - so I'm seeing why I was confused. If you already took 2 MQR courses & take a third: on the add modal it shows that it'll satisfy PBS & MQR (not as desired) but when you add the course it doesn't end up affecting the sidebar (as desired).
> image
> 
> We should make it so that it doesn't show that it fulfills MQR or PBS on the add modal. How? Is there a way to check through the frontend that it will eventually not fulfill the requirement? Or is there a way to ensure that it won't show up at all from some calculation in the requirements logic? (I am asking these questions to scope if this is a problem I need to address or if it's better for u to address right now in this PR) Clearly, some computation is happening after the course is added instead of when the requirements are shown on the modal.

maybe some more computation needs to happen in the `requirement-frontend-computation.ts` file?

hahnbeelee:
Should I be able to run the tests somehow?

SamChou19815:
> Should I be able to run the tests somehow?

@hahnbeelee  `npm run test`

src/requirements/data/minors/dbme.ts
SamChou19815: Not sure what's going on there, just trying to preserve old behavior

src/requirements/data/majors/bme.ts
```diff
@@ -57,84 +54,81 @@ const biomedicalEngineeringRequirements: readonly CollegeOrMajorRequirement[] =
...
+          'BTRY 4381',
+          'BTRY 4840',
+          'CHEM 4810',
+          'CHEME 5430',
+          'CS 4780'
         ),
         counting: 'courses',
-        subRequirementProgress: 'any-can-count',
-        minCount: 13,
+        perSlotMinCount: [13],
```
tcho6319, line 70: Same for this requirement as well as other reqs for the bme major. Why did we combine the `&` relation slots?

tcho6319 (changes requested):
Thanks Sam! I left some comments about how the `perSlotMinCount` was determined for some of the reqs and requests for documentation in certain functions. The next sprint will involve the team looking through our requirements data and making sure it's as accurate as possible so perhaps more changes can be handled there, but I think some of the `perSlotMinCount` values were off. I also wanted to note an issue below:

1. I don't think the frontend is correctly handling the PBS and MQR courses requirement. It only registers the 3 slots and doesn't add another one to complete the requirement. Not sure if it would be best to address in this PR since it's getting large:

https://user-images.githubusercontent.com/54298311/110164828-eb9b9680-7dbf-11eb-8b17-e981d7862b95.mov

src/requirements/data/minors/dbme.ts
einc: yea, this capstone course is 1.5 credits lol

SamChou19815:
@tcho6319 I updated some incorrect requirement `perSlotMinCount` setup. They should be accurate now. It turns out the original setup is also wrong, but this setup now give them the expressive power to specify the requirement correctly now.

@hahnbeelee
In terms of the PBS/MQR requirement, we do need additional work to make them right. However, I think this PR is a good start. To ensure the add modal doesn't report that you can still use a MQR course to fulfill the requirement when all the slots are already used up, we can use the `computeFulfillmentCoursesAndStatistics` function to compute a new statistics after the new course is added, and only allow if the new stat is greater than the old stat.

@tcho6319
In terms of the 3 slot vs 4 slot issue in the PBS/MQR requirement, I believe we should actually use **3 slots**, since logically there are 3 slots. Currently, it's slightly broken because the frontend code assumes that 1 slot can only be filled by 1 course. This is something we might want to change. Loop in @einc for potential design work.

hahnbeelee (approved):
gotcha, sounds good then!

einc:
For the "3 slot vs 4 slot issue in the PBS/MQR requirement", I thought there were supposed to be 4 slots also?
Two for PBS classes, One for MQR, and One for MQR or PBS?

Is there only 3 slots right now because we adding one MQR class counts for both the MQR AND the (MQR or PBS) slot? Is the problem that we currently can't differentiate the (MQR or PBS) sub-requirement from the PBS and MQR subrequirements?

SamChou19815:
> For the "3 slot vs 4 slot issue in the PBS/MQR requirement", I thought there were supposed to be 4 slots also?
> 
> Two for PBS classes, One for MQR, and One for MQR or PBS?
> 
> 
> 
> Is there only 3 slots right now because we adding one MQR class counts for both the MQR AND the (MQR or PBS) slot? Is the problem that we currently can't differentiate the (MQR or PBS) sub-requirement from the PBS and MQR subrequirements?

@einc It's currently 3 slots because the slots are PSB: 2 courses, MQR: 1 course, PSB/MQR: 1 course.

tcho6319 (approved):
LGTM. Thanks Sam! 

>@tcho6319 In terms of the 3 slot vs 4 slot issue in the PBS/MQR requirement, I believe we should actually use 3 slots, since logically there are 3 slots. Currently, it's slightly broken because the frontend code assumes that 1 slot can only be filled by 1 course. This is something we might want to change. Loop in @einc for potential design work.

I think this frontend code fix should be done in a separate PR. I will make a Story item for it on Notion.

PR #346, opened 2021-03-06
Add modal refactor (V3?)

hahnbeelee:
### Summary
- Made the add modal look more like [Ein's newest designs](https://www.figma.com/file/81R3CvQY1EgxFZSxtEJNrj/Ein-Working-File?node-id=540%3A407)
- Created a dropdown for non-auto-fulfilled classes
- Added empty state
- When you select a class, the class name is no longer an editable input (last time i tried to do this though it caused some errors with the input which is why it wasn't implemented... it's looking like it's acting normal as of now, but i haven't extensively tested)

### Screenshots / Flow
image
image
image
image

Empty state:
image

### Test Plan
- Add classes + make sure the requirements side-bar fills up requirements accordingly!

### Notes
Differences between designs @einc 
- I added an option for adding no potential/related requirements in case the user doesn't want to add a class for any of them
- After you add a potential/related requirement it becomes bold so that it affirms the user it selected the requirement.


src/components/Modals/NewCourse/RequirementsDropdown.vue
```diff
@@ -0,0 +1,171 @@
...
+      data-toggle="dropdown"
+      aria-haspopup="true"
+      :aria-expanded="showDropdown"
+      v-if="potentialRequirements.some(req => req.id === selectedID)"
+    >
+      <div class="warning-row">
+        <img class="warning-icon" src="@/assets/images/warning.svg" alt="warning-icon" />
+        {{ selected }}
+      </div>
+      <drop-down-arrow :isFlipped="showDropdown" :fillColor="'#148481'" />
```
SamChou19815, line 16: new constant?

src/components/Modals/NewCourse/RequirementsDropdown.vue
```diff
@@ -0,0 +1,171 @@
...
+    width: 93%;
+    li {
+      a {
+        padding: 13px 8px 8px;
+        display: block;
+        width: 100%;
+        &:active,
+        &:focus,
+        &:hover {
+          background: rgba(50, 160, 242, 0.1);
```
SamChou19815, line 145: new constant?

SamChou19815 (commented):
Some nits on color constants.

src/components/Modals/NewCourse/SelectedRequirementEditor.vue
```diff
@@ -88,15 +65,31 @@ export default Vue.extend({
...
     },
-    editRequirementsText(): string {
-      return this.potentialRequirements.length !== 0
-        ? 'Add these Requirements'
-        : 'Edit Requirements';
+    nonAutoRequirements(): { readonly id: string; readonly name: string }[] {
+      return this.relatedRequirements.concat(this.potentialRequirements);
+    },
+    nonAutoRequirementsText(): string {
+      if (this.selectedRequirementID === '') this.nonAutoRequirements.map(it => it.name).join(', ');
```
benjamin-shen, line 78: Do you need to `return`?

src/components/Modals/NewCourse/RequirementsDropdown.vue
```diff
@@ -0,0 +1,171 @@
...
+      v-else
+      class="dropdown"
+      type="button"
+      @click="showDropdown = !showDropdown"
+      data-toggle="dropdown"
+      aria-haspopup="true"
+      :aria-expanded="showDropdown"
+    >
+      {{ selected }}
+      <drop-down-arrow :isFlipped="showDropdown" :fillColor="'#148481'" />
```
benjamin-shen, line 28: As Sam said, constant here?

src/components/Modals/NewCourse/SelectedRequirementEditor.vue
```diff
@@ -1,75 +1,52 @@
...
-      <div class="newCourse-title">This class can auto-fulfill the following requirement(s):</div>
       <div class="newCourse-requirements-container">
         <div class="newCourse-requirements">
           {{ requirementsThatAllowDoubleCounting.join(', ') }}
         </div>
       </div>
     </div>
-    <div v-if="relatedRequirements.length > 0">
+    <div v-else class="newCourse-requirements-container newCourse-requirements">
+      This class does not automatically fulfill any requirements
```
benjamin-shen, line 14: Is this text supposed to be blue? Also, should there be a period?

benjamin-shen (commented):
Looks good! I left some comments.

A few more high-level comments:
- I'm personally not in love with the empty dropdown text, especially after choosing None of the above. (maybe designers have a better suggestion?)
- Can the dropdown be closed when you click outside of it?
- The dropdown options width is a bit wonky on mobile. This is probably lower priority, but it'd be sick if you could fix it.

src/components/Modals/NewCourse/SelectedRequirementEditor.vue
hahnbeelee: i realized @einc doesnt use periods lol. i agree that there should be a period. Ein, is it cool if i add a period or is this a design decision?

einc: no lol that's just my bad typing habits. yes haha pls add periods or capitalize stuff as normal <3

einc:
> * I'm personally not in love with the empty dropdown text, especially after choosing None of the above. (maybe designers have a better suggestion?)

Can we have the placeholder dropdown be "Select Requirement (optional)" looking like this:
(and get rid of the None option?)
Screen Shot 2021-03-07 at 12 27 41 AM

einc:
Screen Shot 2021-03-07 at 12 33 03 AM
I'm a bit confused why the Computer Science Core requirement isn't automatically fulfilled here? [this is CS major in CoE]

einc:
sorry, I should have changed this in the designs before, but could we also change the "We cannot check the requirements marked with a warning sign, so double check before selecting" text to "**We cannot accurately check the requirements marked with the warning icon, so double check before selecting."**

einc:
HAHNBEE!! THIS IS 🔥!!! left some small comments but looks pretty good overall!
woohoo on also getting the class name to be non-editable 🥳
thanks for helping getting the add modal right and making it make sense!!

hahnbeelee:
> > * I'm personally not in love with the empty dropdown text, especially after choosing None of the above. (maybe designers have a better suggestion?)
> 
> Can we have the placeholder dropdown be "Select Requirement (optional)" looking like this:
> (and get rid of the None option?)
> Screen Shot 2021-03-07 at 12 27 41 AM

If we don't have a 'None' option there's no way for the user to unselect a requirement. So I feel like we should keep the None option to let the user revert their choice if they want to?

hahnbeelee:
> Screen Shot 2021-03-07 at 12 33 03 AM
> 
> I'm a bit confused why the Computer Science Core requirement isn't automatically fulfilled here? [this is CS major in CoE] It's showing up in the dropdown but isn't self check?

hm yeah idk why that's happening. i feel like it isn't me though and is most likely the requirements calculations happening in the backend? I'll take a look at it after this PR.

SamChou19815:
> Screen Shot 2021-03-07 at 12 33 03 AM
> 
> I'm a bit confused why the Computer Science Core requirement isn't automatically fulfilled here? [this is CS major in CoE] It's showing up in the dropdown but isn't self check?

Because it's not marked as allow double counting yet

hahnbeelee:
Most recent commit:
- Added scss variable
- Made it look better on mobile
- Click outside closes the dropdown
- Changed the warning text
- Added placeholder text vvv
image

src/components/Modals/NewCourse/RequirementsDropdown.vue
```diff
@@ -0,0 +1,188 @@
...
+    </ul>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue, { PropType } from 'vue';
+
+import { clickOutside } from '@/utilities';
+import DropDownArrow from '@/components/DropDownArrow.vue';
+import { emGreen } from '@/assets/scss/_variables.scss';
```
SamChou19815, line 67: to resolve the type error, add it to the type declaration in `src/assets/scss/_variables.scss.d.ts`

src/components/Modals/NewCourse/SelectedRequirementEditor.vue
```diff
@@ -1,75 +1,52 @@
...
+    </div>
+    <div v-if="nonAutoRequirements.length > 0">
       <div class="newCourse-title">This class can fulfill the following requirement(s):</div>
-      <div v-if="!editMode" class="newCourse-requirements-container">
-        <div class="newCourse-requirements">
-          {{ relatedRequirements.map(it => it.name).join(', ') }}
+      <div v-if="!editMode">
+        <div class="newCourse-requirements-container">
+          <!-- eslint-disable-next-line vue/no-v-html  -->
+          <div class="newCourse-title" v-html="nonAutoRequirementsText"></div>
```
tcho6319, line 21: Can you use a different approach instead of a `v-html` directive? As addressed in #298, v-html directives could possibly lead to a [security issue](https://vuejs.org/v2/guide/security.html#Injecting-HTML) with HTML injection

src/components/Modals/NewCourse/SelectedRequirementEditor.vue
```diff
@@ -1,75 +1,52 @@
...
+          :potentialRequirements="potentialRequirements"
+          :selectedID="selectedRequirementID"
+          @on-selected-change="toggleSelectRequirement"
         />
       </div>
     </div>
-    <div v-if="!editMode" class="newCourse-link" @click="toggleEditMode()">
-      {{ editRequirementsText }}
-    </div>
+    <div v-if="!editMode" class="newCourse-link" @click="toggleEditMode()">Edit Requirements</div>
```
tcho6319, line 38: Can we make this a button? The Add Modal is actually almost all keyboard accessible 🥰 except for this part!

tcho6319 (changes requested):
This looks great Hahnbee! Left a few suggestions in addition to the one below:
1. Can we document what `relatedRequirements` and `potentialRequirements` are in `NewCourseModal.vue` and what `nonAutoRequirements` are in `SelectedRequirementEditor.vue`?

src/components/Modals/NewCourse/SelectedRequirementEditor.vue
hahnbeelee: I actually can't think of any other way to do this unfortunately :/ . Since, I'm logically rendering the emboldened text based on data.

tcho6319: @hahnbeelee Just pushed my commit for the implementation of an alternative solution. Let me know if you're ok with this. If you are, then this PR should be g2g!

src/components/Modals/NewCourse/SelectedRequirementEditor.vue
```diff
@@ -1,102 +1,115 @@
...
+                index === nonAutoRequirementsTextArray.length - 1
+                  ? reqName.name
+                  : reqName.name + ', '
+              }}</strong>
+              <span v-else>{{
+                index === nonAutoRequirementsTextArray.length - 1
+                  ? reqName.name
+                  : reqName.name + ', '
+              }}</span>
+            </span>
```
tcho6319, line 32: @hahnbeelee I replaced `nonAutoRequirementsText` with `nonAutoRequirementsTextArray` which is an array of objects. Each object has the name of the requirement and a boolean for whether or not it was selected (and should have the strong tag). Using those fields, I conditionally apply the strong tag or span tag. I take a look at the index of `reqName` in `nonAutoRequirementsTextArray` to see if we need to have a `', '` or not.

hahnbeelee: Looks good to me!

PR #348, opened 2021-03-07
fix bug with ap and ib tests not giving correct credits

jason-tung:
### Summary 
AP and IB tests give the correct amount of credits now.

https://www.notion.so/courseplan/Wrong-number-of-credits-for-AP-IB-exam-58589b3b5f0746f69909dae9f41afb64

### Test Plan 

Playtest by trying different test scores that would give different credit amounts and verify it is correct.

image
image

### DIFF
--- a/src/components/Modals/Onboarding/OnboardingTransfer.vue
+++ b/src/components/Modals/Onboarding/OnboardingTransfer.vue
@@ -165,7 +165,7 @@
 
 <script lang="ts">
 import Vue, { PropType } from 'vue';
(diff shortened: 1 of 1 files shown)


SamChou19815:
Since we have #349, why do we still need this?

jason-tung:
> Since we have #349, why do we still need this?

There are plans to bring the credits column back post-launch

tcho6319 (approved):
LGTM thanks Jason!

PR #350, opened 2021-03-07
Save user choice for self check, clean up some semester-related firestore modifiers

benjamin-shen:
### Summary 

This pull request finishes the self-check implementation ([notion](https://www.notion.so/courseplan/Save-User-Choice-for-Self-Check-and-Add-Modal-f99924771ce643d0a6299a89675f56fd)) and cleans up some of the components / semester-related firestore code.

Functionality:
- [x] Save user choice in selectable requirements when adding existing course and when adding new course through the add modal
- [x] When deleting a course, also delete it from selectableRequirementChoices
- [x] Properly read from global data store to generate list of CompletedSubReqCourses (this is implemented with a derived function)
- [x] Remove already-selected courses from self-check dropdown

Behind-the-scenes:
- [x] Refactor deleteCourseFromSemesters to be a global-firestore-data function
- [x] Break down chooseSelectableRequirementOption into friendlier functions that modify selectableRequirementChoices
- [x] Interdepend add/delete course functions with add to / delete from selectableRequirements to reduce code repetition

Code quality and bug fixes:
- [x] Fix small issues:
  - CURLink is no longer broken
  - "CS1110" placeholder changed to "CS 1110"
- [x] Get rid of unnecessary emit chains:
  - deleteCourseFromSemesters
  - addCourse (in self-check)
- [x] Add missing gtags (that I missed in #342)

### Test Plan 

This PR touches almost all of the semester functions in global-firestore-data; along with testing that the self-check add/delete functionality persists, please make sure the semester/course add/delete functionality still works as intended.

### Notes 
- Instead of writing complicated code for `store.state.selectableRequirementChoices`, you can directly reference `store.state.derivedSelectableRequirementData.requirementToCoursesMap` to get the user's selected courses for a given requirement id.
- After this PR is merged with master, I will delete everyone's data in the `user-selectable-requirement-choices` firestore collection.

### DIFF
--- a/src/components/Semester/Semester.vue
+++ b/src/components/Semester/Semester.vue
@@ -324,7 +324,7 @@ export default Vue.extend({
       this.openConfirmationModal(`Added ${courseCode} to ${this.type} ${this.year}`);
     },
     deleteCourse(courseCode: string, uniqueID: number) {
-      deleteCourseFromSemester(this.type, this.year, uniqueID);
+      deleteCourseFromSemester(this.type, this.year, uniqueID, this.gtag);
       // Update requirements menu
       this.openConfirmationModal(`Removed ${courseCode} from ${this.type} ${this.year}`);
     },
--- a/src/components/Semester/Semester.vue
+++ b/src/components/Semester/Semester.vue
@@ -131,7 +131,6 @@ import {
   editSemester,
   addCourseToSemester,
   deleteCourseFromSemester,
-  chooseSelectableRequirementOption,
 } from '@/global-firestore-data';
 import { cornellCourseRosterCourseToFirebaseSemesterCourse } from '@/user-data-converter';
 import store from '@/store';
@@ -319,13 +318,7 @@ export default Vue.extend({
     },
     addCourse(data: CornellCourseRosterCourse, requirementID: string) {
       const newCourse = cornellCourseRosterCourseToFirebaseSemesterCourse(data);
-      addCourseToSemester(this.type, this.year, newCourse);
-      if (requirementID) {
-        chooseSelectableRequirementOption({
-          ...store.state.selectableRequirementChoices,
-          [newCourse.uniqueID]: requirementID,
-        });
-      }
+      addCourseToSemester(this.type, this.year, newCourse, requirementID, this.gtag);
 
       const courseCode = `${data.subject} ${data.catalogNbr}`;
       this.openConfirmationModal(`Added ${courseCode} to ${this.type} ${this.year}`);
--- a/src/components/Semester/Semester.vue
+++ b/src/components/Semester/Semester.vue
@@ -181,7 +181,7 @@ export default Vue.extend({
   props: {
     semesterIndex: { type: Number, required: true },
     type: {
-      type: String as PropType<'Fall' | 'Spring' | 'Winter' | 'Summer'>,
+      type: String as PropType<FirestoreSemesterType>,
       required: true,
     },
     year: { type: Number, required: true },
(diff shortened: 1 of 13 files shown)


src/components/Semester/Semester.vue
```diff
@@ -131,7 +131,6 @@ import {
   editSemester,
   addCourseToSemester,
   deleteCourseFromSemester,
-  chooseSelectableRequirementOption,
 } from '@/global-firestore-data';
 import { cornellCourseRosterCourseToFirebaseSemesterCourse } from '@/user-data-converter';
 import store from '@/store';
```
SamChou19815, line 136: remove this import to fix linter warning

src/global-firestore-data.ts
```diff
@@ -141,14 +159,36 @@ export const chooseToggleableRequirementOption = (
...
+    [courseUniqueID]: requirementID,
+  });
+};
+
+export const deleteCourseFromSelectableRequirements = (courseUniqueID: number): void => {
+  chooseSelectableRequirementOption(
+    Object.assign(
+      {},
+      ...Object.entries(store.state.selectableRequirementChoices)
+        .filter(([k, _]) => parseInt(k, 10) !== courseUniqueID)
```
SamChou19815, line 186: Simply do `.filter(([k]) => parseInt(k, 10) !== courseUniqueID)`

src/store.ts
```diff
@@ -22,12 +22,22 @@ type DerivedCoursesData = {
   readonly courseToSemesterMap: Readonly<Record<number, FirestoreSemester>>;
 };
 
+/**
+ * Some course data that can be derived from selectable requirement choices, but added to the global store
+ * for efficiency and ease of access.
+ */
+type DerivedSelectableRequirementData = {
+  // Mapping from requirement ID to the user-selected courses that fulfill the requirement.
+  readonly requirementToCoursesMap: Readonly<Record<string, FirestoreSemesterCourse[]>>;
```
SamChou19815, line 31: Make the array readonly to avoid accidental mutation

SamChou19815 (commented):
Some nits.

Comments on the `DerivedSelectableRequirementData`. It appears to me that it's more useful for self-check requirements, since they are never part of the requirement graph that allow you to look up the same data. Maybe you can add more comments to clarify this?

willespencer (changes requested):
Left a small nit but otherwise the code looks good to me! Also ran through your test plan and most of it works as expected 😄, just got a few bugs below.

I'm not certain if it is in the scope of this PR, but when adding courses to a self check requirement, they do not always increase the class counter? So for example, in this picture, I added CS 1110 from my plan, and INFO 4300 + INFO 3300 from the self check add modal, but only INFO 3300 incremented the requirement from "0/3 courses" to "1/3 courses."
(Also, should CS 1110 be assignable at all since the warning implies that we only allow 3000+ courses?)

image

Somewhat related to the above, after deleting INFO 3300 from the self check requirement, the self check does reset. However, adding the course back to my plan, even after a refresh, has the counter listed as "1/3 courses" even though I did not add the course back to the requirement.

image

After deleting a semester, courses assigned to the self check requirement stay there with a default color until the page is refreshed or one of the "reset" buttons is pressed. Other non-self check requirements seem to reset when a semester is deleted, so this should be fixable?

image

benjamin-shen:
Thanks @willespencer!

> I'm not certain if it is in the scope of this PR, but when adding courses to a self check requirement, they do not always increase the class counter? So for example, in this picture, I added CS 1110 from my plan, and INFO 4300 + INFO 3300 from the self check add modal, but only INFO 3300 incremented the requirement from "0/3 courses" to "1/3 courses."
> (Also, should CS 1110 be assignable at all since the warning implies that we only allow 3000+ courses?)

and

> Somewhat related to the above, after deleting INFO 3300 from the self check requirement, the self check does reset. However, adding the course back to my plan, even after a refresh, has the counter listed as "1/3 courses" even though I did not add the course back to the requirement.

are not in the scope of this PR. I believe this also didn't work before my PR. I can work on fixing it later (probably with your help lol).

> After deleting a semester, courses assigned to the self check requirement stay there with a default color until the page is refreshed or one of the "reset" buttons is pressed. Other non-self check requirements seem to reset when a semester is deleted, so this should be fixable?

Thanks for catching this! Should be fixed now.

willespencer (approved):
The fix looks good to me! And yeah I'll make a note of those first two on my self check item for this sprint and we can discuss who & how to fix them later!

PR #351, opened 2021-03-07
Consider sub-requirement progress in add modal choices

SamChou19815:
### Summary 

This PR addresses the problem mentions in [this PR comment](https://github.com/cornell-dti/course-plan/pull/344#issuecomment-791846707), where the current add modal choices completely disregards whether adding a course can help increasing the requirement fulfillment progress. This PR implements the proposal I mentioned in [the follow-up PR comment](https://github.com/cornell-dti/course-plan/pull/344#issuecomment-792067796), in which we will compute a new stat with the course added, and check whether it can increase the sub-requirement progress. If not, then the requirement won't show up as a choice.

To be able to call the function, we must create a `CourseTaken` object with real credits information. Therefore, we must generate that info when we are selecting the courses, so we must use the full course json instead of the simplified one. I switched the course selector to use the full course json, and after that the simplified course json becomes unused, so I deleted it.

### Test Plan 

Now the bug is fixed (when I have two math classes, adding one more math class won't cause the MQR requirement to show up):
Screen Shot 2021-03-07 at 15 55 41


src/assets/courses/typed-full-courses.ts
```diff
@@ -4,13 +4,15 @@ const fullCoursesJsonWithStringKey: Readonly<
   Record<string, readonly CornellCourseRosterCourse[]>
 > = json;
 
-const fullCoursesJson: Readonly<
-  Record<number, readonly CornellCourseRosterCourse[]>
-> = Object.fromEntries(
+export type FullCourseJson = Readonly<Record<number, readonly CornellCourseRosterCourse[]>>;
+
+export const fullCoursesJson: FullCourseJson = Object.fromEntries(
```
tcho6319, line 9: Can we add a comment describing what `fullCoursesJson` is?

src/assets/courses/typed-full-courses.ts
```diff
@@ -4,13 +4,15 @@ const fullCoursesJsonWithStringKey: Readonly<
...
   Object.entries(fullCoursesJsonWithStringKey).map(([stringCourseID, courses]) => [
     parseInt(stringCourseID, 10),
     courses,
   ])
 );
 
-export default fullCoursesJson;
+export const fullCoursesArray: readonly CornellCourseRosterCourse[] = Object.values(
```
tcho6319, line 16: Similarly for this one

src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -17,47 +17,44 @@
         :class="['search-result', currentFocus === index ? 'autocomplete-active' : '']"
         @click="selectCourse(matchingCourse)"
       >
-        {{ matchingCourse.title }}
-        <input type="hidden" :value="matchingCourse.title" :name="matchingCourse.roster" />
```
tcho6319, line 21: I think `<input>` here was for accessibility purposes. Is it ok to remove?

tcho6319 (changes requested):
Thanks Sam! I left some comments below:
1. I would check out `NewSelfCheckCourseModal.vue` b/c I think deleting `courses.json` might have had a side effect for that component.

https://user-images.githubusercontent.com/54298311/110390288-c1eaa580-8033-11eb-9eb4-4040fa087f5b.mov

src/components/Modals/NewCourse/CourseSelector.vue
SamChou19815: I think it's not. It was there back when we have DOM manipulation and we use the input to read out the data we need instead of using emit.
The type is hidden, so the semantics is that it will also be hidden when viewed by the accessibility software, so there should be no impact on a11y.

SamChou19815:
@tcho6319 fixed the broken self-check add modal and added some doc for full course json

tcho6319 (approved):
LGTM! Ty Sam 😀

PR #355, opened 2021-03-13
maxBottomBarTabs = 1 for small screens

SamChou19815:
### Summary 

According to our discussion in Slack, we decided to make maxBottomBarTabs be 1 when the screen is small.

### Test Plan 

Screen Shot 2021-03-12 at 21 14 28

### DIFF
--- a/src/containers/Dashboard.vue
+++ b/src/containers/Dashboard.vue
@@ -151,7 +151,11 @@ export default Vue.extend({
     resizeEventHandler() {
       this.isMobile = window.innerWidth <= 440;
       this.isTablet = window.innerWidth <= 878;
-      this.maxBottomBarTabs = window.innerWidth <= 1347 ? 2 : 4;
+      if (window.innerWidth <= 1347) {
+        this.maxBottomBarTabs = window.innerWidth <= 480 ? 1 : 2;
+      } else {
+        this.maxBottomBarTabs = 4;
+      }
       this.updateSemesterView();
     },
     toggleRequirementsBar() {
--- a/src/containers/Dashboard.vue
+++ b/src/containers/Dashboard.vue
@@ -76,6 +76,31 @@ import surfing from '@/assets/images/surfing.svg';
 
 import store, { initializeFirestoreListeners } from '@/store';
 import { immutableBottomBarState } from '@/components/BottomBar/BottomBarState';
+import {
+  smallBreakpoint,
+  mediumBreakpoint,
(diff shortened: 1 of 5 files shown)


src/containers/Dashboard.vue
```diff
@@ -151,7 +151,11 @@ export default Vue.extend({
     resizeEventHandler() {
       this.isMobile = window.innerWidth <= 440;
       this.isTablet = window.innerWidth <= 878;
-      this.maxBottomBarTabs = window.innerWidth <= 1347 ? 2 : 4;
+      if (window.innerWidth <= 1347) {
+        this.maxBottomBarTabs = window.innerWidth <= 480 ? 1 : 2;
```
tcho6319, line 155: Can we define these magic numbers as constants somewhere? I know that most of these values refer to the breakpoint variables:
https://github.com/cornell-dti/course-plan/blob/875d2d17ce41b1e24604128017adfd9a7bf814b4/src/assets/scss/_variables.scss#L37-L40

tcho6319 (changes requested):
Thanks Sam! I left a suggestion for some of the magic numbers we're using. Also, the CI check seems to be failing during the TypeCheck. Once those points are addressed, this PR should be g2g!

src/containers/Dashboard.vue
```diff
@@ -76,6 +76,31 @@ import surfing from '@/assets/images/surfing.svg';
...
+  veryLargeBreakpoint.substring(0, veryLargeBreakpoint.length - 2),
+  10
+);
+
+const getMaxButtonBarTabs = () => {
+  if (window.innerWidth <= smallBreakpointPixels) {
+    return window.innerWidth <= veryLargeBreakpointPixels ? 1 : 2;
+  }
+  return 4;
+};
```
tcho6319, line 103: Wouldn't we want to swap for `veryLargeBreakpointPixels` and `smallBreakpointPixels` here?

The behavior is a little wonky:

https://user-images.githubusercontent.com/54298311/111017380-2f266f80-8381-11eb-9b22-4b1f4860e81e.mov

tcho6319 (changes requested):
Almost there! Found a bug from the latest changes.

src/containers/Dashboard.vue
SamChou19815: @tcho6319 fixed now

tcho6319 (approved):
LGTM 😎

PR #356, opened 2021-03-13
Walkthrough update

willespencer:
### Summary 

Update the walkthrough steps to highlight new CoursePlan features, update the styling to match the finalized designs, and fix some walkthrough bugs. (See [notion](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8))

image

image

- [x] Updated steps to focus on requirements menu
- [x] Wrote specific case to open requirements menu and PE requirement for second step
- [x] Updated TourWindow designs with new text/assets for the first and final modals
- [x] Updated tooltip designs (button styling, progress count, etc.)
- [x] Fixed bug where walkthrough would pop up every time you clicked finish on the last page of the onboarding modal
- [x] Fixed styling bug where the semester view would show up under the walkthrough if large enough, see below image

image

Remaining TODOs:

- [x] Update styling again to match designs updated after previous commits (mainly `TourWindow` changes and adding a 4th step)
- [ ] Make step 1 and 2 of the tooltips more seamless/differentiated. I suggested to @einc only having the PE course section highlighted in the second step, but was unable to get a tooltip to show up on that subrequirement div, as tooltips only show up if they are in the html when the tour is started. This tooltip would not be bound to that html element until the reqs menu was opened fully, and thus did not show up in the tour even after opened. Could explore more.
- [x] Was getting the follow error in my build until I set the tour object to type `any` instead of the default `IntroJS`: `Property '_currentStep' does not exist on type 'IntroJs'.` Is there a better way of handling this? I also had to put a comment to stop eslint from breaking on the use of an underscore property, but that seems to be the way to access it from IntroJS.

### Test Plan 

1. Confirm that the onboarding modal now cannot be opened by finishing onboarding a second time, but deleting your firestore data and going through onboarding will have it show up.

To test the rest of the changes, change the if statement in the `endOnboarding()` function of `Dashboard.vue` to `if (true)` locally so that you can open the walkthrough by clicking through the Onboarding modal.

2. Confirm that with semesters extending off the bottom of the screen, the walkthrough modal now displays correctly
3. Click through walkthrough steps and make sure it works as expected and looks like designs
4. Skip the walkthrough during the PE step and make sure you can collapse and reopen that requirement


tcho6319 (changes requested):
Thanks Will! Excited to see the Walkthrough coming together for launch 🤩

Noted a couple suggestions/fixes:
1. The color for the title of each tooltip in the Walkthrough is a bit off. I believe it should be `#3D3D3D` (there's an scss variable). 
Screen Shot 2021-03-14 at 12 20 09 PM

2. The Walkthrough screen for Adding Courses is a bit off. I have a darker background color instead of our standard `#F6FAFC`. Also, I noticed how it highlighted the "+ Course" div and not the semester div. 
Screen Shot 2021-03-14 at 12 23 25 PM
Screen Shot 2021-03-14 at 12 30 34 PM

3. I couldn't seem to tab through the walkthrough. Can we make sure that the buttons are `<button>` tags and not divs or is this something that the library can't support?

4. I believe we'll have a more limited walkthrough for mobile (i.e. the breakpoints where we don't show req bar on the side). This task will probably go to @SamChou19815 since he's working on the mobile update and you're already assigned to finishing up self check. 

> Make step 1 and 2 of the tooltips more seamless/differentiated. I suggested to @einc only having the PE course section highlighted in the second step, but was unable to get a tooltip to show up on that subrequirement div, as tooltips only show up if they are in the html when the tour is started. This tooltip would not be bound to that html element until the reqs menu was opened fully, and thus did not show up in the tour even after opened. Could explore more.

Are we able to have the PE subreq opened by default at the beginning of the tour?

> Was getting the follow error in my build until I set the tour object to type any instead of the default IntroJS: Property '_currentStep' does not exist on type 'IntroJs'. Is there a better way of handling this? I also had to put a comment to stop eslint from breaking on the use of an underscore property, but that seems to be the way to access it from IntroJS.

I'll explore this!

> 4. Skip the walkthrough during the PE step and make sure you can collapse and reopen that requirement

I don't think I'm able to skip the walkthrough at that point. Can you clarify?

src/components/Requirements/Requirements.vue
```diff
@@ -107,13 +118,13 @@ type Data = {
   showAllCourses: ShowAllCourses;
   shouldShowAllCourses: boolean;
   showAllPage: number;
+  tourStep: number;
 };
 
 // This section will be revisited when we try to make first-time tooltips
-const tour = introJs().start();
+const tour: any = introJs().start();
```
SamChou19815, line 125: why we do need any here

willespencer:
> 2. The Walkthrough screen for Adding Courses is a bit off. I have a darker background color instead of our standard `#F6FAFC`. 

Confirmed with @einc that the background color should be the same as the other modals!

> > Make step 1 and 2 of the tooltips more seamless/differentiated. I suggested to @einc only having the PE course section highlighted in the second step, but was unable to get a tooltip to show up on that subrequirement div, as tooltips only show up if they are in the html when the tour is started. This tooltip would not be bound to that html element until the reqs menu was opened fully, and thus did not show up in the tour even after opened. Could explore more.
> 
> Are we able to have the PE subreq opened by default at the beginning of the tour?

Per discussion with @einc and @tcho6319, keeping it as is as we do not want the PE courses the focus of the first step.
 
> > 4. Skip the walkthrough during the PE step and make sure you can collapse and reopen that requirement
> 
> I don't think I'm able to skip the walkthrough at that point. Can you clarify?

My bad, there were previously buttons in the introjs modals that let you exit the walkthrough at that point. I wrote a special case to fix the requirements menu and close the PE requirement if a user quit the walkthrough during that step, but since it cannot be skipped anymore, this is no longer an issue.

Everything else I did not respond to is fixed!

src/components/Requirements/Requirements.vue
willespencer: I mentioned this in the last bullet of "Remaining TODOs" above, but I was getting the following error...

`IntroJS: Property '_currentStep' does not exist on type 'IntroJs'.`

Per @tcho6319's suggestion, I was about to remove this `any` type by using `require` instead of `import` syntax for my newest commit, but I am not sure if this is better? Let me know what you think.

SamChou19815: Using require simply suppresses the problem, which is almost same as `any`...
I think it might help to bump introjs and its typedef. In recent 3.0 release, you can get the current step more easily by `introjs().currentStep()`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48969/files

tcho6319 (changes requested):
Almost there! Just left a small nit:
1. There seems to be a border-top and some extra padding.
Screen Shot 2021-03-17 at 2 06 26 PM

willespencer:
> Almost there! Just left a small nit:
> 
> 1. There seems to be a border-top and some extra padding.
>    Screen Shot 2021-03-17 at 2 06 26 PM

Thanks for catching this! It looks like updating the package caused a few design discrepancies, but they should all be fixed now!

tcho6319 (approved):
LGTM! Thanks Will :D

PR #357, opened 2021-03-14
Add information science concentrations

SamChou19815:
### Summary 

Title

### Test Plan 

See https://infosci.cornell.edu/undergraduate/info-sci-majors/ba-information-science-college-arts-sciences/degree-requirements-0 and compare with the code.


src/requirements/data/majors/info.ts
```diff
@@ -58,8 +58,208 @@ const infoRequirements: readonly CollegeOrMajorRequirement[] = [
...
+          ['INFO 2770', 'INFO 3350', 'INFO 4100', 'INFO 4120', 'INFO 4130', 'INFO 4300', 'CS 4740'],
+          [
+            'INFO 3200',
+            'INFO 3561',
+            'INFO 4200',
+            'INFO 4240',
+            'INFO 4250',
+            'INFO 4270',
+            'INFO 4561',
+            '​INFO 4940',
```
tcho6319, line 115: INFO 4940 doesn't appear in the See All for this slot (third slot or C. Big Data Ethics) of the Data Science concentration. Lol I noticed this bc I was trying to track my Data Science concentration fulfillment.

tcho6319 (changes requested):
Thanks Sam! Excited to see INFO concentrations 🤗

I noticed a couple issues that I commented below. I also checked this requirements data for the CALS INFO concentrations and they seem identical to the A&S INFO concentrations so I think it's good for CALS people as well.

src/requirements/data/majors/info.ts
tcho6319: Interesting I get this when I try to add INFO 4940 from the add modal
Screen Shot 2021-03-15 at 10 08 19 AM

einc:
A few concentrations can't be completed, but I think this is fixed in your PR #359 right?

Screen Shot 2021-03-15 at 10 54 34 AM

einc (approved):

src/requirements/data/majors/info.ts
einc: This also happened to me ^ but I see it in the Add Modal/See All for the Behavioral Sciences Concentration

SamChou19815:
Haha looks like the problem of missing classes is caused by the hidden zero-width space `U+200B`:
Screen Shot 2021-03-15 at 13 42 14
They are probably there when I copy paste the names from the website...

src/requirements/data/majors/info.ts
```diff
@@ -58,8 +58,208 @@ const infoRequirements: readonly CollegeOrMajorRequirement[] = [
...
+          ],
+          ['INFO 3300', 'INFO 3350', 'INFO 4120', 'INFO 4130', 'INFO 4300']
+        ),
+        perSlotMinCount: [1, 1, 1, 1],
+      },
+      'Interactive Technologies': {
+        description:
+          'A. Building with Hardware (choose one), ' +
+          'B. Working with Data/Software (choose one), ' +
+          'C. Context/Application Domains (choose one)',
```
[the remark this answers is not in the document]
SamChou19815: Added a comment on this. It turns out it's not easy, since adding CS 2110 here might introduce double counting warning? I added a TODO here.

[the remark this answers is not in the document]
tcho6319: Thanks! Is this something that can be done before launch or would this be more disruptive since we would be changing the format of the toggleable reqs?

src/requirements/data/majors/info.ts
tcho6319: I still can't seem to find INFO 4940 for the Data Science concentration. I don't think the zero-width space for this was removed in your commit.

Also, I wonder why CALS Credits no longer shows up as a possible option for INFO 4940 in the add modal. (This is the add modal I get when I'm an A&S INFO major Data Science concentration).
Screen Shot 2021-03-15 at 7 37 45 PM

tcho6319 (changes requested):
Almost there! Still can't seem to find INFO 4940 in the See All for Data Science concentration.

SamChou19815:
@tcho6319 the problem is fixed now.

> Also, I wonder why CALS Credits no longer shows up as a possible option for INFO 4940 in the add modal. (This is the add modal I get when I'm an A&S INFO major Data Science concentration).

I think it's because A&S does't have CASL credit requirement...

src/requirements/data/majors/info.ts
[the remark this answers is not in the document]
SamChou19815: @tcho6319 It should be pretty straightforward. It requires a slight tweak in requirement algorithm. I might do this live for devs who wants to learn more about requirement algorithm.

tcho6319 (approved):
LGTM ty Sam! Yes, sorry disregard my comment re: CALS credits not showing up as an A&S INFO major 😅  I've been switching b/w colleges while testing so I mixed myself up.

PR #358, opened 2021-03-14
Implement new mobile design for navbar

SamChou19815:
### Summary 

Implement the design [here](https://www.figma.com/file/vz9T397rHdewbB2McmUDEr/Yuxuan-Working-File?node-id=914%3A760).

### Test Plan 

Desktop UI is unchanged:
Screen Shot 2021-03-14 at 16 28 37

Mobile UI:
Screen Shot 2021-03-14 at 16 28 46 1
Screen Shot 2021-03-14 at 16 28 49
Screen Shot 2021-03-14 at 16 28 52
Screen Shot 2021-03-14 at 16 28 57
Screen Shot 2021-03-14 at 16 28 58


src/components/NavBar.vue
```diff
@@ -61,52 +90,71 @@ export default Vue.extend({
...
+
+  .navbar-menu {
+    position: fixed;
+    z-index: 3;
+    left: 0;
+    width: 16.5rem;
+    top: $mobile-navbar-height;
+    bottom: 0;
+    display: none;
+    background: $offWhite;
```
tcho6319, line 149: I think the background color is off. In the designs, the background color is supposed to be `#F7FAFC`. `$offWhite` is `#F8F8F8`.

src/components/NavBar.vue
```diff
@@ -39,20 +63,25 @@ export default Vue.extend({
 <style scoped lang="scss">
 @import '@/assets/scss/_variables.scss';
 
+$icon-size: 2.5rem;
```
tcho6319, line 66: I think `$icon-size` would be useful in _variables.scss since we plan on adding more icons to the nav bar

tcho6319 (changes requested):
Looks great Sam! Thank you 😀  I left some nits below.

1. I wanted to note that the implementation seems to be missing the "Cornell DTI @ 2021". Do you know what the link should be?
Screen Shot 2021-03-15 at 10 15 13 AM

src/components/NavBar.vue
SamChou19815: I think adding here makes more sense, since as you said, there will be more icons in _nav bar_...

tcho6319 (approved):
LGTM! Ty Sam

PR #359, opened 2021-03-15
Fix subrequirement slot with minCount > 1

SamChou19815:
### Summary 

Resolve [this notion item](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=d8736f944ef44d49bf02febd430ccdfc) with option 2.

Background: PBS/MQR requirement with perSlotMinCount will have 3 slots, with minCount to be 2, 1, 1. Our frontend always assumes that one slot can only contain one course. The option 1 in that notion item isn't doable now, since we would need data providing the name for each slot. Option 2 is very doable. In fact, this PR implements option 2 in a very principled way that removes a lot of existing hack.

We first deal with requirements fulfilled with credits. By observation and unit testing, we know that these requirements only has one checker and one slot. Therefore, we can easily special case that.
Requirement fulfilled by slots can be more complicated. However, now we have the perSlotMinCount, we can match the minCount with existing courses and generate variable number of completed and incomplete subrequirement.

E.g. there is a sub-requirement slot with `perSlotMinCount` 3. In the slot, we find there are two courses already taken. Then we can generate two CompletedSubReqCourse and one IncompleteSubReqCourse.

Note that this algorithm automatically deals with the case when we have one slot and multiple courses needed to fulfilled by that slot. In the past, it's handled by `subRequirementProgress: 'any-can-count'` and special case for single-slot requirements, but now it can be elegantly handled by the algorithm described above without any special case.

### Test Plan 

Added test to ensure requirements fulfilled by credits only have one checker.

Screen Shot 2021-03-14 at 20 45 32
Screen Shot 2021-03-14 at 20 45 50

### DIFF
--- a/src/components/Requirements/SubRequirement.vue
+++ b/src/components/Requirements/SubRequirement.vue
@@ -97,7 +97,7 @@
         "
         class="subreqcourse-wrapper"
       >
-        <div v-for="(subReqCourseSlot, id) in subReqCoursesSlots" :key="id">
+        <div v-for="(subReqCourseSlot, id) in requirementCoursesSlots" :key="id">
           <div v-if="subReqCourseSlot.isCompleted" class="completedsubreqcourse-wrapper">
             <completed-sub-req-course
               :subReqCourseId="id"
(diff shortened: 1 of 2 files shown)


src/components/Requirements/SubRequirement.vue
```diff
@@ -233,28 +233,34 @@ export default Vue.extend({
...
+      if (subReqSpec.fulfilledBy === 'credits') {
+        if (this.isCompleted) {
+          slots.push({ isCompleted: true, courses: this.subReq.courses[0] });
         } else {
           slots.push({
             isCompleted: false,
-            courses: generateSubReqIncompleteCourses(allTakenCourseIds, subReqEligibleCourses[i]),
+            courses: generateSubReqIncompleteCourses(allTakenCourseIds, subReqEligibleCourses[0]),
           });
         }
```
tcho6319, line 244: By special casing `'credits'` like this, I think we lost the ability to generate a new IncompleteSubReqCourse slot for credits. Left is this branch's staging and right is dev staging.


https://user-images.githubusercontent.com/54298311/111172601-572df280-857c-11eb-9783-fae72c327056.mov

tcho6319 (changes requested):
Thanks Sam! Found a bug and left a nit.

1. Can we add documentation for `subReqCoursesSlots()`?

tcho6319 (approved):
LGTM! Ty Sam :)

PR #360, opened 2021-03-15
Frontend Cypress Testing

willespencer:
### Summary 

This pull requests sets up frontend Cypress testing on CoursePlan along with several basic tests ([Notion](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=d872d2fbe63e466f8603b9defdcd1793)) and as of 4/18, integrated with the CI.

image

- [x] Set up cypress with a test file, instructions on how to run below.
- [x] Semester tests (add, edit, and delete semesters)
- [x] Change college/major test 
- [x] Add a course test (and confirm requirements can be changed) 

Remaining TODOs:

- [ ] Turn this one test file into a thorough testing suite. See below suggestions
- [x] Better handle the log in page so that the tests do not run until the user is logged in and on the dashboard 
- [x] Handle Google Authentication and integrate Cypress with the CI. Ideally we could have testing credentials that the CI could use to log into CoursePlan, instead of having developers manually run cypress before merging in PRs.
- [x] Change the user ID of the testing account used. It is currently the ID for my cornell email account, but it would be better to use something like our cornell courseplan email

### Test Plan 

Instructions on how to run Cypress (and confirm this PR works)

1. Serve the site (run `npm run serve`)
2. While serving, in another terminal, run `npm run cypress:open`
3. Wait for the cypress program to open, then click on the integration test file `test.spec.ts`
4. ~~On the new browser that opens, press the stop button on the top left, then log in to CoursePlan. (A user is only sent to the log in page the first time they run cypress. If you are having trouble logging in, try refreshing the tests, pressing stop immediately, and then pressing the log in button. If that fails, try restarting cypress)~~
5. ~~Once you are on the dashboard, press the refresh button "Run all Tests" to run the tests (note, the tests will overwrite most of your dev semester data) and confirm they pass~~
6. Cypress will log in and automatically complete all the tests. Confirm that the tests all pass.

*NOTE*: You need a serviceAccount.json in the root of the repo to run locally, which must be downloaded (and not committed) from the dev firebase site.

### Test Plan 2 (as of 4/18)

Now the CI will confirm it works by itself! It runs `npm run cypress:e2e` which both serves the site and runs all the test files in electron. Confirm the CI passes.

### Notes 

If we decide to expand on frontend testing in the future, I suggest that we create multiple integration test files for different components or features to test, and try to add significant tests to test existing features. We should also recommend that when devs add new features, they not only run cypress to confirm that nothing has broken, but add their own tests to confirm that what they added works, and so that if it breaks in the future it will be caught by the tests.

I did not have too much time to add significant testing, but here are some testing suggestions I have based on bugs that have come up in the past or currently untested functionality
- AP/IB Credit (adding it in Onboarding, working as expected in Requirements)
- Multiple major/minor support
- Clicking outside of all modals closes them (except for the first time Onboarding)
- Swim test requirement works as expected
- Course credits and colors can be edited
- Everything works for both normal and compact view
- Mobile & other breakpoints display as expected
- Bottom bar functionality (reviews and other course info show up)
- More extensive requirement selecting and self check testing
- Requirements bar testing (see all, toggle requirements, credit counting, etc.)
- Walkthrough displaying at the correct time and being able to click through it


cypress/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["plugin:cypress/recommended"]
```
SamChou19815, line 2: Can you try to move it to the root config? Otherwise it won't be enforced by CI.

src/components/Modals/Onboarding/OnboardingReview.vue
```diff
@@ -40,13 +40,17 @@
           <div class="onboarding-selectWrapper-review">
             <label class="onboarding-label">College*</label>
             <div>
-              <label class="onboarding-label--review">{{ collegeText }}</label>
+              <label id="onboarding-college" class="onboarding-label--review">{{
```
SamChou19815, line 43: let's avoid id so that they won't accidentally collide to introduce a11y problems. Usually the convention is to use `data-test-id`: https://docs.cypress.io/api/commands/get.html#Find-5-elements-with-the-given-data-attribute

SamChou19815 (changes requested):
Some nits. I want to put on hold here before we find a way to integrate this in CI. Without CI checks, I'm pretty sure these tests can be broken without us noticing very quickly.

cypress/integration/test.spec.ts
```diff
@@ -0,0 +1,170 @@
+// A test file containing a number of basic tests to confirm general frontend functionality works
+// Can and should be expanded on in more files to test more specific functionality and ensure future bugs are caught earlier
+
+it('Visit site logged in', () => {
+  cy.visit('localhost:8080');
+});
+
+// Delete existing semesters to ensure existing data does not mess with tests
+// Note: will fail if the user testing has no semesters, but the other tests will continue as expected
```
tcho6319, line 9: Thank you for the documentation!

tcho6319 (changes requested):
Thank you so much for introducing frontend unit testing to CP, Will! 🤩  My biggest concern is similar to an issue that @SamChou19815 brought up which is using easily changeable CSS attributes like `class` or `id`. Instead of using these volatile attributes which can easily break our tests, we want to create `data-*` attributes like `data-testid` that clearly indicate these attributes are for testing and not styling purposes. Here's documentation for best practices when selecting elements: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements. 

Here's an example.
Instead of what we have with `id="onboarding-college"` below:
https://github.com/cornell-dti/course-plan/blob/516b581fa2ce2907ff770ba5fc359f4a6fa7060f/src/components/Modals/Onboarding/OnboardingReview.vue#L43-L45

we should have an attribute like `data-testid="onboarding-college"` so then we can do
`cy.get('[data-testid=onboarding-college]').click()` and this element selector is more resilient to any styling attribute changes.

package.json
```diff
@@ -19,16 +19,19 @@
     "format": "prettier --write \"**/*.{js,ts,css,scss,html,vue}\"",
     "format:check": "prettier --check \"**/*.{js,ts,css,scss,html,vue}\"",
     "format:check:full": "prettier --check \"**/*.{js,ts,vue,scss,css,html}\"",
-    "test": "jest"
+    "test": "jest",
```
neketka, line 22: Is it necessary to keep this dependency here?

willespencer: Good catch! But yeah even though I'm not using Jest for this front-end testing, we still have jest tests for some of the requirements functionality that runs with `npm run test` so I think we do still need this.

SamChou19815 (approved):

SamChou19815 (approved):
One nit: It would be better to move the test into its own CI file or its own job, so that it can run parallel with other jobs. In this way, we won't significant increase the max CI time.

tcho6319 (changes requested):
Thank you so much Will! I left some nits and clarification questions. Once those are addressed or noted, this PR should be good to go!

> Change the user ID of the testing account used. It is currently the ID for my cornell email account, but it would be better to use something like our cornell courseplan email

Yes, I would be ok with that! The Cornell CP email would be good.

I also agree with your suggestions in the Notes section. When creating the Notion documentation for Frontend Testing, we should incorporate these notes in a "Future Plan" section. Once this PR is merged in, can you create a Fall 2021 Notion Story that includes the bullet points of the additional tests you suggested?

package.json
```diff
@@ -13,15 +13,21 @@
     "build:staging": "vue-cli-service build --mode staging",
     "format": "prettier --write \"**/*.{js,ts,css,scss,html,vue}\"",
     "format:check": "prettier --check \"**/*.{js,ts,css,scss,html,vue}\"",
-    "test": "jest"
+    "test": "jest",
+    "cypress:open": "cypress open",
+    "cypress:run": "cypress run",
+    "cypress:e2e": "concurrently \"npm run serve\" \"wait-on http-get://localhost:8080 && npm run cypress:run\" --kill-others --success first"
```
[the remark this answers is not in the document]
willespencer: I did kind of just grab it from stack overflow so should be good for my understanding too haha

cypress/integration/test.spec.ts
```diff
@@ -0,0 +1,188 @@
+// A test file containing a number of basic tests to confirm general frontend functionality works
+// Can and should be expanded on in more files to test more specific functionality and ensure future bugs are caught earlier
```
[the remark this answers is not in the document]
willespencer: Changing this, but out of curiosity, why is this better than just having two comments?

cypress.json
```diff
@@ -0,0 +1,5 @@
+{
+  "env": {
+    "TEST_UID": "bfaEzfuRHrSfHBdfGETAuXBYsaK2"
```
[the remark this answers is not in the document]
willespencer: I don't believe comments are allowed in JSON files, so added a comment in the test file where the UID is used instead!

cypress/integration/test.spec.ts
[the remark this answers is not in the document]
tcho6319: I believe the docstring is useful here since we're documenting the overall function of this file.

.gitignore
```diff
@@ -22,3 +22,6 @@ yarn-error.log*
 *.njsproj
 *.sln
 *.sw?
+
+# Firebase service account, used for things like logging in for Cypress tests
+serviceAccount.json
```
tcho6319, line 27: Good idea since we don't want to accidentally commit this!

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
tcho6319, line 5: Nit: add newline at end of file

tcho6319 (approved):
Left a small nit but LGTM! Excited to have frontend testing :) Thanks Will!

PR #363, opened 2021-03-16
[EZ] 2/n Accessibility - Page regions

hahnbeelee:
### Summary
Added page regions
image

### Test Plan
- Make sure frontend still looks fine
- Make sure that the page regions I added make sense (refer to above picture)


SamChou19815 (approved):
LGTM. You just need to merge in master so that the CI tests can pass.

PR #364, opened 2021-03-18
[EZ] Navbar Z-index bug fix

hahnbeelee:
### Summary 
I think changing the navbar from div to nav did something funky.
Before
image
After
image

### Test Plan
- Make sure there are no gimmicky frontend things going on

PR #365, opened 2021-03-18
4/n Accessibility - Heading structure

hahnbeelee:
### Summary 
Added global headers to add meaningful sequence and a heading structure.
image

### Test Plan 
- ensure there are no visual inconsistencies

### Notes
Things I found while doing this:
- [ ] Edit Semester modal needs to be refactored s.t. it follows the same format as all the other modals.

To-Do
- [ ] change divs to buttons
- [ ] accessible dropdown component
- [ ] accessible + global button component
- [ ] onboarding needs to have its own set of headings
- [ ] aria tags
- [ ] more keyboard accessibility (should happen gradually)


src/components/Modals/EditSemester.vue
```diff
@@ -135,13 +128,14 @@ export default Vue.extend({
...
-    color: $sangBlue;
+    height: 1.8rem;
+    color: $activeGray;
     border-radius: 3px;
-    border: 1px solid $sangBlue;
+    border: 1px solid $primaryGray;
     background-color: $white;
     display: flex;
     justify-content: center;
+    align-items: center;
```
tcho6319, line 138: I believe it should be `color: $sangBlue` and `border: 1px solid $sangBlue`. Is there a reason why the colors were changed?

This is an example of a modal design from Figma:
Screen Shot 2021-03-18 at 11 02 11 AM

tcho6319 (changes requested):
Thanks Hahnbee! Left some nits and wanted to note an issue.

1. Since the `z-index: 2` for the `.navbar` in this feature PR's staging site (left), modals are covered by the nav bar. This seems to be resolved in master (right) because `z-index: 1`. However, this change also introduced a new bug as mentioned in https://github.com/cornell-dti/course-plan/pull/364#issuecomment-801962078 where the nav menu on mobile disappears. I've created a Notion bug item: [Funky z-index for Nav Bar (impact on mobile)](https://www.notion.so/courseplan/Funky-z-index-for-Nav-Bar-impact-on-mobile-b52786ee0ba44e68b1c1c8558023bdb9). This bug does not have to be resolved here and may be delegated to another dev but just wanted to lyk!
Screen Shot 2021-03-18 at 11 04 57 AM

For the remaining TODO items you listed in your summary, would you like more help or do you think you can finish them this Sprint?

hahnbeelee:
> Thanks Hahnbee! Left some nits and wanted to note an issue.
> 
> 1. Since the `z-index: 2` for the `.navbar` in this feature PR's staging site (left), modals are covered by the nav bar. This seems to be resolved in master (right) because `z-index: 1`. However, this change also introduced a new bug as mentioned in [#364 (comment)](https://github.com/cornell-dti/course-plan/pull/364#issuecomment-801962078) where the nav menu on mobile disappears. I've created a Notion bug item: [Funky z-index for Nav Bar (impact on mobile)](https://www.notion.so/courseplan/Funky-z-index-for-Nav-Bar-impact-on-mobile-b52786ee0ba44e68b1c1c8558023bdb9). This bug does not have to be resolved here and may be delegated to another dev but just wanted to lyk!
>    Screen Shot 2021-03-18 at 11 04 57 AM
> 
> For the remaining TODO items you listed in your summary, would you like more help or do you think you can finish them this Sprint?

I have a plan for everything except the aria tags. If somebody has bandwidth and has time to do that that'd be great! But other than that I think my research + gained knowledge will make me the most suited to make those changes + most efficient.

tcho6319 (approved):
LGTM! Ty Hahnbee :)

PR #367, opened 2021-03-18
Remove all hardcoded width in Course.vue and Semester.vue

SamChou19815:
### Summary 

Prepare for semester view mobile design. This PR removes the hardcoded width in the current css.

The hardcoded ones are mostly used to place various elements at the correct place. However, this is the incorrect way of writing css since it makes it much harder to make it responsive on mobile: you need to change all the width values instead of only a few from the top-level. I replaced the absolute child width mostly with default width, but instead try to constraint the width on the parents. i.e. pick a correct width for containers of course component instead of the course component itself. 

### Test Plan 

Screen Shot 2021-03-18 at 19 27 58
Screen Shot 2021-03-18 at 19 28 01

### DIFF
--- a/src/components/AddCourseButton.vue
+++ b/src/components/AddCourseButton.vue
@@ -53,7 +53,7 @@ export default Vue.extend({
     display: flex;
     justify-content: center;
     align-items: center;
-    border: 2px dashed #d8d8d8;
+    border: 2px dashed $borderGray;
     color: $medGray;
     margin-left: 1.125rem;
     margin-right: 1.125rem;
--- a/src/components/AddCourseButton.vue
+++ b/src/components/AddCourseButton.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    class="semester-courseWrapper semester-addWrapper"
+    class="semester-addWrapper"
     :class="{
       'semester-addWrapper--compact': compact,
       'my-2 mx-0': shouldClearPadding,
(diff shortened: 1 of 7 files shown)


src/components/AddCourseButton.vue
```diff
@@ -45,32 +45,21 @@ export default Vue.extend({
...
   &-addWrapper {
     margin-top: -5rem;
-    width: 21.375rem;
+    margin-bottom: 0.5rem;
     height: 4.625rem;
     border-radius: 0.5rem;
     display: flex;
     justify-content: center;
     align-items: center;
+    border: 2px dashed #d8d8d8;
```
tcho6319, line 56: There should be an scss variable for `#d8d8d8`

tcho6319 (changes requested):
Thanks Sam! Left a nit and noted the following:

1. Since the course component widths are dependent on the width of the parent containers, I think we get cases where the course components are a little longer than what the designs call for. When I check the staging site on my mobile device, it looks fine but at some medium screen width sizes (smaller than desktop but larger than mobile), we get some wonkier looking course cards. Left is this PR and right is master. Do you think there is an elegant solution to this? Maybe setting the width of the container to be smaller or increasing padding within the container?
Screen Shot 2021-03-19 at 3 13 05 PM

2. Similarly for the See All courses:
Screen Shot 2021-03-19 at 3 13 31 PM

tcho6319 (approved):
LGTM! Ty Sam 😎

PR #368, opened 2021-03-19
Unbreak total academic credits toggle

SamChou19815:
### Summary 

Add an empty array so that the code that tries to iterate over the first slot courses array doesn't crash.

### Test Plan 

Screen Shot 2021-03-18 at 20 39 37


einc (approved):
LGTM! Thank you Sam!

PR #370, opened 2021-03-20
Remove all edges when user has no requirement choice on a course

SamChou19815:
### Summary 

This PR fixes the problem when user doesn't select any requirement for a course to fulfill, but it causes the course to fulfill all matching requirements and results in double counting warning.

The issue is in step 3 of the requirement graph building:

> Remove all the edges from courses to requirements that are not the picked requirement in the choice.

This step doesn't remove edges at all when the user doesn't make any choice, which causes the problem described above.

The fix itself is very straightforward: take a look at **every course**. If the course has no choice, when the chosen requirement is `null`. Since `null` is not equal to any connected requirement that doesn't allow double counting, all those edges will be correctly removed.

This changes cause some test that abuse the wrong old behavior to break, so I also fixed those tests.

**While fixing the requirement, I realized that this change also makes double counting impossible**, so I removed all the code to support double counting warnings as well. It's because for each course, it can only connect to one courses that doesn't allow double counting. Previously it's possible because if you don't select any requirement, it's matched to all courses. This PR removes the ability to do that.

### Test Plan 

Add CS4300 without assigning any requirement to it. Now it doesn't auto fulfill CS electives any more.

### Breaking Changes  

This is actually a big breaking change for users on prod. They don't have any requirement choice data, so they won't fulfill any requirement that doesn't allow double counting. We should probably add another type of warning when a course is not used to fulfill any requirement.

I implemented the new warning in this PR, so they existing users on prod can know why the requirement bar stops to show the added courses fulfill the requirements.

Screen Shot 2021-03-21 at 21 04 23

### DIFF
--- a/src/requirements/__test__/requirement-graph-builder.test.ts
+++ b/src/requirements/__test__/requirement-graph-builder.test.ts
@@ -22,14 +22,17 @@ const getAllCoursesThatCanPotentiallySatisfyRequirement = (
 };
 
 it('buildRequirementFulfillmentGraph phase 1 test 1', () => {
-  const { requirementFulfillmentGraph: graph } = buildRequirementFulfillmentGraph({
-    requirements,
-    userCourses: [CS3410, CS3420, MATH4710],
-    userChoiceOnFulfillmentStrategy: {},
-    userChoiceOnDoubleCountingElimination: [],
-    getAllCoursesThatCanPotentiallySatisfyRequirement,
-    allowDoubleCounting: () => false,
-  });
+  const graph = buildRequirementFulfillmentGraph(
+    {
+      requirements,
+      userCourses: [CS3410, CS3420, MATH4710],
+      userChoiceOnFulfillmentStrategy: {},
+      userChoiceOnDoubleCountingElimination: [],
+      getAllCoursesThatCanPotentiallySatisfyRequirement,
+      allowDoubleCounting: () => false,
+    },
+    /* keepCoursesWithoutDoubleCountingEliminationChoice */ true
+  );
 
   expect(graph.getConnectedCoursesFromRequirement('CS3410/CS3420')).toEqual([CS3410, CS3420]);
   expect(graph.getConnectedCoursesFromRequirement('Probability')).toEqual([MATH4710]);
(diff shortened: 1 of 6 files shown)


src/requirements/__test__/requirement-graph-builder.test.ts
```diff
@@ -22,14 +22,17 @@ const getAllCoursesThatCanPotentiallySatisfyRequirement = (
...
+    {
+      requirements,
+      userCourses: [CS3410, CS3420, MATH4710],
+      userChoiceOnFulfillmentStrategy: {},
+      userChoiceOnDoubleCountingElimination: [],
+      getAllCoursesThatCanPotentiallySatisfyRequirement,
+      allowDoubleCounting: () => false,
+    },
+    /* keepCoursesWithoutDoubleCountingEliminationChoice */ true
+  );
```
tcho6319, line 35: For my own understanding: we have the testing flag `keepCoursesWithoutDoubleCountingEliminationChoice` as `true` because it we didn't have this flag at all, we would accidentally not match the possible courses to the requirement in phase 1? For example, the edges between the courses `CS3410` and `CS3420` and `'CS3410/CS3420'` would not exist without this testing flag because the user choice for these courses would be null?

src/requirements/__test__/requirement-graph-builder.test.ts
```diff
@@ -88,81 +100,73 @@ it('buildRequirementFulfillmentGraph phase 2-2 test', () => {
...
 
-// The following 3 tests test that we will detect all double counted courses.
-it('buildRequirementFulfillmentGraph phase 4 test 1', () => {
-  const { illegallyDoubleCountedCourses } = buildRequirementFulfillmentGraph({
+// The following test ensures that we will remove edges when user makes no choice on courses.
+it('buildRequirementFulfillmentGraph phase 3 test 3', () => {
+  const graph = buildRequirementFulfillmentGraph({
     requirements,
     userCourses: [CS3410, CS3420, MATH4710],
     userChoiceOnFulfillmentStrategy: { 'CS3410/CS3420': [CS3410.courseId] },
```
tcho6319, line 143: Does this mean that the course `CS3410` is applied to the `'CS3410/CS3420'` requirement or just that they're going to fulfill the requirement using CS 3410 but haven't applied to course to it yet?

tcho6319 (commented):
Thanks Sam! Left a couple clarification questions. I also wanted to note some issues:

1. I'm not sure if this issue is resolved by #371 but when you drag a course from the req bar and into the schedule, the course only applies to the requirement from which I dragged if it happens to be one that the course automatically fulfills. For example, since CS 1110 automatically fulfills Introductory Programming, when I drag the course from that req, it applies correctly. However, since CS 2800 does not automatically fulfill any requirement, when I drag the course from the Computer Science Core req, CS 2800 does not apply to it and the progress counter does not increase.

https://user-images.githubusercontent.com/54298311/111906717-63a4c600-8a28-11eb-8f23-00b8da7b54bf.mov

2. Does this fix make it easier to solve the problem I brought up about including CS 2110 for the Interactive Tech INFO concentration?
> I would add CS 2110 as its own slot because it's required for this concentration and then update the description
> 
> Screen Shot 2021-03-15 at 9 50 25 AM
https://github.com/cornell-dti/course-plan/pull/357#discussion_r594351849

SamChou19815:
> Thanks Sam! Left a couple clarification questions. I also wanted to note some issues:
> 
> 
> 
> 1. I'm not sure if this issue is resolved by #371 but when you drag a course from the req bar and into the schedule, the course only applies to the requirement from which I dragged if it happens to be one that the course automatically fulfills. For example, since CS 1110 automatically fulfills Introductory Programming, when I drag the course from that req, it applies correctly. However, since CS 2800 does not automatically fulfill any requirement, when I drag the course from the Computer Science Core req, CS 2800 does not apply to it and the progress counter does not increase.
> 
> 
> 
> https://user-images.githubusercontent.com/54298311/111906717-63a4c600-8a28-11eb-8f23-00b8da7b54bf.mov
> 
> 
> 
> 2. Does this fix make it easier to solve the problem I brought up about including CS 2110 for the Interactive Tech INFO concentration?
> 
> > I would add CS 2110 as its own slot because it's required for this concentration and then update the description
> 
> > 
> 
> > Screen Shot 2021-03-15 at 9 50 25 AM
> 
> https://github.com/cornell-dti/course-plan/pull/357#discussion_r594351849
> 
> 
> 
> 

1. Yes it will be resolved by that PR
2. I don't think so. We still need to annotate each toggleable choices as allowDoubleCounting which has nothing to do with this PR

tcho6319 (approved):
LGTM! @einc does the warning look ok?

PR #372, opened 2021-03-21
Make mobile semester view occupy the entire width minus margin

SamChou19815:
### Summary 

This PR makes the mobile view looks slightly better by making the semester card occupying the entire width, minus 2rem margin on left and right.

### Test Plan 

Screen Shot 2021-03-21 at 20 59 43
Screen Shot 2021-03-21 at 20 59 47

### Notes 

I originally plan to implement the full mobile design as shown here:

Screen Shot 2021-03-21 at 15 07 43

Then I realized that it doesn't work with the drag and drop library we are using, which expects a one-column linear flow. I wonder whether I should do one of the alternative design instead?

Screen Shot 2021-03-21 at 15 08 52

### DIFF
--- a/src/components/Semester/Semester.vue
+++ b/src/components/Semester/Semester.vue
@@ -560,4 +560,12 @@ export default Vue.extend({
     }
   }
 }
+
+@media only screen and (max-width: $small-breakpoint) {
+  .semester,
+  .semester--compact {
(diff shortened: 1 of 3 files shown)


tcho6319 (approved):
LGTM! Ty Sam

PR #373, opened 2021-03-21
Self Check Frontend Finish

willespencer:
### Summary 

This pull request puts the finishing touches on the Frontend for the Self Check add modal and the self check courses on the requirements sidebar. ([Notion](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=0c4db7e8431845c7bbda1343693dc232))

- [x] Filter out courses from self-check dropdowns if they are assigned to another requirement (if that req and this req are both not double countable)
- [x] Add a placeholder on the self check add modal for season/year instead of defaulting to the current season/year
- [x] Styling updates to match Figma (add req name in add modal title, increase margin around dropdowns, style toggleable dropdown the same as self check dropdown)
- [x] Resolve bugs (Hide dropdown if self check is fulfilled, show "Add Course #" for course self checks, replace unchanged isFulfilled with isCompleted which brings back some old reqs styling)

image
image

Remaining TODOs:

- [ ] Resolve the bigger issue of courses assigned to self checks even if they fail the checkers (next Sprint)
- [ ] Resolve the two bugs on notion found in edit semester that now apply to the self check modal (next Sprint)

### Test Plan 

1. Ensure that a course assigned to a non-double countable requirement does not show up in any non-double countable self check dropdowns
2. Ensure that a course assigned to a double countable req shows up in other self-check dropdowns. A great test for this is dragging a course from the probability req for CS to your plan, and then opening a self-check dropdown.
3. Ensure a course assigned to a non-double countable req still shows up in a double-countable self check. **NOTE:** I have not tested this, do we have any examples where this is the case?
4. Ensure self-check dropdowns cannot be added to for self-checks that are fulfilled. **NOTE:** I checked this with courses, but I don't think we have a credit self-check right now? Is there a way to test this?
5. Ensure the self-check add modal matches designs, that the default semester/year is "Select," and that adding it adds it to the correct semester
6. Ensure the req bar better matches styles now, especially for self-check and toggleable dropdowns, in addition to margins.
7. Ensure that new semester and edit semester work like they do on the web app (since some common code was changed).

### Notes 

I was unsure of how best to handle not showing courses in the self-check dropdowns depending on double counting cases, so the code is heavily commented and might not be that clean. Please let me know if I should remove any of my comments or handle this in a better way!

Also just found a bug when making this PR that I will investigate, where this ECON course fulfills Probability and Technical but only shows up for one.
image


src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -60,19 +66,64 @@ export default Vue.extend({
...
-            courses[course.code] = course;
+
+          // if course is mapped to another req(s), only allow it if all other reqs are double countable
+          let isAddable = true;
+          const otherReqsMappedTo = store.state.requirementFulfillmentGraph.getConnectedRequirementsFromCourse(
+            { uniqueId: course.uniqueID }
+          );
+
+          let allOtherReqsDoubleCountableIfAny = true;
+          let thisReqDoubleCountable = false;
```
tcho6319, line 86: Can we add some comments that explain what these two variables mean?

tcho6319 (changes requested):
Thanks Will!! Self Check is coming together :D. I noted some issues:

1. I don't think the resetting functionality of the dropdowns when editing semester in the new self check add modal works properly. This may be what you mentioned earlier:
> Resolve the two bugs on notion found in edit semester that now apply to the self check modal (next Sprint)

https://user-images.githubusercontent.com/54298311/112203260-baf19480-8be8-11eb-9d80-bac42eb379c8.mov

If you haven't already, can you make a Notion bug item for this or try to resolve in this PR if it's an easier fix? Ty.

2. I feel like the dropdown can get very long. Can we decrease the size of the dropdown div to show like 4 at a time?
Screen Shot 2021-03-23 at 3 06 09 PM

3. Can you add this bug to Notion and assign yourself as champion?:
> Also just found a bug when making this PR that I will investigate, where this ECON course fulfills Probability and Technical but only shows up for one.

4. Make sure to merge in latest master.
> Ensure that a course assigned to a double countable req shows up in other self-check dropdowns. A great test for this is dragging a course from the probability req for CS to your plan, and then opening a self-check dropdown.

I tried to follow above ^ with CS 1110, and it seems to get automatically assigned to the Computing req (in addition to autofulfilling Introductory Programming) so it's hard to test this right now. This problem is resolved on master.

willespencer:
> 1. I don't think the resetting functionality of the dropdowns when editing semester in the new self check add modal works properly. This may be what you mentioned earlier:
> 
> > Resolve the two bugs on notion found in edit semester that now apply to the self check modal (next Sprint)
> 
> If you haven't already, can you make a Notion bug item for this or try to resolve in this PR if it's an easier fix? Ty.

Yep this is already on [notion](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=97b029f66e7a4a38a076813889965a84) for the bug fixes week! Will try to work on it asap in a separate PR, but the only way I was able to fix it was by breaking the edit semesters dropdown, so need to spend some more time with it.

> 
> 2. I feel like the dropdown can get very long. Can we decrease the size of the dropdown div to show like 4 at a time?

Done! Please confirm it looks good now 😄 

> 3. Can you add this bug to Notion and assign yourself as champion?:
> 
> > Also just found a bug when making this PR that I will investigate, where this ECON course fulfills Probability and Technical but only shows up for one.
> 

Yeah, I just did [here](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=f82e9f3d222145bab804253194e8a74c). However, I can't reproduce it anymore - by dragging, adding through the self check add modal, or the main add modal. Could you try and see if you can? Otherwise it might have been fixed in someone else's PR.

> 4. Make sure to merge in latest master.

Done! Let me know if the test works now

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -104,10 +158,25 @@ export default Vue.extend({
...
+  height: 1px;
+  width: 100%;
+  background-color: $inactiveGray;
+  margin-top: 1rem;
+}
+
+.top {
+  font-size: 14px;
+  line-height: 17px;
+  color: #757575;
```
tcho6319, line 171: Nit: I think we have an scss variable

tcho6319 (approved):
LGTM! Thanks Will :)

src/components/Requirements/IncompleteSelfCheck.vue
willespencer: You're right we do my bad, but it looks like this hex code is used in a few different places. Making a notion item for myself to fix this soon!

PR #374, opened 2021-03-21
[easy] Do not show incorrect semester warning when semesters are empty

SamChou19815:
### Summary 

Fix https://www.notion.so/courseplan/Warnings-Missing-Semester-70496a5911ee4ee096a4abd348a77c0a

We should simply not show it, since missing `semesters` array means that we don't know when it will be offered.

### Test Plan 

No longer show up.

Screen Shot 2021-03-21 at 19 54 03


einc (approved):
Tested and looks good!

PR #375, opened 2021-03-22
Correctly apply AP/IB transfer credit to requirements

benjamin-shen:
### Summary 

This pull request fixes [this notion bug](https://www.notion.so/courseplan/AP-credit-for-Total-Academic-Credits-f2fbf395640044bdb675ed6397f48191).

- [x] updated `courseIsAllEligible` to allow ap/ib transfer credit
- [x] added option to disallow applying transfer credits to requirement

### Test Plan 

AP CS now applies to total academic credits but not A&S credits.
image

### Notes 

Note that there is now a flag in `requirement-types.d.ts` under `RequirementCommon` type: `disallowTransferCredit`. This is defaulted to undefined/falsy. In the future, if a requirement needs to be prevented from having transfer credits apply to it, just include the flag and set it to true. For example, in this PR, I added it to `asPreFA2020.ts` under the `A&S Credits` requirement.


src/requirement-types.d.ts
```diff
@@ -7,6 +7,9 @@ type RequirementCommon = {
   readonly source: string;
   /** If this is set to true, then an edge to the course doesn't count towards double counting. */
   readonly allowCourseDoubleCounting?: true;
+  /** If this is set to true, then AP/IB credits cannot be applied towards this requirement. */
+  readonly disallowTransferCredit?: true;
```
SamChou19815, line 11: Not sure whether this is the right move. It increases the complexity of the requirement algorithm by introducing one more flag to worry about. While the same can be achieved by making requirements opt-in for allowing AP/IB credits by explicitly checking for AP/IB credit only course id.

src/requirements/data/checkers-common.ts
```diff
@@ -39,7 +38,6 @@ export const courseMatchesCodeOptions = (course: Course, codeOptions: readonly s
  * @returns if the course satisfies FWS requirement.
  */
 export const courseIsFWS = (course: Course): boolean =>
-  course.crseId === FWS_COURSE_ID ||
```
benjamin-shen, line 42: I removed this because during the checker/precomputation phase, a `crseId` will never be our custom `FWS_COURSE_ID`. Adding courses with our custom course ids (type `CourseTaken`, opposed to `Course`) happens after the AP/IB exam conversion, during the frontend computation.

src/requirements/requirement-frontend-computation.ts
```diff
@@ -12,22 +13,13 @@ type FulfillmentStatistics = {
...
 } & RequirementFulfillmentStatistics;
 
-/**
- * @param course course object with useful information retrived from Cornell courses API.
- * @returns true if the course is AP/IB equivalent course or credit
- */
-const courseIsAPIB = (course: CourseTaken): boolean =>
-  [CREDITS_COURSE_ID, FWS_COURSE_ID].includes(course.courseId) ||
-  ['AP', 'IB'].includes(course.code.split(' ')[0]);
-
```
benjamin-shen, line 22: I moved this function to `requirement-frontend-utils.ts` because it is also used there.

src/requirements/requirement-frontend-computation.ts
```diff
@@ -12,22 +13,13 @@ type FulfillmentStatistics = {
...
 /**
  * Used for total academic credit requirements for all colleges except EN and AR
  * @param course course object with useful information retrived from Cornell courses API.
  * @returns true if the course is not PE or 10** level
  */
 const courseIsAllEligible = (course: CourseTaken): boolean => {
-  if (course.courseId === CREDITS_COURSE_ID) return true;
-  if (courseIsAPIB(course)) return false;
+  if (courseIsAPIB(course)) return true;
```
benjamin-shen, line 22: This `courseIsAPIB(course)` check is still necessary to ensure that AP/IB transfer credits are always counted (even if, for some reason, they conflict with `subject !== 'PE' && !number.startsWith('10')`).

The previous line is no longer necessary, because it's included in the `courseIsAPIB` function.

src/requirements/requirement-frontend-utils.ts
```diff
@@ -168,20 +177,22 @@ export function computeFulfillmentCoursesAndStatistics(
...
     const { courseId } = courseTaken;
-    for (
-      let subRequirementIndex = 0;
-      subRequirementIndex < eligibleCourses.length;
-      subRequirementIndex += 1
-    ) {
-      if (
-        eligibleCourses[subRequirementIndex].includes(courseId) &&
-        subRequirementProgress[subRequirementIndex] < perSlotMinCount[subRequirementIndex]
+    if (!(requirement.disallowTransferCredit && courseIsAPIB(courseTaken))) {
```
benjamin-shen, line 180: I'm not sure why the formatting on Github is weird, but the only thing I added was this line. It ensures that a course can only fulfill the requirement if the requirement doesn't have the `disallowTransferCredit` flag (or, if it does, that the course isn't an AP/IB equivalent course).

SamChou19815 (approved):

PR #376, opened 2021-03-22
Match to figma and add ap/ib reset feature

neketka:
### Summary 

Fixed the reset confirmation modal design to match the figma design.

image
image

Added the ability to remove transfer credits as well.

- [x] Implemented firestore commit to the onboarding data in order to remove the transfer course 
- [x] Changed css styling to match figma style.

Remaining TODOs:

- [ ] Make the firestore commit more flexible. Onboarding data has a really finicky way of committing to firestore because AppOnboardingData does not match FirestoreOnboardingData.

### Test Plan 

Check that the style is correct and transfer credits are always removed.

### DIFF
--- a/src/components/Requirements/CompletedSubReqCourse.vue
+++ b/src/components/Requirements/CompletedSubReqCourse.vue
@@ -42,7 +42,7 @@ import ReqCourse from '@/components/Requirements/ReqCourse.vue';
 import ResetConfirmationModal from '@/components/Modals/ResetConfirmationModal.vue';
 import store from '@/store';
 import { deleteCourseFromSemesters } from '@/global-firestore-data';
-import { db, onboardingDataCollection } from '@/firebaseConfig';
+import { onboardingDataCollection } from '@/firebaseConfig';
 import getCurrentSeason, { getCurrentYear } from '@/utilities';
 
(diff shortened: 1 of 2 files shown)


src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -87,7 +88,25 @@ export default Vue.extend({
...
+          db.batch()
+            .set(onboardingDataCollection.doc(store.state.currentFirebaseUser.email), {
+              tookSwim: onBoardingData.tookSwim,
+              colleges: [{ acronym: onBoardingData.college }],
+              class: onBoardingData.transferCourse,
+              majors: onBoardingData.major.map(m => ({ acronym: m })),
+              minors: onBoardingData.minor.map(m => ({ acronym: m })),
+              exam: onBoardingData.exam.filter(e => !(e.type === type && e.subject === name)),
+            } as FirestoreOnboardingUserData)
+            .commit();
```
SamChou19815, line 107: You don't really need batch here, since you are only updating one document here. You can also use `.update` so that you don't have to copy over the unchanged fields like `majors, minors`

SamChou19815 (approved):

PR #377, opened 2021-03-22
Created landing page

neketka:
### Summary 

Updated landing page according to the new figma

- [x] Added asset files
- [x] Changed Login.vue and Footer.vue

### Test Plan 

Test if the landing page is responsive enough and if the page looks as was presented


src/assets/images/landing_new/icon1.svg
```diff
@@ -0,0 +1,11 @@
+<svg width="57" height="71" viewBox="0 0 57 71" fill="none" xmlns="http://www.w3.org/2000/svg">
```
willespencer, line 1: Since you replaced a bunch of the landing page images, can you delete the outdated ones from the repo that are not used anymore?

src/components/Footer.vue
```diff
@@ -2,14 +2,16 @@
...
+        >
       </div>
       <div class="col-3">
-        <p class="footer">Built with LOVE</p>
+        <p class="footer">Built with ❤️</p>
       </div>
       <div class="col-3 footer">
         <a
-          class="footer"
+          class="footer underline"
```
willespencer, line 14: Looks like adding this class makes the underline on the footer links permanent instead of showing up on hover. Do we not want this to be a hover state, and if not, do we want any other sort of hover state? Tagging @einc on this because I'm not sure what the answer is

src/containers/Login.vue
```diff
@@ -6,20 +6,20 @@
...
-
+    <div class="landing top-sec">
       <!--PLAN AHEAD-->
       <div class="container p-0 m-0">
         <div class="row top-bar phonepad no-gutters">
           <div class="col">
             <div class="row no-gutters top-section">
-              <h1 class="plan-head mt-5">Plan Your Courses Ahead</h1>
+              <h1 class="plan-head mt-5">
+                Schedule Smarter <br />
```
willespencer, line 16: If you can, can you implement this without the line break element? It is usually considered better practice to use multiple elements and CSS styles to add space between text, especially in case this text or formatting needs to be changed later

src/containers/Login.vue
```diff
@@ -29,7 +29,7 @@
                   class="email-button email-button--top"
                   variant="primary"
                 >
-                  ACCESS BETA
+                  GET STARTED
```
willespencer, line 32: The font sizes on the landing page seem to be a bit different than the ones on Figma, especially for this button. Could you update them to make sure they match?

src/containers/Login.vue
```diff
@@ -38,75 +38,59 @@
...
-            <div class="row tasks">
-              <div class="col-1 tasks">
-                <img src="@/assets/images/Task4.svg" alt="Starred comment" />
-              </div>
-              <div class="col-11">
-                <p class="sub sub--task">Recommends courses based on your needs</p>
-              </div>
-            </div>
+          <div class="col tasks">
+            <figure class="figure">
```
willespencer, line 52: Huh, I'd never heard of `figure` or `figcaption` elements before. Just out of curiosity, is there a reason why you used these instead of `img` tags?

src/containers/Login.vue
```diff
@@ -115,58 +99,23 @@
...
-            </div>
-          </div>
+          <img
+            class="hide schedule"
+            src="@/assets/images/landing_new/img2.svg"
+            alt="Plan preview"
+          />
         </div>
       </div>
+      <br style="clear: both" />
```
willespencer, line 118: What does this do?

SamChou19815 (approved):

src/containers/Login.vue
```diff
@@ -179,7 +128,6 @@ import Vue from 'vue';
 import firebase from 'firebase/app';
 
 import CustomFooter from '@/components/Footer.vue';
-import TopBar from '@/components/TopBar.vue';
```
willespencer, line 182: Now that the landing page does not use the topbar component with the CoursePlan logo, do we still want it on the 404 page? @tcho6319

src/containers/Login.vue
```diff
@@ -310,6 +258,18 @@ export default Vue.extend({
 <style scoped lang="scss">
 @import '@/assets/scss/_variables.scss';
 
+.top-sec {
+  background-image: url('../assets/images/landing_new/top_rect.svg');
```
willespencer, line 262: Can you use the `~@` notation on this url just in case this file ever gets moved? There are some examples of `background-image: url(~@/...)` elsewhere in the codebase.

src/containers/Login.vue
```diff
@@ -322,10 +282,10 @@ export default Vue.extend({
   max-width: 100%;
 }
 .top-bar {
-  padding: 59px 0px 0px 104px;
+  padding: 29px 0px 0px 104px;
```
willespencer, line 285: I know you didn't create these padding numbers, but could you update them to use rem values instead of pixel values throughout this file? And just as an FYI you might need to round them before converting to get nicer rem numbers (so maybe 28px -> 1.75rem).

Good article on why you should use rem if you're not familiar with it: https://engageinteractive.co.uk/blog/em-vs-rem-vs-px

src/containers/Login.vue
```diff
@@ -345,22 +305,19 @@ export default Vue.extend({
...
 }
-.plan {
-  padding: 59px 0px 59px 104px;
-}
 .plan-head {
   padding-bottom: 30px;
   font-size: 60px;
   line-height: 60px;
-  color: #4f4f4f;
+  color: white;
```
willespencer, line 312: Can you make sure to use the color variables we have, so $white here and other places throughout the file? This would make it easy for us to update the color scheme of the site if we needed to, as we could just change the value of the variable.

src/containers/Login.vue
```diff
@@ -379,7 +336,7 @@ input {
 }
 .email-button {
   border: 0;
-  background-color: $emGreen;
+  background-color: #539d9b;
```
willespencer, line 339: This hex color seems to only be used on this login button right now, which makes it seem like it might be the wrong color on Figma... @einc is this the color we want the landing button to be?

src/containers/Login.vue
```diff
@@ -425,16 +377,14 @@ input {
...
+
+.new-way {
+  margin-top: 350px;
 }
 
 .new {
-  background-color: $emGreen;
-  padding: 50px 0px 96px 104px;
+  padding: 50px 104px 96px 104px;
+  color: black;
```
willespencer, line 387: Please use $black here and other places in the file!

willespencer (changes requested):
The new landing page looks really good! Just left some comments & questions (mostly about the code), and I'm excited for this update to be on the site soon 😄

src/components/Footer.vue
einc: Design wise, let's add hover states to the links and set them lighter.
Copying from Theresa: "By default, CP implements hovering on buttons (not divs) by setting the opacity to 0.95 on hover. See _global.scss."

hahnbeelee:
I checked up on this to make sure things were accessible and overall looks really good!
can you add aria-labels to the images?

src/containers/Login.vue
```diff
@@ -6,20 +6,20 @@
...
-              <h1 class="plan-head mt-5">Plan Your Courses Ahead</h1>
+              <h1 class="plan-head mt-5">
+                Schedule Smarter <br />
+                Plan Ahead
+              </h1>
             </div>
             <div class="row no-gutters top-section">
               <p class="plan-subhead">
-                Introducing the new and easiest way to plan courses at Cornell
+                Introducing the newest and easiest way to plan courses at Cornell
```
jason-tung, line 22: image

It looks like this part is broken up into two lines and the font is a bit different on Figma

src/containers/Login.vue
```diff
@@ -38,75 +38,59 @@
...
+            class="hide preview"
+            src="@/assets/images/landing_new/img1.svg"
+            alt="Dragging preview"
+            style="position: absolute"
+          />
+          <div class="comment comment-drag">
+            <h1 class="head">Drag and Drop Courses</h1>
+            <p class="sub-text">
               CoursePlan’s intuitive interface recommends courses based on unfulfilled requirements
               and allows you to easily drag and drop them into your planner
```
jason-tung, line 95: I think this should be aligned more to the right and also I get that this is as per Figma spec but it looks absolutely massive on my browser

src/containers/Login.vue
jason-tung: It clears the float he used to make "Plan Your Semesters" go to the right

src/containers/Login.vue
```diff
@@ -115,58 +99,23 @@
       </div>
 
       <!-- SEMESTERS -->
-      <div class="container-fluid p-0 small">
-        <div class="row semester phonepad m-0">
-          <div class="col-12 col-md-5 comment">
+      <div class="container-fluid p-0 center">
+        <div class="semester phonepad row no-gutters float-right">
```
jason-tung, line 103: flexbox is generally considered the better way to do something like this. Look up `justify-content: flex-end`.

src/containers/Login.vue
```diff
@@ -448,75 +398,119 @@ input {
   align-items: center;
   justify-content: center;
   padding: 10px;
+  color: black;
```
jason-tung, line 401: https://github.com/cornell-dti/course-plan/pull/377/files#r598816030

src/containers/Login.vue
```diff
@@ -448,75 +398,119 @@ input {
...
+
+figure {
+  text-align: center;
+}
+
+.sub-text {
+  font-weight: normal;
+  font-size: 32px;
+  text-align: left;
+  color: black;
```
jason-tung, line 418: https://github.com/cornell-dti/course-plan/pull/377/files#r598816030

src/containers/Login.vue
```diff
@@ -448,75 +398,119 @@ input {
...
+    text-align: left;
+  }
+}
+
 .sub {
   font-weight: normal;
   font-size: 24px;
-  color: #ffffff;
+  text-align: center;
+  color: black;
```
jason-tung, line 430: https://github.com/cornell-dti/course-plan/pull/377/files#r598816030

src/containers/Login.vue
```diff
@@ -448,75 +398,119 @@ input {
...
+    vertical-align: middle;
+    padding-bottom: 38px;
   }
 }
 .head {
   font-weight: 600;
+  line-height: 38px;
   font-size: 40px;
-  color: #ffffff;
+  color: black;
```
jason-tung, line 442: https://github.com/cornell-dti/course-plan/pull/377/files#r598816030

src/containers/Login.vue
```diff
@@ -448,75 +398,119 @@ input {
...
   padding-bottom: 0px;
 }
 .head-center {
   font-weight: 600;
   font-size: 40px;
-  color: #ffffff;
+  color: black;
```
jason-tung, line 448: https://github.com/cornell-dti/course-plan/pull/377/files#r598816030

jason-tung (changes requested):
Site looks nice!

src/containers/Login.vue
jason-tung: > Please use $black here and other places in the file!

src/containers/Login.vue
neketka: Can you post a screenshot?

tcho6319 (changes requested):
Thanks Nikita! Landing page looks awesome :D I agree w/ the comments/changes that the other devs made. I noted some additional changes:

1. Nit: Can we decrease the empty space between the Plan Your Semesters div and the left a little?
PR:
Screen Shot 2021-03-23 at 5 29 47 PM

Designs:
Screen Shot 2021-03-23 at 5 30 44 PM

2. At the point when the images disappear, can we have the text be aligned center like so?
Screen Shot 2021-03-23 at 9 07 19 PM

3. At smaller breakpoints, we get the following alignment of the icons:
Screen Shot 2021-03-23 at 9 10 14 PM

Instead can we make them vertical for the smaller breakpoints before we get the alignment as ^ (with `flex-direction: column`):
Screen Shot 2021-03-23 at 9 11 04 PM

src/containers/Login.vue
jason-tung: image

cursor in bottom right for comparison

src/containers/Login.vue
neketka: The attach captions to images more easily.

willespencer (approved):
Looks good! All my issues were addressed. Excited for this to go live soon 😀

tcho6319 (approved):
LGTM! Thanks for addressing my requested changes :)

jason-tung (approved):

PR #381, opened 2021-03-23
5/n - Divs to buttons (keyboard accessibility)

hahnbeelee:
### Summary 
Changes all divs with a @click event to a button. This makes things much more keyboard accessible!

### Test Plan 
- make sure frontend hasn't changed
- press tab and make sure everything that should be clickable is highlighted

### Notes
This pretty much makes 99% of the site keyboard accessible. 
Other things left for keyboard accessibility:
- dropdown component
- modal refactor (if you open a model and use tab things behind the modal get highlighted unless you click on the modal)
- Couldn't get the "+ New Course" button to look good when i changed it from a div to a button for the life of me. If anybody confident in their CSS want to take a stab please do, but I just wanted to get the majority of this PR in.

To-Do
- [x] change divs to buttons
- [x] more keyboard accessibility (should happen gradually)
- [ ] accessible dropdown component
- [ ] modal refactors
- [ ] global button component
- [ ] onboarding needs to have its own set of headings
- [x] aria tags

@SamChou19815 what are your thoughts on an eslint rule where if an element has a @click tag it has to be a button type?


SamChou19815 (approved):
Everything seems to work.

> @SamChou19815 what are your thoughts on an eslint rule where if an element has a @click tag it has to be a button type?

That looks reasonable to me. Have you tried the eslint a11y plugin for vue?

src/components/Semester/SemesterView.vue
```diff
@@ -26,18 +26,19 @@
         data-disable-interaction="1"
         data-step="4"
         data-tooltipClass="tooltipCenter"
+        v-if="!isMobile"
```
tcho6319, line 29: I believe we brought back the ability to toggle between compact and not compact view in mobile so you can remove this line

tcho6319 (changes requested):
Thank you so much Hahnbee! This looks great 🥰. I left a comment about a regression but once that's addressed, this PR should be g2g!

tcho6319 (approved):
LGTM! Thanks Hahnbee :)

PR #384, opened 2021-03-24
New Majors & Minors (ECE, ISST, DEA) + minor fixes

einc:
### Summary 

Title 

- Added TODO comments for requirements we can't fully support and GitHub comments for any changes that might need explanation

### Test Plan 

- ISST now has new requirements based on incoming year. I split them up as a temp fix and tested that both show.
- I'm going to ask designers to thoroughly check the ISST majors and DEA minors (those are their majors/minors) and update based on their feedback so just do quick check and test that nothing breaks as a result of these new majors/minors.


src/requirements/data/colleges/en.ts
```diff
@@ -192,7 +193,6 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
     ),
     fulfilledBy: 'courses',
     perSlotMinCount: [1],
-    allowCourseDoubleCounting: true,
```
einc, line 195: some EN majors like ISST do not allow this requirement to be double counted. so we should add allowCourseDoubleCounting: true to the specific major requirements

src/requirements/data/majors/oldIsst.ts
```diff
@@ -0,0 +1,391 @@
...
+    fulfilledBy: 'toggleable',
+    fulfillmentOptions: {
+      'Information Science': {
+        description:
+          'A. Three courses from Information Systems ' +
+          'B. One course from Mathematical Modeling in IT ' +
+          'C. Three electives, all from one of Human-Centered Systems or Social Systems ' +
+          'D. Two major-approved electives chosen from areas I-VI. INFO 4900 may be used to fulfill one major-approved elective ',
+        counting: 'courses',
+        checker: includesWithSubRequirements(
```
einc, line 72: @SamChou19815 Is there a better way to do checkers for toggleable requirements with multiple requirements? Is using includesWithSubRequirements the only option currently? 
[A lot of these requirements are repeated so I can make them into arrays to make it easier to check but wasn't sure if I could do that]

src/requirements/data/minors/cs.ts
```diff
@@ -25,12 +30,34 @@ const csMinorRequirements: readonly CollegeOrMajorRequirement[] = [
...
+    checker: [
+      (course: Course): boolean => {
+        if (
+          courseMatchesCodeOptions(course, [
+            'CS 4090',
+            'CS 4997',
+            'CS 4998',
+            'CS 4999',
+            'CS 7***',
+            'CS 3110',
```
einc, line 47: temporarily excluding courses that fulfill other requirements in the same minor to be extra safe as we don't have ways to disallow double counting for minors

src/requirements/data/colleges/en.ts
```diff
@@ -152,6 +152,7 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
     source:
       'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements',
     fulfilledBy: 'self-check',
+    allowCourseDoubleCounting: true,
```
SamChou19815, line 155: I don't think this can be double counted.

src/requirements/data/colleges/en.ts
SamChou19815: I think if you do that you will need to copy this to every majors? For majors that in multiple colleges like CS, this will then be a problem.

einc:
> [diff-counting] Significant lines: 1145. This diff might be too big! Developer leads are invited to review the code.

i'll try to keep these requirements data updates smaller next time. just wanted to put all of the changes for majors/minors needed for interviews in one place

src/requirements/data/minors/dea.ts
```diff
@@ -0,0 +1,60 @@
+import { CollegeOrMajorRequirement } from '../../types';
+import { includesWithSingleRequirement, includesWithSubRequirements } from '../checkers-common';
+
+const deaMinorRequirements: readonly CollegeOrMajorRequirement[] = [
+  {
+    name: 'DEA Core Courses',
+    description: 'DEA 1101, DEA 1110/DEA 1200, and DEA1500',
+    source: 'https://www.human.cornell.edu/dea/academics/undergraduate/minors/dea',
+    checker: includesWithSubRequirements(['DEA 1050'], ['DEA 1101', 'DEA 1200'], ['DEA1500']),
```
SamChou19815, line 9: `DEA 1500` instead of `DEA1500`?

src/requirements/data/majors/bio.ts
```diff
@@ -113,7 +112,6 @@ const bioRequirements: readonly CollegeOrMajorRequirement[] = [
     fulfilledBy: 'courses',
     perSlotMinCount: [1, 1],
   },
-  // TODO: temp fix for reqs w options and diff minCount
```
SamChou19815, line 116: can you elaborate

src/requirements/data/majors/ece.ts
```diff
@@ -0,0 +1,170 @@
...
+    name: 'Upper-Level ECE Electives: 4000 level',
+    description:
+      'These courses must be technical ECE courses ' +
+      'Three (3) courses at the 4000-level or above. ',
+    source:
+      'https://www.ece.cornell.edu/ece/programs/undergraduate-programs/majors/program-requirements',
+    checker: [
+      (course: Course): boolean => {
+        if (courseMatchesCodeOptions(course, ['CS 4120', 'CS 4410'])) {
+          return false;
```
SamChou19815, line 119: I think this should be true since

> CS 4120 and/or CS 4410 may be used as 4000-level ECE technical electives.

src/requirements/data/majors/ece.ts
```diff
@@ -0,0 +1,170 @@
...
+  {
+    name: 'Upper-Level ECE Electives',
+    description: 'These courses must be technical ECE courses ',
+    source:
+      'https://www.ece.cornell.edu/ece/programs/undergraduate-programs/majors/program-requirements',
+    checker: [
+      (course: Course): boolean => {
+        if (courseMatchesCodeOptions(course, ['CS 4120', 'CS 4410'])) {
+          return false;
+        }
```
SamChou19815, line 99: I don't think we need this here.

SamChou19815 (changes requested):
I think the CS minor and DEA major change look good. ECE also looks good except a small problem. I would recommend splitting these out into a separate PR to make this one smaller.

src/requirements/data/majors/bio.ts
einc: oh this was an earlier comment I had but perSlotMinCount fixed this problem!

src/requirements/data/colleges/he.ts
```diff
@@ -1,31 +1,39 @@
...
-      'a technical writing course must be taken as an engineering distribution, liberal studies, Advisor-approved electives, or Major course.',
-    source:
-      'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements',
-    fulfilledBy: 'self-check',
-  },
   {
     name: '9 Credits In HE Outside Major',
     description:
-      'Students must earn 9 credits in Human Ecology departments outside their major department with rules:',
+      'Students must earn 9 credits in Human Ecology departments outside their major department with rules',
```
tcho6319, line 23: Nit: Period at the end

src/requirements/data/index.ts
```diff
@@ -160,8 +168,13 @@ const json: RequirementsJson = {
...
       requirements: infoRequirements,
     },
+    ISST1: {
+      name: 'Information Science, Systems, and Technology [before Fall 2020]',
+      schools: ['EN'],
+      requirements: oldIsstRequirements,
+    },
     ISST: {
-      name: 'Information Science, Systems, and Technology',
+      name: 'Information Science, Systems, and Technology [Fall 2020 and after]',
```
tcho6319, line 177: Nit: Could we do `ISST1` for the pre Fall 2020 ISST major and `ISST2` for the Fall 2020 and after ISST major? Just to be consistent with how we're naming `AS1` and `AS2`

tcho6319 (commented):
Thank you so much Ein! I left a few nits but once Sam's requested changes are addressed, this PR should be g2g.

src/requirements/data/majors/oldIsst.ts
SamChou19815: right now that is the only way.

src/requirements/data/colleges/en.ts
einc: I added it back. This causes a problem for ISST majors as INFO 1200 double counts for Eng Comm when it can't be but I don't think there's a different way to do it currently wo causing problems for other majors. Added it as a TODO item

src/requirements/data/colleges/en.ts
einc: removed!

tcho6319 (approved):
LGTM! Ty Ein :) Next time, I would try to keep the PR sizes smaller.

SamChou19815 (approved):

PR #387, opened 2021-03-25
added built by cornell dti

jason-tung:
### Summary 

Added `Built with ❤️ by Cornell Design & Tech Initiative` to bottom right of dashboard.
image

https://www.notion.so/courseplan/Add-DTI-name-349db7a11349432a989b9f515554c3e6

### Test Plan 

Look on staging to see if there is any breaking changes this causes.


SamChou19815 (changes requested):
Not sure whether introducing a scrollbar while there is no need is what we want.

jason-tung:
> Not sure whether introducing a scrollbar while there is no need is what we want.

Yeah I just noticed this too I'll fix this

src/containers/Dashboard.vue
```diff
@@ -54,6 +54,12 @@
       @hide="showTourEndWindow = false"
       v-if="showTourEndWindow"
     />
+    <div class="dashboard-builtBy">
+      Built with ❤️ by
```
tcho6319, line 58: @einc can you drop a heart svg for Jason? We should use the svg version of the emoji instead of ascii

src/containers/Dashboard.vue
```diff
@@ -54,6 +54,12 @@
...
+      <a href="https://www.cornelldti.org/projects/courseplan/">
```
tcho6319, line 59: For the link can you add target="_blank" so that it opens up in a new tab?

src/containers/Dashboard.vue
```diff
@@ -246,6 +252,17 @@ export default Vue.extend({
...
   }
+
+  &-builtBy {
+    text-align: right;
+    font-size: 15px;
+    padding: 8px 10px;
+
+    a {
+      color: $medGray;
+      text-decoration: underline $medGray;
```
tcho6319, line 263: I believe we want underline on hover. @einc can you confirm?

tcho6319: Also I'm not sure if the color for "Built with .. by..." is exactly right. I can't seem to find the Figma file for this... @einc can you drop the Figma link?

tcho6319 (changes requested):
Thanks Jason 😎  Left some suggestions/requested changes. Hopefully @einc can provide some answers for the styling and my additional questions below:

1. Do we want the "Built by Cornell..." to be like a fixed footer? @einc 
2. This is how the "Built by Cornell...." functions currently w/ the Bottom Bar. Is this what we intend according to designs? @einc 

https://user-images.githubusercontent.com/54298311/112671736-eff42600-8e38-11eb-8daf-4d40456c34b2.mov

src/containers/Dashboard.vue
tcho6319: Also I'm not sure if the color for "Built with .. by..." is exactly right. I can't seem to find the Figma file for this... @einc can you drop the Figma link?

einc: interaction with bottom bar seems fine to me! and no to fixed footer

src/components/Semester/SemesterView.vue
```diff
@@ -9,86 +9,101 @@
...
+          <div
+            class="semesterView-switchImage semesterView-twoColumn"
+            @click="setNotCompact"
+            :class="{ 'semesterView-twoColumn--active': !compact }"
+          ></div>
+          <div
+            class="semesterView-switchImage semesterView-fourColumn"
+            @click="setCompact"
+            :class="{ 'semesterView-fourColumn--active': compact }"
+          ></div>
```
tcho6319, line 45: This is a regression. These should be buttons.

tcho6319 (changes requested):
Almost there! I noted a regression. Once that's fixed, this PR should be g2g

tcho6319 (approved):
LGTM!

SamChou19815 (approved):

PR #388, opened 2021-03-26
Do not show ineligible requirement in add modal

SamChou19815:
### Summary 

Remove the special case for requirement with checker warnings. In this way, we can natually filter away requirements that cannot be fulfilled by the course.

### Test Plan 

Technical elective no longer shows up since it's not 3000+ level class.
Screen Shot 2021-03-25 at 20 22 22


tcho6319 (approved):
LGTM! Ty Sam :)

PR #390, opened 2021-03-26
Pre-allocate all subject colors

SamChou19815:
### Summary 

Fix the problem of flashing colors problem by pre-allocating all colors for all subjects during app start time.
I decide to keep the subject colors collection to respect user's existing subject color mappings.

### Test Plan 

Colors for all majors got added to the firestore
Screen Shot 2021-03-26 at 13 50 36


src/utilities.ts
```diff
@@ -43,6 +44,50 @@ export function getMinorFullName(acronym: string): string {
...
+export function allocateSubjectColor(subjectColors: Record<string, string>, subject: string): void {
+  if (subjectColors[subject]) return;
+  const usedColors = new Set(Object.values(subjectColors));
+  // Filter out used colors
+  const unusedSubjectColors = SUBJECT_COLORS.filter(color => !usedColors.has(color.hex));
+  const subjectColorCandidates =
+    unusedSubjectColors.length !== 0 ? unusedSubjectColors : SUBJECT_COLORS;
+  subjectColors[subject] =
+    subjectColorCandidates[Math.floor(Math.random() * subjectColorCandidates.length)].hex;
+}
```
SamChou19815, line 89: Moved from `global-firestore-data` and rewritten. Added a test to ensure correctness.

src/__test__/utilities.test.ts
```diff
@@ -0,0 +1,16 @@
...
+    const subjectColor: Record<string, string> = {};
+    allocateSubjectColor(subjectColor, 'CS');
+    allocateSubjectColor(subjectColor, 'MATH');
+    allocateSubjectColor(subjectColor, 'ECE');
+    allocateSubjectColor(subjectColor, 'INFO');
+    allocateSubjectColor(subjectColor, 'ISST');
+    allocateSubjectColor(subjectColor, 'ORIE');
+    expect(new Set(Object.values(subjectColor)).size).toBe(6);
+  }
+  // Cannot test other things since object key order is unreliable.
```
benjamin-shen, line 15: See my other comment about `allocateSubjectColor`; why is it important that the first 6 colors have no duplicates if the colors are pre-allocated for all (>>6) subjects?

If we still go with this implementation, I think you may still want to test >6 function calls because it's an edge case. For example, allocateSubjectColor 7 times and ensure that the 7th subject's color is valid / equal to one of the first 6. This can guard against errors in the `unusedSubjectColors` and `subjectColorCandidates` assignments.
I think the function is correct though.

src/utilities.ts
```diff
@@ -43,6 +44,50 @@ export function getMinorFullName(acronym: string): string {
   return requirementJSON.minor[acronym].name;
 }
 
+export function getAllSubjects(): ReadonlySet<string> {
+  const set = new Set<string>();
+  fullCoursesArray.forEach(it => set.add(it.subject));
+  return set;
```
benjamin-shen, line 50: Functionally, `return new Set<string>(fullCoursesArray.map(it => it.subject));`? Or something like that

src/utilities.ts
```diff
@@ -43,6 +44,50 @@ export function getMinorFullName(acronym: string): string {
...
+
+export function allocateSubjectColor(subjectColors: Record<string, string>, subject: string): void {
+  if (subjectColors[subject]) return;
+  const usedColors = new Set(Object.values(subjectColors));
+  // Filter out used colors
+  const unusedSubjectColors = SUBJECT_COLORS.filter(color => !usedColors.has(color.hex));
+  const subjectColorCandidates =
+    unusedSubjectColors.length !== 0 ? unusedSubjectColors : SUBJECT_COLORS;
+  subjectColors[subject] =
+    subjectColorCandidates[Math.floor(Math.random() * subjectColorCandidates.length)].hex;
```
benjamin-shen, line 88: What is the difference between this and assigning a random color to every subject?
It seems like `subjectColorCandidates` is only `unusedSubjectColors` for the first 6 subjects, and then colors are assigned randomly from then on.

benjamin-shen (commented):
Works for me after deleting my firestore data :)

Some nits and question about `allocateSubjectColor`.

src/utilities.ts
SamChou19815: Mostly trying to save some space by avoiding constructing array with a lot of duplicate elements.

src/__test__/utilities.test.ts
SamChou19815: I'm mostly trying to preserve the current behavior.

src/utilities.ts
SamChou19815: There might be no difference on the larger scale. Maybe we can just randomize everything since the first 6 subjects means nothing now...

src/utilities.ts
```diff
@@ -43,6 +44,50 @@ export function getMinorFullName(acronym: string): string {
...
+  },
+  {
+    text: 'Purple',
+    hex: 'C478FF',
+  },
+  {
+    text: 'Pink',
+    hex: 'F296D3',
+  },
+];
```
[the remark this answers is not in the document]
tcho6319: Good catch for the dupe info @benjamin-shen. Why do you think this would be a breaking change?

src/store.ts
```diff
@@ -306,8 +311,16 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
   const subjectColorUnsubscriber = fb.subjectColorsCollection
     .doc(simplifiedUser.email)
     .onSnapshot(snapshot => {
-      const subjectColors = snapshot.data() || {};
-      store.commit('setSubjectColors', subjectColors);
+      const subjectColors = snapshot.data();
+      if (subjectColors == null) {
+        store.commit('setSubjectColors', subjectColors);
```
tcho6319, line 316: If `subjectColors == null`, doesn't that mean the user has no doc for the `user-subject-colors` collection? Why would we commit null to the store? Would `store.commit('setSubjectColors', {})` be valid?

src/utilities.ts
tcho6319: Yeah could we possibly just randomize everything? I agree that there prob wouldn't be much meaning to the first 6 subjects that have colors assigned, especially if it ends up just being the first 6 subjects alphabetically. My only note in that case would be that the test would prob have to change. The best we could prob do for the test would be to make sure that all assigned colors are elements of `SUBJECT_COLORS` unless there is a smarter test we can think of..

src/store.ts
tcho6319: I believe it's causing the bug I took a video of. I deleted my `user-subject-colors` doc and tried to view courses on the requirements bar and add a course but got an error.

https://user-images.githubusercontent.com/54298311/112689263-6c463380-8e50-11eb-8aba-0ac65e7bd3f4.mov

tcho6319 (changes requested):
Thanks for the speedy PR Sam! I found a bug and responded to some of the discussions that @benjamin-shen brought up.

src/utilities.ts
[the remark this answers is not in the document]
benjamin-shen: `colors.ts` stores hexes as `#FFFFFF`, with a hashtag. The hexes in this PR and the hexes on firestore are like `FFFFFF`, without a hashtag. So, when firestore data is converted to color styling, a hashtag is added (see `Course.vue` for an example). To write a color (i.e. due to a user color change) to the database, the hashtagged hex code is first substringed (see `CourseMenu.vue`).

So I initially thought it would be a good idea to store the hashtag with the hex code. But, although it offers convenience and consistency on our side, this would also be a waste of data storage (because the hashtag is implied and would need to be stored `#users`*`#subjects` times). Perhaps it would be better to achieve consistency by omitting the hashtag (what is currently done in firestore), in which case it wouldn't be a breaking change and can be refactored in the future.

src/store.ts
SamChou19815: fixed

SamChou19815:
Now all the color allocation is completely randomized.

src/utilities.ts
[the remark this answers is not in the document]
tcho6319: Yes @benjamin-shen I think refactoring later so that we only have one place where we have subject colors make sense. I agree that omitting the hashtag would make it easier. I'll make a Notion item!

tcho6319 (approved):
LGTM!

PR #393, opened 2021-03-27
Add dry-run option for data migration script

SamChou19815:
### Summary 

As discussed in the meeting, it helps to add dry run options to help increase the confidence of data migration.
Now these scripts are runnable via `node [script.js] --dry-run`.

The dry-run option will write the changes to a file instead of actually writing to the database.

### Test Plan 

Cannot test now since I don't have prod credential. @tcho6319 can you help test this?

### Notes

It will be easier to review if you turn the option hide whitespace changes on
Screen Shot 2021-03-27 at 11 06 20

### DIFF
--- a/functions/split-user-data.js
+++ b/functions/split-user-data.js
@@ -17,24 +17,70 @@ const subjectColorsCollection = db.collection('user-subject-colors');
 const uniqueIncrementerCollection = db.collection('user-unique-incrementer');
 const onboardingDataCollection = db.collection('user-onboarding-data');
 
-userDataCollection.get().then(userData => {
-  Promise.all(
-    userData.docs.map(userDataDocument => {
-      const batch = db.batch();
+if (process.argv[2] === '--dry-run') {
+  userDataCollection.get().then(userData => {
+    const oldData = [];
+    const newUsernameCollection = {};
+    const newSemestersCollection = {};
+    const newToggleableRequirementChoicesCollection = {};
+    const newSubjectColorsCollection = {};
+    const newUniqueIncrementerCollection = {};
+    const newOnboardingDataCollection = {};
+    userData.docs.forEach(userDataDocument => {
       const data = userDataDocument.data();
-      batch.set(usernameCollection.doc(userDataDocument.id), data.name);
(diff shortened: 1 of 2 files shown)


tcho6319 (commented):
LGTM Sam! I think this will be helpful for checking if the semesters backward compatibility fix will work as expected for prod Firestore. I elevated your prod access to Editor but I think backing up the prod Firestore before we test this would be safest. I plan on running the dry run for these scripts to test later tonight when hopefully not a lot of users are on (after 10 PM) and after I back up prod Firestore.

functions/split-user-data.js
```diff
@@ -17,24 +17,70 @@ const subjectColorsCollection = db.collection('user-subject-colors');
...
-      const batch = db.batch();
+if (process.argv[2] === '--dry-run') {
+  userDataCollection.get().then(userData => {
+    const oldData = [];
+    const newUsernameCollection = {};
+    const newSemestersCollection = {};
+    const newToggleableRequirementChoicesCollection = {};
+    const newSubjectColorsCollection = {};
+    const newUniqueIncrementerCollection = {};
+    const newOnboardingDataCollection = {};
```
tcho6319, line 28: We could handle this in a different PR if you feel that it's outside of the scope of this PR, but I want to note that we should also make sure to create a collection for `user-selectable-requirement-choices` which is on the dev Firestore but not prod.  This would be reflected in the dry run and the actual data change processes in this script.

tcho6319 (approved):
Thank you so much Sam! Seems to work well now. I added a note about accounting for `user-selectable-requirement-choices` collection - which could be handled in a separate PR but I wanted to make you aware of it.

functions/split-user-data.js
SamChou19815: I don't think it's relevant here. `user-selectable-requirement-choices` is added after we do the split on dev, so it's never been in the user data collection anyways. Prod user also doesn't have this data anyways.
In terms of creating new collections, it's not an issue. Firestore will create a new collection for you automatically whenever you add the first document into that collection.

PR #383, opened 2021-03-24
Do not show ineligible courses in self-check add course modal

SamChou19815:
### Summary 

Resolve task 3 in [this notion item](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=a203824189894848bd7ce9275674aa96).

It adds an optional filter to `course-selector`, and make `new-self-check-course-modal` pass in the filter.

### Test Plan 

correctly filtered
Screen Shot 2021-03-24 at 12 30 46
true self-check not affected
Screen Shot 2021-03-24 at 12 33 11

PR #395, opened 2021-03-27
Self-check and related modal fixes

willespencer:
### Summary 

Fix a number of issues with self-check and related modals.

- [x] Filter out courses from the self-check dropdown if they do not pass that requirements checker (part 2 of the larger self-check fix)
- [x] Convert edit & delete sem to flexible modal (can now be clicked outside)
- [x] Have dropdowns reset properly for the new self check add modal and edit sem modal
- [x] Have all new semester dropdowns (so new semester, edit semester, and new self check add modal) have gray placeholder text until selected. 

image

### Test Plan 

1. Confirm that courses that do not fulfill self-check reqs with warnings do not show up in the dropdown, but do otherwise. An example would be "INFO 1998" when added to the plan should not show up in the CS External/Technical reqs, but should show up in the Engineering Advisor Approved Self-Check. A course like INFO 4300 should still show up in external/technical reqs.
2. Confirm edit semester and delete semester look and work as before, but can now be clicked outside
3. Confirm edit semester dropdowns are reset after choosing a season or year, and then clicking outside the modal, and clicking either the 'x' or 'cancel' button.
4. Confirm the same for the new self-check add modal
5. Confirm that edit semester, new semester, and self-check add modals all have gray text when opening now until the year/season is changed. Confirm they reset to gray when opening again.
6. In general, confirm courses can still be added from the self-check modal, reqs can still be assigned from self-check dropdown, and sems can still be added/edited/deleted


src/components/Modals/DeleteSemester.vue
```diff
@@ -1,45 +1,30 @@
...
-            <span class="deleteSemesterModal-button-left-text">Delete</span>
-          </div>
-        </button>
-      </div>
+  <flexible-modal
+    title="Delete Semester"
+    class="deleteSemesterModal"
+    content-class="content-delete"
+    left-button-text="CANCEL"
+    right-button-text="DELETE"
```
tcho6319, line 7: Can you do "Cancel" and "Delete"? It shouldn't be all caps

src/components/Modals/EditSemester.vue
```diff
@@ -1,51 +1,34 @@
...
-          </div>
-        </button>
-      </div>
-    </div>
-  </div>
+  <flexible-modal
+    title="Edit Semester"
+    content-class="content-semester"
+    left-button-text="CANCEL"
+    right-button-text="EDIT"
```
tcho6319, line 6: Same here. "Cancel" and "Edit"

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -133,6 +138,24 @@ export default Vue.extend({
...
+      if (requirement == null) return () => true;
+      const requirementSpec = getMatchedRequirementFulfillmentSpecification(
+        requirement,
+        toggleableRequirementChoices
+      );
+      // If a requirement is truly self-check, then all courses can be used.
+      if (requirementSpec == null) return () => true;
+      const eligibleCourseIds = new Set(requirementSpec.eligibleCourses.flat());
+      return eligibleCourseIds.has(course.crseId);
+    },
```
tcho6319, line 158: This could be good to put in `requirement-frontend-utils.ts` in case we want to reuse this functionality in other places later on.

tcho6319 (changes requested):
Overall LGTM Will! Left some nits that are design regressions and a suggestion for moving a function to the relevant utils file. Once the comments are addressed, this PR should be g2g.

src/components/Modals/DeleteSemester.vue
willespencer: Yep! It looks like we also use the capitalized text on the add course modal, should I change it there as well?

willespencer: Oh, and also on the see all buttons

src/components/Requirements/IncompleteSelfCheck.vue
willespencer: I agree, especially because I based this code heavily off of @SamChou19815 's code in #383 so there is already some code re-use. However, migrating it to `requirement-frontend-utils.ts` results in the following error when trying to import `import { VuexStoreState } from '@/store';`

`Dependency cycle via ./requirements/requirement-frontend-computation:5eslintimport/no-cycle`

Looks like this is because `store.ts` also imports `requirement-frontend-utils.ts.` Is there somewhere else I should put this code, or should I leave it like it is for now?

src/components/Modals/DeleteSemester.vue
tcho6319: Woah good catch! Yeah I think changing it for the add course modal and see all buttons to be consistent makes sense. @einc do you agree?

src/components/Requirements/IncompleteSelfCheck.vue
tcho6319: Ah right we want to avoid cyclical imports. I suspected this could be a problem....what if instead of  
`{ userRequirementsMap, toggleableRequirementChoices }: VuexStoreState` your function would have the params `userRequirementsMap: Readonly<Record<string, RequirementWithIDSourceType>>` and `toggleableRequirementChoices: AppToggleableRequirementChoices`? Then you wouldn't have to import the store. Does that make sense?

See:

https://github.com/cornell-dti/course-plan/blob/b5f1ae5fe25fc15f4c6d5aefe6c03dfca0da8911/src/store.ts#L46-L48

tcho6319 (approved):
LGTM! Thanks Will (and thanks for the documentation for the new functions!)

PR #401, opened 2021-03-28
RFC: choose a requirement for a course to fulfill by default

SamChou19815:
### Summary 

Based on feedback and user testing, it seems like the users are very confused when they add a class the requirement bar doesn't change at all. To clarify, there is no requirement bug or UI bug in this case, but more like a UX bug:

The requirement editor chooses nothing by default. Since the user doesn't read the text carefully, they will just click through with a course bind to no requirement.

The fix is easy. Choose one for them. I decided to first choose from related requirement, so that a known and solid requirement is prioritized.

### Test Plan 

Example of a course with related requirement:
Screen Shot 2021-03-28 at 14 49 14
Example of a course with only self-check requirement:
Screen Shot 2021-03-28 at 17 53 08

### Notes 

Maybe we also need to tweak the design a little. Right now it takes me a while to realized that the bolded requirement is the one you selected. Maybe we should just move the dropdown right in the page without click a button to find the dropdown?


hahnbeelee (approved):
LGTM!

PR #404, opened 2021-03-28
Fix subject color allocation

SamChou19815:
### Summary 

Currently subject color is broken, because we only allocate all of them if the user doesn't already have subject color. This breaks the existing user's data.

I changed it so that subject color is only fetched once, and passed it to allocator, and then set to store and firestore. This is safe since nowhere in the app we will change the subject color again.

### Test Plan 

- Delete all your subject color, reload. Colors are fine.
- Delete all your subject color but keep a few, reload. Colors are fine.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -307,18 +307,15 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
       selectableRequirementChoiceInitialLoadFinished = true;
       emitOnLoadWhenLoaded();
     });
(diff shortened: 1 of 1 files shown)


tcho6319 (commented):
LGTM. As per @willespencer's suggestion. Before I approve, I am going to send the staging link for this PR to Helen's thread in cp-roast to see if her issue is resolved as an extra check.

SamChou19815:
Got helen's confirmation it's working.

tcho6319 (approved):
LGTM!

PR #407, opened 2021-03-29
Update ILR, HADM, and AEM majors

einc:
### Summary 

Title. Most of the updates are with AEM concentrations. (some requirements aren't fully accurate with our infra so I tried my best and I left comments as needed)

I also updated Liberal studies for Engineering bc of double count issues.

### Test Plan 

- Test AEM, ILR school, and HADM major


src/requirements/data/colleges/en.ts
```diff
@@ -105,61 +106,73 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
           distribution => course.catalogDistr?.includes(distribution) ?? false
         ),
     ],
-    checkerWarning: 'We do not check that at least two courses are at the 2000 level or higher.',
+    checkerWarning:
+      'We do not check that all categories are being met and that at least two courses are at the 2000 level or higher.',
     fulfilledBy: 'courses',
     perSlotMinCount: [6],
   },
+  // TODO: Spliting Liberal Studies into 3 parts means they can all never be
```
einc, line 114: The liberal studies req couldn't currently be fulfilled (bc it would require triple counting). Splitting it into three requirements also meant that a lib study course couldn't meet other requirements too. So for now, just made it one requirement with a warning

src/requirements/data/colleges/en.ts
```diff
@@ -105,61 +106,73 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
...
-    description:
-      'Six credits of electives are required and must be approved by the student’s faculty advisor.',
-    source:
-      'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements',
-    fulfilledBy: 'self-check',
-    minCount: 6,
+    perSlotMinCount: [3],
   },
-  // TODO: INFO 1200 currently fulfills Eng Comm for ISST majors but should not be double-counted
+  // TODO: Remove warning once we can filter reqs based on majors (ISST) and petitions
```
einc, line 170: By default, INFO 1200 automatically fulfills this and can't be changed for the ISST major. This is wrong. Because we can't support petitions and don't have the option to edit requirements either, I changed this to include warnings so the req isn't wrong for ISST majors and more open ended

src/requirements/data/colleges/en.ts
SamChou19815: It can be fulfilled. Screenshot from master.
Screen Shot 2021-03-29 at 13 52 36
I think what we can do is to add another requirement that checks 2000 level or higher part. Then the warning can be removed. Then with the change in #401, it will be selected by default so user won't be confused.

SamChou19815 (approved):

src/requirements/data/majors/aem.ts
```diff
@@ -20,16 +20,295 @@ const aemRequirements: readonly CollegeOrMajorRequirement[] = [
     fulfilledBy: 'credits',
     perSlotMinCount: [55],
   },
+  // TODO: only supports one concentration
   {
     name: 'AEM Concentration Requirement',
+    // TODO: counting is temp courses. should eventually be credits
```
SamChou19815, line 26: The code that mandates only one slot for credits is here: https://github.com/cornell-dti/course-plan/blob/54a7e3e00bfee1b15edcec5f41e4ed5fcc6aede2/src/components/Requirements/SubRequirement.vue#L257-L269

I think the code can be easily extended to support multiple slots to support your use case, but the end result might be a little confusing for end users.

e.g. suppose we have 2 credits slots with `perSlotMinCount  = [6, 6]`. Suppose for each slot we have only one existing course of 3 credit. Then the frontend might render something like

```text
-- Add Course 1 ----------------------------
| Course foo for slot 1, already fulfilled |
--------------------------------------------

-- Add Course 2 ----------------------------
| unfulfilled course for slot 1            |
--------------------------------------------

-- Add Course 3 ----------------------------
| Course foo for slot 1, already fulfilled |
--------------------------------------------

-- Add Course 4 ----------------------------
| unfulfilled course for slot 1            |
--------------------------------------------
```

so we should probably do this for post launch after we have slot names?

einc: that sounds like a good idea, ty sam!

PR #411, opened 2021-03-31
Unbreak AP/IB requirement computation

### LINKED ISSUE #397: Incorrect courseplan AP credit logic
My AP credit is not carried over correctly. See gw284@cornell.edu 's account

I input that I have AP credit for:
AP Physics C E&M
AP Physics C Mechanics

These provide AP credit for PHYS 1112 and PHYS 2213

However, my dashboard says that I have not fulfilled any physics courses and need to take PHYS 1112 and PHYS 2213

The same problem occurs for chemistry and mathematics, but I took A level chemistry and got credit for those so I'm not sure how it should work exactly. I did indicate I took the placement test for chemistry and mathematics though

SamChou19815:
### Summary 

This bug is introduced during my live coding session. In that change, I will remove all edges when there is no user choice associated with a unique ID. However, I forget to consider the case when all the AP/IB equivalent courses all have the same unique ID -1. If we do the same for AP/IB courses, then all these courses will never be able to fulfill anything.

Therefore, we must bail out when `uniqueID == -1`.

### Test Plan 

Added a regression test about this bug.

### Notes 

This change will fix #397.

### DIFF
--- a/src/requirements/requirement-graph-builder.ts
+++ b/src/requirements/requirement-graph-builder.ts
@@ -102,6 +102,10 @@ const buildRequirementFulfillmentGraph = <
 
   // Phase 3: Respect user's choices on double-counted courses.
   userCourses.forEach(({ uniqueId }) => {
+    // uniqueId === -1 means it's AP/IB course.
+    // User never gets to make a choice about these courses, so it will never appear in the choices.
+    // Therefore, removing those edges will nullify all these credits.
+    if (uniqueId === -1) return;
     const chosenRequirement = userChoiceOnDoubleCountingElimination[uniqueId];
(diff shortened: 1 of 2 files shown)


src/requirements/requirement-graph-builder.ts
```diff
@@ -102,6 +102,10 @@ const buildRequirementFulfillmentGraph = <
 
   // Phase 3: Respect user's choices on double-counted courses.
   userCourses.forEach(({ uniqueId }) => {
+    // uniqueId === 1 means it's AP/IB course.
```
benjamin-shen, line 105: `=== -1`?

benjamin-shen (approved):
That makes sense. Thanks Sam!!

PR #413, opened 2021-03-31
Onboarding and requirements sidebar frontend bugs

willespencer:
### Summary 

This pull request fixes a small number of front-end issues with the Onboarding modal and Requirements sidebar ([Notion](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=9a7345937b304fdcbedba87b6daeb8e0))

- [x] Added a label for college in the reqs sidebar like for majors and minors
- [x] Flipped AP/IB credit dropdown numbers
- [x] Implement text-wrapping for long AP/IB test names
- [x] Line up onboarding dropdowns with their parent input 

image

image

Remaining TODOs:

- [ ] I've already mentioned this to @tcho6319, but the major/minor labels now have hover states even when not clickable. This now applies to the college label as well.
- [x] The requirement header code is kind of messy (styling in the html) and replicated heavily between college/major/minor. Could be a low priority task post-launch to clean it up (@tcho6319 lmk if I should make a notion item)

### Test Plan 

1. Confirm there is now a label for all colleges on the requirements bar
2. Confirm text wraps for long AP/IB test names (like BC non-engineering) on small break points
3. Confirm dropdown children now line up with their input boxes
4. Outside of the above changes, confirm all onboarding dropdowns look the same on the webapp/figma
5. Confirm AP/IB score dropdowns are now from top to bottom (and save that way)

### DIFF
--- a/src/components/Modals/Onboarding/Onboarding.scss
+++ b/src/components/Modals/Onboarding/Onboarding.scss
@@ -52,7 +52,7 @@ input {
   }
 
   &-main {
-    background: #ffffff;
+    background: $white;
     border-radius: 9px;
     margin-left: auto;
     margin-right: auto;
@@ -135,7 +135,7 @@ input {
   }
   &-subsection {
     width: 100%;
-    border: 1px solid #c4c4c4;
+    border: 1px solid $inactiveGray;
     box-sizing: border-box;
     border-radius: 9px;
     padding: 3rem;
@@ -160,13 +160,13 @@ input {
     &--font {
       color: $darkGray2;
       flex-direction: row;
-      background-color: white;
+      background-color: $white;
     }
     &--review {
       font-weight: normal;
       padding: 5px;
       margin-left: 10px;
-      background-color: white;
+      background-color: $white;
       color: $lightPlaceholderGray;
       font-size: 16px;
     }
@@ -180,9 +180,9 @@ input {
     margin-bottom: 0.75rem;
 
     &-fillRow {
-      border: 1px solid #ffffff;
+      border: 1px solid $white;
       width: 100%;
-      border-bottom-color: #c4c4c4;
+      border-bottom-color: $inactiveGray;
       margin-bottom: 1.5rem;
       margin-top: 20px;
     }
@@ -209,7 +209,7 @@ input {
       justify-self: end;
       flex-direction: row;
       &---bold {
-        color: black;
+        color: $black;
         font-weight: bold;
         margin-right: 5px;
       }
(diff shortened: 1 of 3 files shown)


src/components/Requirements/RequirementHeader.vue
```diff
@@ -8,7 +8,27 @@
...
+            'font-weight': '500',
+            color: `#${reqGroupColorMap[req.groupName][0]}`,
+          }"
+          class="college-title-top"
+        >
+          {{ getCollegeFullName(onboardingData.college) }}
+        </p>
+      </button>
+    </div>
+    <!-- TODO change for multiple colleges -->
```
SamChou19815, line 30: this comment is copied over twice. remove the one above

SamChou19815 (approved):
nit

src/components/Requirements/RequirementHeader.vue
willespencer: I purposefully copied it over because the indexing in both `v-if`s would be broken if we ever implemented multiple colleges. Do you still think it should be removed?

SamChou19815: Ok then i think that's fine.

src/components/Modals/Onboarding/Onboarding.scss
```diff
@@ -237,8 +237,9 @@ input {
     line-height: 17px;
     color: $lightPlaceholderGray;
     width: 100%;
-    height: 1.75rem;
-    padding-left: 8px;
+    height: auto;
+    min-height: 1.75rem;
+    padding: 5px 8px;
     background-color: white;
```
tcho6319, line 243: Nit: Can we replace these colors with scss variables?

tcho6319 (approved):
Thanks Will! LGTM but left a couple nits.

1. Can we move the dropdown content a little lower? It seems to be covering the input box.
Screen Shot 2021-03-31 at 12 54 56 AM

Can we try to have something like this:
Screen Shot 2021-03-31 at 12 56 35 AM





> I've already mentioned this to @tcho6319, but the major/minor labels now have hover states even when not clickable. This now applies to the college label as well.

I've assigned this task to myself. But I will keep in mind to do this fix for college as well!: [Conditionally add hover state to major/minor buttons if there are more than 1 majors/minors](https://www.notion.so/Conditionally-add-hover-state-to-major-minor-buttons-if-there-are-more-than-1-majors-minors-4f12bb92aa8445a3b3f5552898bc9ffb)

> The requirement header code is kind of messy (styling in the html) and replicated heavily between college/major/minor. Could be a low priority task post-launch to clean it up (@tcho6319 lmk if I should make a notion item)

Yes, looks like a good task for Code Clean Up :). Can you make sure to add the Codebase tag when you make the Notion item? Thanks Will!

tcho6319 (approved):
LGTM!

PR #391, opened 2021-03-27
Hover States

tcho6319:
### Summary 

This pull request adds hover states to buttons (i.e. clickable elements) on CP. I globally set buttons to have 0.5 opacity on hover in `src/assets/scss/_global.scss`. This means that every `button` on CP should have a hover state. 

**NOTE: Some buttons like the nav bar icons, course/semester menu items, semester view compact/not compact toggle buttons should have full opacity so I set the `opacity: 1` to those less frequent special cases.**

For Notion item: [Add hover states to buttons](https://www.notion.so/Add-hover-states-to-buttons-ae61c3e48dc145edad9fd69444b69e96)

- Onboarding:
   - [x] Change "Next" & "Finish" button hover states to 50% lighter of the same color (it's currently gray)
   - [x] Change the "< Previous" text to also have 50% lighter hover.
   - [x] Add hover state to the "+ add another major", "edit" text

- Add Semester
   - [x] Cancel and Add buttons. Change text to "Cancel" and "Add". (In general, CP follows this pattern) Add hover states for them. For "Add" make the background color lighter.
   - [ ] For "Cancel" button, only make text lighter **NOTE: I could not figure out how to do this. If anyone knows, pls comment!**

- Edit Semester
   - [x] Cancel and Edit buttons. Add hover states. Please also center text horizontally and vertically.
 
- [x] Changed some "X" imgs to buttons and some clickable divs to buttons

### Test Plan 
Be sure to look through the site and ensure that nothing visually changed (besides the added hover states). Note that many requirements bar dropdown titles, major/minor names have a hover state now because they are all buttons. Lmk if you want them to be full opacity (`opacity: 1`).

PR #416, opened 2021-03-31
Conditional Hover States for College/Major/Minor Tabs

tcho6319:
### Summary

This pull request implements conditional hover states for the College/Major/Minor tabs on the requirements sidebar.

- [x] College tab always has `opacity: 1` on hover since we only support one college
- [x] Major/Minor tabs will have `opacity: 1` on hover unless there are more than 1. If there are more than 1 major/minor tabs, then `opacity: 0.5` on hover, which is the default for all buttons.
- [x] Refactor special cases of `opacity: 1` on hover with `'full-opacity-on-hover'`class  and `opacity: 0.8` on hover with `'slightly-lower-opacity-on-hover'` class found in `src/assets/scss/_global.scss`.  See #391 for original hover states implementation of special cases.

### Test Plan 
- College should not have lower opacity on hover
- Major/Minor should not have lower opacity on hover if there is only 1 Major and/or 1 Minor
- Add a second Major: Now the Major tabs should have lower opacity on hover
-  Add a second Minor: Now the Minor tabs should have lower opacity on hover

- Be sure to look through the site and ensure that nothing visually changed (besides the added hover states)

### DIFF
--- a/src/assets/scss/_global.scss
+++ b/src/assets/scss/_global.scss
@@ -245,3 +245,20 @@ form {
 .fade-leave-to {
   opacity: 0;
 }
+
+// full opacity on hover.
+// Intended to undo reduced opacity for button:hover
+.full-opacity-on-hover {
+  &:hover {
+    opacity: 1;
+  }
+}
+
+// slightly lower opacity on hover.
+// Intended to have higher than the reduced opacity for button:hover but lower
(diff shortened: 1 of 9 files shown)


SamChou19815 (approved):
LGTM

PR #417, opened 2021-03-31
Give every user's AP/IB equivalent course a negative unique ID

SamChou19815:
### Summary 

The requirement graph is implemented as a hashtable with requirement name and unique ID used as key for both sides. It really doesn't like two courses to have the same unique ID.

Unfortunately, that's the case for AP/IB credits' equivalent courses. They are all hardcoded to have `uniqueID === -1`. This causes issues like #397 to occur when a requirement is connected to two AP credits, only one is recorded in the graph due to unique ID duplication.

This PR tries to fix the problem by giving all AP/IB equivalent course a NEGATIVE unique id, starting from -2. (-1 is reserved for swim test). The unique IDs are provisioned according to the order they appear in the user exam object. The stability of the IDs are not that important, since those IDs are never stored in the database, and will always be recomputed on the fly when the requirement is recomputed.

### Test Plan 

It should fix @abcdefguan's problem when only one physics AP credit is counted.
Screen Shot 2021-03-31 at 12 26 07

### DIFF
--- a/src/requirement-types.d.ts
+++ b/src/requirement-types.d.ts
@@ -56,7 +56,7 @@ type DecoratedCollegeOrMajorRequirement = RequirementCommon &
 type CourseTaken = {
   /** The course ID from course roster, or our dummy id to denote special courses like FWS equiv. */
   readonly courseId: number;
-  /** Using the unique ID of firestore course for real course, and less than 0 for AP/IB/Swim */
+  /** Using the unique ID of firestore course for real course, -1 for swim test and < -1 for AP/IB. */
   readonly uniqueId: number;
   /**
    * Course code like 'CS 2112', 'AP CS'.
--- a/src/requirement-types.d.ts
+++ b/src/requirement-types.d.ts
@@ -56,7 +56,7 @@ type DecoratedCollegeOrMajorRequirement = RequirementCommon &
 type CourseTaken = {
   /** The course ID from course roster, or our dummy id to denote special courses like FWS equiv. */
   readonly courseId: number;
-  /** Using the unique ID of firestore course for real course, and -1 for AP/IB/Swim */
+  /** Using the unique ID of firestore course for real course, and less than 0 for AP/IB/Swim */
   readonly uniqueId: number;
   /**
    * Course code like 'CS 2112', 'AP CS'.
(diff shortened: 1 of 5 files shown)


src/requirement-types.d.ts
```diff
@@ -56,7 +56,7 @@ type DecoratedCollegeOrMajorRequirement = RequirementCommon &
 type CourseTaken = {
   /** The course ID from course roster, or our dummy id to denote special courses like FWS equiv. */
   readonly courseId: number;
-  /** Using the unique ID of firestore course for real course, and -1 for AP/IB/Swim */
+  /** Using the unique ID of firestore course for real course, and less than 0 for AP/IB/Swim */
```
tcho6319, line 59: Nit: can we specify -1 is reserved for Swim and <-1 is for AP/IB?

src/requirements/requirement-graph-builder.ts
```diff
@@ -102,10 +102,10 @@ const buildRequirementFulfillmentGraph = <
 
   // Phase 3: Respect user's choices on double-counted courses.
   userCourses.forEach(({ uniqueId }) => {
-    // uniqueId === -1 means it's AP/IB course.
+    // uniqueId < 0 means it's AP/IB course.
```
tcho6319, line 105: Again could we specify that AP/IB course if uniqueId < -1 since it's swim test for uniqueId === -1

tcho6319 (changes requested):
LGTM overall but had a few questions that should be addressed before approval:

1. This change should also work with any AP/IB courses that will be used for FWS and have `FWS_COURSE_ID` right? Note: Currently, AP/IB credit is not being applied to requirements that take FWS courses. This is a bug that @benjamin-shen and @hahnbeelee will work on tomorrow. 
2. In the test below, would you have to update the `uniqueId` to be a negative number other than -1 to be consistent with the choice to reserve swim test for uniqueId: -1?

https://github.com/cornell-dti/course-plan/blob/132368d9ce7816236df0df64e30a9cbd66db2074/src/requirements/__test__/requirement-graph-builder.test.ts#L174-L191

tcho6319 (approved):
Ty Sam. LGTM

PR #419, opened 2021-03-31
[easy] Use ts-node to run requirement-generator script

SamChou19815:
### Summary 

Use ts-node to run the script without compiling all the code first. I also use the `-T` option to do transpilation only (no type checking) to speed things up. (Type checking will be done in CI anyways.)

On my machine, it improves the requirement generator runtime from 10s to 4s.

### Test Plan 

`npm run req-gen`


tcho6319 (approved):
LGTM! Ty Sam

PR #420, opened 2021-03-31
Fix engineering liberal studies requirement by restoring all checkers

SamChou19815:
### Summary 

This is a strict improvement over what we have now. The current data doesn't support one categories with multiple classes. In addition, it can check all possible cases.

According to the discussion in slack, we will still make the liberal studies as allow double counting while they are not, since it seems to have limited impact and can immediately resolve bugs now. For principled fixes would come after the lunch.

### Test Plan 

Add six liberal studies courses to courseplan and see requirements fulfilled


tcho6319 (approved):
LGTM! @einc are you ok with this? Would this make PR #412 irrelevant?

SamChou19815:
This this will make #412 irrelevant.

PR #421, opened 2021-03-31
Semesters Backwards Compatibility Script

tcho6319:
### Summary

This pull request resolves the semesters backwards compatibility issue.

Currently, on [courseplan.io](courseplan.io), users who added semesters before CP decided to order semesters from most recent to least recent semester may end up with a schedule where the semesters are out of order.

See the example of the bug below. The schedule has the old order (least recent to most recent semester) so that Spring 2015 appears above Winter 2028. The add new semester functionality is up-to-date with our decision to have semesters go in most recent to least recent. Therefore, when I add Spring 2021, it is added to the top.

https://user-images.githubusercontent.com/54298311/113212182-6d0e0980-9244-11eb-8ba6-dba1a8dee1cc.mov

**The solution for this issue is to create a script that we would run during the data migration at launch so that the semesters are ordered correctly from most recent to least recent.**

- [x] Implemented `update-user-data-semesters.js` with dry run option

### Test Plan

- Navigate to `functions\` folder
- Run `node update-user-data-semesters.js --dry-run`
- Compare output files: `'old-userData-semesters.json'` and `'new-userData-semesters.json'`

**NOTE: ** In order to do the test plan, you will need access to prod. If you do not have access, but would like to test, slack @tcho6319 so that she can screenshare and show you the files generated. The output files will not be committed as they contain the emails of our users.

### Notes
I'm confused as to why the semester ordering is correct (i.e. Winter 2028, Spring 2021, Spring 2015) in `'old-userData-semesters.json'` because on Firestore, the ordering is incorrect (Spring 2021, Spring 2015, Winter 2028). The good thing is that the semester ordering is correct in `'new-userData-semesters.json'`, but if anyone might have an idea of why, please lmk!

Screen Shot 2021-03-31 at 5 20 22 PM

### DIFF
--- a/functions/update-user-data-semesters.js
+++ b/functions/update-user-data-semesters.js
@@ -81,7 +81,8 @@ const transformData = data => ({
 if (process.argv[2] === '--dry-run') {
   userDataCollection.get().then(userData => {
     const oldUserData = userData.docs.map(doc => ({ id: doc.id, ...doc.data() }));
-    const newUserData = oldUserData.map(transformData);
+    // Using JSON parse and stringify trick so we get a deep copy of the nested oldUserData
+    const newUserData = JSON.parse(JSON.stringify(oldUserData)).map(transformData);
     fs.writeFileSync('old-userData-semesters.json', JSON.stringify(oldUserData, undefined, 2));
     fs.writeFileSync('new-userData-semesters.json', JSON.stringify(newUserData, undefined, 2));
   });
--- a/functions/update-user-data-semesters.js
+++ b/functions/update-user-data-semesters.js
@@ -1,0 +1,97 @@
+// Updates the order of the semesters in the userData collection to be in order
+// from most recent to least recent
+
+const admin = require('firebase-admin');
+const fs = require('fs');
+const serviceAccount = require('./service-account.json');
+
+admin.initializeApp({
+  credential: admin.credential.cert(serviceAccount),
+  databaseURL: 'https://cornelldti-courseplan-dev.firebaseio.com',
+});
+
+const db = admin.firestore();
+const userDataCollection = db.collection('userData');
+
+/**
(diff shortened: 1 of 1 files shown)


SamChou19815 (approved):
The changes look good to me. I think the order look correct in the old one because you are using the dev service account for testing?

tcho6319:
> The changes look good to me. I think the order look correct in the old one because you are using the dev service account for testing?

@SamChou19815 I was using the prod service account to generate the output files so still confused about how the order is correct. But according to the `'new-userData-semesters.json'` it seems the script does have semesters in the right order so I guess we'll go with it for now?

SamChou19815:
@tcho6319 oh I know why. sort will mutate the original array, so what you get is some identical list. Could you copy over the original array using `[...oldData]` and then see whether the results are still good?

functions/update-user-data-semesters.js
```diff
@@ -0,0 +1,98 @@
...
+const transformData = data => ({
+  ...data,
+  semesters: semesterObjectsByMostRecentTypeAndYear(data.semesters),
+});
+
+if (process.argv[2] === '--dry-run') {
+  userDataCollection.get().then(userData => {
+    const oldUserData = userData.docs.map(doc => ({ id: doc.id, ...doc.data() }));
+    // Using JSON parse and stringify trick so we get a deep copy of the nested oldUserData
+    const newUserData = JSON.parse(JSON.stringify(oldUserData)).map(transformData);
```
tcho6319, line 85: At first I tried `[...oldUserData]` but it makes a shallow copy for the nested parts of `oldUserData` so the sort function modified oldUserData and the old output was ordered. Instead I used the parse and stringify trick to create a deep copy of the nested `oldUserData`. Now, `oldUserData` shows what is on Firestore.

SamChou19815: @tcho6319  ok looks good, then I think this PR is good to go then.

PR #425, opened 2021-04-01
Fixed long a&s names and hover bugs.

neketka:
### Summary 

- [x] Added 12rem max width for rec names
- [x] Added cursor: default for logo

### Test 

Check that wrapping works correctly and the logo doesn't hover anymore.

Before:
image
After:
image

### DIFF
--- a/src/components/Requirements/SubRequirement.vue
+++ b/src/components/Requirements/SubRequirement.vue
@@ -572,7 +572,7 @@ button.view {
   &-name {
     text-align: left;
     margin-left: 11px;
-    max-width: 14rem;
+    max-width: 12rem;
   }
 }
 </style>
--- a/src/components/Requirements/SubRequirement.vue
+++ b/src/components/Requirements/SubRequirement.vue
@@ -572,6 +572,7 @@ button.view {
   &-name {
     text-align: left;
     margin-left: 11px;
+    max-width: 14rem;
   }
 }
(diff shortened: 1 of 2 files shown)


SamChou19815:
> Check that wrapping works correctly and the logo doesn't hover anymore.

@neketka could you provide some example screenshots when the master deployment is bad but after your change looks good?

src/components/Requirements/SubRequirement.vue
```diff
@@ -572,6 +572,7 @@ button.view {
   &-name {
     text-align: left;
     margin-left: 11px;
+    max-width: 12rem;
```
tcho6319, line 575: Nit: Could we do `max-width: 11rem`? 

When `max-width: 12rem`, it would be better if the "15/18 credits" line for the Liberal Studies Distribution: 18 credits requirement weren't broken up into 2 lines as shown below:

Screen Shot 2021-04-01 at 8 32 29 PM

If we make `max-width: 11rem`, then "15/18 credits" fits on one line.
Screen Shot 2021-04-01 at 8 34 03 PM

tcho6319 (approved):
LGTM but left a nit that would be great if you could address! Also, I echo @SamChou19815's request to include screenshots, as they help describe the bug that you resolved. I would also add a link to the relevant Notion item as well.

PR #428, opened 2021-04-01
AP/IB credit bug fixes

benjamin-shen:
### Summary 

This pull request fixes some AP/IB and checker-related bugs. It also fixes the issue of AP/IB credits not showing up under self-check reqs.

- [x] Fixed typos (missing space) for ENG advisor-approved and CS major-approved electives
- [x] Fixed checkers (disallow special courses, change logic) for ENG advisor-approved and CS major-approved electives
- [x] Fixed FWS checker (allow special FWS course), allowed FWS to apply to all colleges
- [x] Fixed AP/IB credits not showing up under self-check reqs

### Test Plan 

[These bugs](https://www.notion.so/courseplan/AP-credit-bugs-b090af7ffaea45d2886ea70cce8f70da) should now **all** be fixed :)


src/requirements/data/checkers-common.ts
```diff
@@ -38,9 +39,19 @@ export const courseMatchesCodeOptions = (course: Course, codeOptions: readonly s
...
   course.titleLong.includes('FWS:') ||
   (course.catalogSatisfiesReq?.includes('First-Year Writing Seminar') ?? false);
 
+/**
+ * Detects special (synthetic) courses, as defined in requirement-json-generator.ts
+ *
+ * @param course course object with useful information retrived from Cornell courses API.
+ * @returns if the course is a special course
+ */
+export const courseIsSpecial = (course: Course): boolean =>
```
tcho6319, line 52: Can you more explicitly state what a "special course" is?

src/requirements/data/colleges/en.ts
```diff
@@ -139,14 +140,15 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
...
-      'At least 3 credit hours total. All academic courses count.' +
+      'At least 3 credit hours total. All academic courses count. ' +
       'No PE courses, courses numbered 10xx, and ROTC courses below the 3000-level allowed.',
     source:
       'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirement',
     checker: [
       (course: Course): boolean => {
-        const { catalogNbr } = course;
-        return !(ifCodeMatch(course.subject, 'PE') && ifCodeMatch(catalogNbr, '10**'));
+        if (courseIsSpecial(course)) return false;
```
tcho6319, line 149: Why would the checker return `false` for a special course (i.e. an FWS, course that applies to total academic credits requirement) if all academic courses count?

src/components/Requirements/SubRequirement.vue
```diff
@@ -305,9 +304,38 @@ export default Vue.extend({
         ? `${this.subReq.minCountFulfilled}/${this.subReq.minCountRequired} ${this.subReq.fulfilledBy}`
         : 'self check';
     },
-    fulfilledSelfCheckCourses(): readonly FirestoreSemesterCourse[] {
-      const reqId = this.subReq.requirement.id;
-      return store.state.derivedSelectableRequirementData.requirementToCoursesMap[reqId];
+    fulfilledSelfCheckCourses(): readonly CourseTaken[] {
```
tcho6319, line 307: Some comments in this function would be helpful! Like about `selectedCourses` v `fulfilledCourses`. Why we filter `fulfilledCourses` for just AP/IB, etc.

src/requirements/data/majors/cs.ts
```diff
@@ -119,14 +120,15 @@ const csRequirements: readonly CollegeOrMajorRequirement[] = [
...
-      'At least 3 credit hours total. All academic courses count.' +
+      'At least 3 credit hours total. All academic courses count. ' +
       'No PE courses, courses numbered 10xx, and ROTC courses below the 3000-level allowed.',
     source:
       'https://www.cs.cornell.edu/undergrad/rulesandproceduresengineering/choosingyourelectives',
     checker: [
       (course: Course): boolean => {
-        const { catalogNbr } = course;
-        return !(ifCodeMatch(course.subject, 'PE') && ifCodeMatch(catalogNbr, '10**'));
+        if (courseIsSpecial(course)) return false;
```
tcho6319, line 129: Same here. Why would the checker return false for a special course (i.e. an FWS, course that applies to total academic credits requirement) if all academic courses count?

tcho6319 (changes requested):
LGTM but left some questions and nits that should be addressed before approval!

src/requirements/data/checkers-common.ts
benjamin-shen: It's explicitly defined in `requirement-json-generator.ts`. I'm hesitant to include something like "credits-only or FWS equivalent course" here in case we modify the special courses in the future, because then we'd need to change the documentation in multiple places. Maybe there should be another location where it's documented?
image

src/requirements/data/colleges/en.ts
benjamin-shen: I don't think FWS courses count, and "All academic courses count" is a bit misleading.
image

src/requirements/data/majors/cs.ts
benjamin-shen: See my comment above^

src/requirements/data/checkers-common.ts
tcho6319: I see your point. Since you did indicate that specialCourses are defined in requirement-json-generator.ts in the spec, I think this should be ok for now.

src/requirements/data/colleges/en.ts
tcho6319: I see. @einc is there a way to reform this sentence so that it's clearer for the users? Like "All academic courses, except FWS, count."

tcho6319 (approved):
LGTM!

src/requirements/data/colleges/en.ts
benjamin-shen: If we reform the wording, we should also make sure to update the checkers. Currently the checkers allow regular FWS courses (other than our special FWS equivalent course) to fulfill the reqs because of the original wording.

PR #429, opened 2021-04-03
EN and CS major are fully updated

einc:
### Summary 

- Added FL checker for liberal studies (FL courses don't have the ‘FL’ category distribution) 
- Filtered out CS classes for CS external spec
- Fixed broken link & description for advisor approved electives

### Test Plan 

- EN school & CS major requirements should be as perfect as we can get them now. 
- Trying adding a language course (it should fulfill liberal studies now)
- I tested links, checked descriptions, and did as much filtering as we could for now!


tcho6319 (approved):
LGTM!

PR #433, opened 2021-04-03
always expand bottom bar when clicking course

jason-tung:
### Summary 

https://www.notion.so/courseplan/clicking-on-a-course-when-bottom-bar-is-already-open-but-uncollapsed-doesn-t-open-it-7914556251d14f02b2d44b8983f1bdd5

### Test Plan 

Play around with bottom bar on staging to make sure the only changed behavior is clicking course while bottom bar is closed will open the bottom bar and focus the course.

### DIFF
--- a/src/components/BottomBar/BottomBarState.ts
+++ b/src/components/BottomBar/BottomBarState.ts
@@ -66,6 +66,7 @@ export const addCourseToBottomBar = (course: FirestoreSemesterCourse): void => {
       vueForBottomBar.bottomCourseFocus = i;
       return;
     }
+    vueForBottomBar.isExpanded = true;
   }
 
   if (vueForBottomBar.bottomCourses.length === 0) {


src/components/BottomBar/BottomBarState.ts
```diff
@@ -66,6 +66,7 @@ export const addCourseToBottomBar = (course: FirestoreSemesterCourse): void => {
       vueForBottomBar.bottomCourseFocus = i;
       return;
     }
+    vueForBottomBar.isExpanded = true;
```
SamChou19815, line 69: The bottom bar won't expand if it returns early as above.

tcho6319 (approved):
LGTM!

It'd be nice if we also expand the bottom bar when we move a bottom bar course from see more to the first.

Adding `vueForBottomBar.isExpanded = true` in `moveBottomBarCourseToFirst`

https://github.com/cornell-dti/course-plan/blob/38707b10fe50dd30c850d2a817f04a867c73fc26/src/components/BottomBar/BottomBarState.ts#L124-L132

src/components/BottomBar/BottomBarState.ts
tcho6319: Yes, +1 this ^^

PR #434, opened 2021-04-04
Add privacy policy page

willespencer:
### Summary 

This pull request adds our own privacy policy page to CoursePlan

- [x] Add privacy policy page to router and add link to footer
- [x] Update privacy policy text as provided by @einc 
- [x] Slightly style html text and update top bar styling

image

### Test Plan 

1. Confirm that the new privacy policy page can be reached by clicking on the link in the footer.
2. Confirm that the styling looks nice
3. Confirm the 404 page can still be accessed by typing in a malformed url, and that it still looks the same.
4. Confirm that CoursePlan can still be logged in and out of.

### Notes 

We switched from using term.ly to this page on our own domain because term.ly was not allowing us to use their own page without paying. The html they provided to copy was extremely malformed, and was unable to be fixed in 30 minutes. Instead, @einc generated this text with getterms.io with some of the term.ly text (like hotjar) copied in.

### DIFF
--- a/src/components/Footer.vue
+++ b/src/components/Footer.vue
@@ -13,12 +13,7 @@
         </p>
       </div>
       <div class="col-3 footer">
-        <a
-          class="footer"
-          target="_blank"
-          href="https://app.termly.io/document/privacy-policy/fcecc0e8-8af2-472d-8d27-b6b89d02a2be"
-          >Privacy Policy</a
-        >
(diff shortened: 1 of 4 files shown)


einc (approved):
Went through the test plan & it all worked! TYSM for doing this Will!!!!!!!

tcho6319 (approved):
LGTM! Tysm :)

PR #435, opened 2021-04-04
Update hist, ORIE, mechE, comm major

einc:
### Summary 

Title.
Note: A lot of the broad checkers for requirements with warnings are contributing to the size increase. Post launch, we should write special functions for these.

### Test Plan 

Test hist, ORIE, mechE, comm majors for no breaking changes.


src/requirements/data/majors/orie.ts
```diff
@@ -1,16 +1,25 @@
...
+import { CollegeOrMajorRequirement, Course } from '../../types';
+import { ifCodeMatch, includesWithSubRequirements } from '../checkers-common';
 
 const orieRequirements: readonly CollegeOrMajorRequirement[] = [
+  {
+    name: 'Engineering Distribution Courses',
+    description:
+      'ENGRD 2700 and ENGRD 2110 is recommended as the second Engineering Distribution course. ',
+    source: 'https://www.orie.cornell.edu/orie/programs/undergraduate-programs/degree-requirements',
+    checker: includesWithSubRequirements(['ENGRD 2700']),
```
tcho6319, line 10: Why do we not add ENGRD 2110 here?

tcho6319 (approved):
Left a small comment/question but LGTM overall

src/requirements/data/majors/orie.ts
einc: it's under college & only recommended, not required. eventually we should remove it from college and just add to major

PR #436, opened 2021-04-04
Public policy minor

willespencer:
### Summary 

This pull request adds the public policy minor to CoursePlan

image

### Test Plan 

1. Confirm that the minor can be completely fulfilled.
2. Confirm that the gateway course and methods courses cannot be used as electives

### Notes 

Just a couple things

- When trying to exclude 'GOVT 3032' from the electives checker, it still could fulfill it until I also excluded its cross-listed course code. Is this intentional / can we fix this?
- This is the first requirement I've written from scratch, so please let me know if I did anything incorrectly!


einc (approved):
LGTM!! For note #2, yea this is a problem (cross-listed courses) that we'll address post launch!

tcho6319 (approved):
LGTM

PR #430, opened 2021-04-03
Dynamic z index

hahnbeelee:
### Summary
This PR:
- [x] Ensures that when you are on mobile and have the requirements bar open, the hamburger menu properly triggers the navbar
image
- [x] When you open any modal it should be above everything else
image
- [x] Works on Safari
image
- [x] Fixed funky Safari Desktop modals
- [ ] Works on Safari mobile
IMG_DADDCC07091B-1
For some reason, things look different on Safari on my MacBook and Safari on my iPhone. For now, the Hamburger button works and things look way better than they used to on Safari Desktop:
Before:
image
After:
image

Majority of the main functionality on mobile/desktop is implemented, so i feel comfortable merging this despite the modals from within the requirements bar still not working 100% as desired. This PR allows the user to perform majority of the functionality.

Components inherit the z-index of their parents. Therefore, the modals within the requirement bar need to sit above everything, but the requirement bar needs to be a lower z-index than the navbar. This simply isn't possible with the architecture we have right now.
image

This is a _temporary fix_ for launch. We're going to need to do a refactor sometime post launch.

### Test Plan 
1. Go on mobile
  - Go to requirements
  - Click on the hamburger menu again and ensure that it triggers the side nav bar
2. Open up every instance of a modal you're aware of and ensure that it shows up above the navbar
  - Repeat on mobile
  - Repeat on Safari

# NOTE
All the code in this PR should be removed sometime in the future with a giant refactor.

PR #437, opened 2021-04-04
Delete sem z-index fix

willespencer:
### Summary 

This pull request fixes a bug where the delete semester modal does not properly display above the navbar like the rest of the modals now do.

(Before Images)
image
image

### Test Plan 

Test that this modal shows up above the nav-bar (and does not look like the above images) for both mobile and desktop.

### Notes 

Two things:
1. I thought I was running into this bug for the reset modal as well (and sent screenshots to @hahnbeelee) but I can't reproduce it anymore. Not sure if this could be at all related to this problem existing for Safari in #430 even though I'm on Chrome, or if I just imagined this problem? 
2. It's not a good sign that after all the z-index testing this modal still slipped through... there were also a ton of other random z-index modal styles I found throughout the codebase when trying to figure out which one would fix this issue. Should I make a notion item to try and re-factor modals more to remove outdated and duplicated code @tcho6319 ?

### DIFF
--- a/src/components/Semester/Semester.vue
+++ b/src/components/Semester/Semester.vue
@@ -527,7 +527,7 @@ export default Vue.extend({
   .semester-modal {
     display: none; /* Hidden by default */
     position: fixed; /* Stay in place */
-    z-index: 1; /* Sit on top */
+    z-index: 2; /* Sit on top */
     left: 0;
     top: 0;
     width: 100%; /* Full width */


tcho6319 (approved):
LGTM. Thanks Will.

> It's not a good sign that after all the z-index testing this modal still slipped through... there were also a ton of other random z-index modal styles I found throughout the codebase when trying to figure out which one would fix this issue. Should I make a notion item to try and re-factor modals more to remove outdated and duplicated code @tcho6319 ?

Yes, that would be great! I think refactoring modals will be a major task post launch.

PR #439, opened 2021-04-04
Updated CivE, chemE, and BE majors

einc:
### Summary 

Title.

BE has one 'self-check' requirement because I need to find more info on it post launch.

### Test Plan 

Test to see that CivE, chemE, and BE major doesn't have any breaking changes!


tcho6319 (approved):
LGTM

PR #440, opened 2021-04-04
Update current minors

einc:
### Summary 

Title.
For some requirements I had to exclude courses that fulfilled a previous minor requirement to prevent double counting within minor reqs.

### Test Plan 

No breaking changes for any of the minors


tcho6319 (approved):
LGTM

PR #443, opened 2021-04-04
Add graduation year

jason-tung:
### Summary 

Adds graduation field to onboarding.

https://www.notion.so/courseplan/Add-Graduation-Year-6b8a28de1ca14eb3b08b3fc8fe00cf03

### Test Plan 

Playtest on staging make sure nothing is broken.

Tests for Will's changes
1. Delete your onboarding data and ensure there is no graduation year by default (so for new users). Then confirm you cannot submit onboarding without inputting it.
2. Delete just the "gradYear" field from your onboarding (to simulate existing users) and ensure graduation year is still not selected by default.
3. Note that only required courses have the * mark on them (major asterisks removed)
4. Ensure the grad year dropdown autoscrolls to 2021

### Will's Notes
1. Because the onboarding modal can be clicked outside for existing users, we have no way to force users who have used our beta to input their graduation years. If this is important, we should force them to do so if the field is empty with some new modal in the future. Is this something I should make an item for to explore?
2. ~~The autoscroll, implemented the same as the add-semester modal, does not work as well for the onboarding modal because it is vertically larger than a screen, so it also scrolls the whole modal and not just the dropdown. If this is not visually pleasing, I could either try to change the scroll to be `smooth` instead of `auto` or remove it entirely.~~ FIXED

### Breaking Changes 

Adds new `'gradYear' : string` field to firestore and vuex onboarding data types.

### DIFF
--- a/src/components/Modals/Onboarding/Onboarding.vue
+++ b/src/components/Modals/Onboarding/Onboarding.vue
@@ -121,7 +121,8 @@ export default Vue.extend({
       if (
         this.name.firstName === '' ||
         this.name.lastName === '' ||
-        this.onboarding.college === ''
+        this.onboarding.college === '' ||
+        this.onboarding.gradYear === ''
       ) {
         this.isError = true;
       } else {
--- a/src/components/Modals/Onboarding/Onboarding.vue
+++ b/src/components/Modals/Onboarding/Onboarding.vue
@@ -132,6 +132,7 @@ export default Vue.extend({
             lastName: this.name.lastName,
           })
           .set(onboardingDataCollection.doc(store.state.currentFirebaseUser.email), {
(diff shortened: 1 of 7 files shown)


tcho6319 (approved):
LGTM! Thanks @jason-tung and @willespencer 

> Because the onboarding modal can be clicked outside for existing users, we have no way to force users who have used our beta to input their graduation years. If this is important, we should force them to do so if the field is empty with some new modal in the future. Is this something I should make an item for to explore?

@willespencer yes this was a concern that I had as well. Maybe we can create a modal that appears when a user first logs in if the gradYear is undefined. One of the buttons can open up Onboarding. But yeah a Notion item to document this exploration would be great. ty

PR #446, opened 2021-04-04
Add script to fix unique ID

SamChou19815:
### Summary 

Due to some past bugs I don't know, prod data has courses with duplicate unique ID. I added a script to fix that.

No need to merge this PR. If the script is good we can just run it against prod db. Once prod db is cleaned up, we can just close this PR.

### Test Plan 

I already run it against the dev db and having no issues. I also provided a dry-run option just in case.

### DIFF
--- a/functions/fix-unique-id.js
+++ b/functions/fix-unique-id.js
@@ -1,0 +1,53 @@
+const admin = require('firebase-admin');
+const fs = require('fs');
+const serviceAccount = require('./service-account.json');
+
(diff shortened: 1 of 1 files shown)


tcho6319 (approved):
LGTM

SamChou19815:
@tcho6319 You can just check out this PR and run it against prod. After that I can close it.

SamChou19815:
Seems to be working now. So no need for the script

PR #448, opened 2021-04-04
Release Fixes (Duplicate semesters, broken course menu)

willespencer:
### Summary 

This pull request fixes release issues found after launching the site but before removing the final whitelist.

- [x] Prevents duplicate semesters by resetting sem and year when dropdowns are reset for the new semester modal
- [x] Fixes issue where old courses that do not have credit ranges defined do not prevent the edit color/delete course menu from opening

image

Remaining TODOs:

- [ ] Slacked issues

### Test Plan 

1. Confirm duplicate semesters cannot be created like they can on courseplan.io: Delete all of your semesters, than add a semester of any season/year besides Spring 2021, then open the add semester modal and click add immediately. This fix should make it Spring 2021 instead of a duplicate sem.
2. Confirm these dropdowns still work and reset when modals are clicked outside of and submitted for the other places it is used in the code (edit semester and add course from self check).
3. Confirm that for a course like INFO 4998, when added to your plan, credits can still be edited.
4. Now delete the "creditRange" field from that course on the dev firebase and reload the page. Confirm that credits cannot be edited but colors can change and the course can be deleted

### Notes
- Should I make an item for solving the credit issue with a script long term?

### DIFF
--- a/src/components/Course/Course.vue
+++ b/src/components/Course/Course.vue
@@ -26,7 +26,7 @@
       @delete-course="deleteCourse"
       @color-course="colorCourse"
       @edit-course-credit="editCourseCredit"
-      :getCreditRange="getCreditRange"
+      :getCreditRange="getCreditRange || []"
       v-click-outside="closeMenuIfOpen"
     />
   </div>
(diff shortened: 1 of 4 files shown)


SamChou19815 (approved):

tcho6319 (approved):
LGTM! Thanks Will.

> Should I make an item for solving the credit issue with a script long term?

Yes, can you make Notion items for the credit issue and changing Arts & Science?

PR #454, opened 2021-04-10
Add linguistics major and minor

willespencer:
### Summary 

This pull request adds the linguistics major and minor to CoursePlan

Major
image

Minor
image

### Test Plan 

1. Confirm the checkers match the information here: https://linguistics.cornell.edu/undergraduate
2. More specifically, add the core classes and confirm they cannot fulfill other linguistics requirements
3. Confirm that additional courses logic holds up (split similarly to liberal studies, has a double countable credit req)
4. Confirm any course can be added to the ancillary skills courses (outside the core courses).
5. Confirm these requirements work for the minor as well.
5. Check that all the text makes sense and is typo-free, and that the urls link where they should

### Notes 

Ran into a few issues when creating this major and minor, and was wondering if we currently have any way of handling these.
- Is there any way we can restrict courses to only display ones that are 3 (or more) credits? Right now I have an 18 credit checker for the additional courses, which is not technically correct, as 2 credit courses could be used to fulfill this requirement when they can't in reality.
- Do we have any way to check a course does not have the "CU-UGR" designation? It shows up in student center under "Course Attribute": https://classes.cornell.edu/browse/roster/FA20/class/LING/4491
- Is there any way to check the restriction that "language courses" cannot be used for additional courses? I don't think so because LING courses can be crosslisted with language course codes and not count as a "language course," but still unsure.
- Right now, one course can be used to fulfill the 3000, 2000, and 1000 course for the minor  but not the major, I think because of the way double counting works (even though none of the requirements are set to allow double counting). Should I rework this requirement for the minor and make it different than the major?


src/requirements/data/majors/ling.ts
```diff
@@ -0,0 +1,95 @@
...
+    checker: [
+      (course: Course): boolean =>
+        ifCodeMatch(course.subject, 'LING') &&
+        !courseMatchesCodeOptions(course, ['LING 1101', 'LING 3302', 'LING 3303', 'LING 3314']) &&
+        !courseIsFWS(course),
+    ],
+    allowCourseDoubleCounting: true,
+    fulfilledBy: 'credits',
+    perSlotMinCount: [18],
+  },
```
tcho6319, line 78: > Is there any way we can restrict courses to only display ones that are 3 (or more) credits? Right now I have an 18 credit checker for the additional courses, which is not technically correct, as 2 credit courses could be used to fulfill this requirement when they can't in reality.

I think you could check the credits by looking at `unitsMinimum` or `unitsMaximum`? Checking `unitsMaximum` may be better than checking `unitsMinimum`. A helper function that checks that maximum credits a course could possibly be would be nice to reuse and have in `src/requirements/data/checkers-common.ts`

https://github.com/cornell-dti/course-plan/blob/c002a9ac0585d1aad76a6bc024a9038b2eb25877/src/requirements/types.ts#L1-L18

src/requirements/data/minors/ling.ts
```diff
@@ -0,0 +1,85 @@
...
+    checker: [
+      (course: Course): boolean =>
+        ifCodeMatch(course.subject, 'LING') &&
+        !courseMatchesCodeOptions(course, ['LING 1101', 'LING 3302', 'LING 3303', 'LING 3314']) &&
+        !courseIsFWS(course),
+    ],
+    allowCourseDoubleCounting: true,
+    fulfilledBy: 'credits',
+    perSlotMinCount: [10],
+  },
```
tcho6319, line 82: The helper function that checks the credits of a course could be helpful here too!

tcho6319 (commented):
Thanks Will! Left some comments to answer your questions.

> Do we have any way to check a course does not have the "CU-UGR" designation? It shows up in student center under "Course Attribute": https://classes.cornell.edu/browse/roster/FA20/class/LING/4491

No. Currently, our Course type doesn't include the `catalogAttribute` field:

https://github.com/cornell-dti/course-plan/blob/c002a9ac0585d1aad76a6bc024a9038b2eb25877/src/requirements/types.ts#L1-L18

Screen Shot 2021-04-11 at 4 20 04 PM
(screenshot is from cornell course roster api)

We could update the fetcher and src/requirements/types.ts so that we also collect this info from the course roster api and then check.


> Is there any way to check the restriction that "language courses" cannot be used for additional courses? I don't think so because LING courses can be crosslisted with language course codes and not count as a "language course," but still unsure.

Unsure since the FL doesn't seem to be an actual distribution category :(. Though I can't vet how accurate this would be, I think we'd have to manually check, similarly to how @einc listed all the foreign language subjects (e.g. KOREA, FREN, etc.). If you did opt for this approach, can you factor out the code so we don't duplicate?

https://github.com/cornell-dti/course-plan/blob/c002a9ac0585d1aad76a6bc024a9038b2eb25877/src/requirements/data/colleges/en.ts#L27-L58

> Right now, one course can be used to fulfill the 3000, 2000, and 1000 course for the minor but not the major, I think because of the way double counting works (even though none of the requirements are set to allow double counting). Should I rework this requirement for the minor and make it different than the major?

Yes that's strange. It could be because minor reqs are automatically double countable. I think you'll have to rework the reqs for the minor.

tcho6319 (approved):
LGTM thanks Will!

PR #455, opened 2021-04-11
Add Human Development major & minor

willespencer:
### Summary 

Add the Human Development major and minor to CoursePlan

Major
image

Minor
image

### Test Plan 

1. Test the major fully by looking at the 2020 requirements [here](https://www.human.cornell.edu/sites/default/files/Academics/Registrar/Curriculum%20sheets/2020-2021/HD%20Curriculum%20Sheet%202020-2021.pdf)
2. Test the minor fully by looking at the requirements [here](https://www.human.cornell.edu/hd/academics/undergraduate/minor)
3. Confirm all spelling, links, etc. look as expected.

### Notes 

Have a few requirements questions based on adding this data
- Having a problem with the minor, I think because of minor double counting. According to my roommate, you can count introductory courses towards the electives if you take more than one so I can't restrict them from the course elective checker, but it does mean that right now a course that someone uses to fulfill the introductory course will also count as an elective course despite both not being double countable. Is there a way to resolve this?
- 2 of the requirements specifically mentioned that AP biology can be used to fulfill it, and another specifically mentioned AP statistics cannot be used to fulfill it. Do we have/want to have any way of handling this, and should I document this somewhere?
- One requirement (that is the same as DEA) specifically says "Language courses may count here," but the existing DEA checker for this just checked the 'FL' category. However, FL courses don't actually show up right? Should we update these checkers to use the list of language course categories?


src/requirements/data/majors/hd.ts
```diff
@@ -0,0 +1,263 @@
...
+      'BTRY 3010',
+      'ILRST 2100',
+      'STSCI 2100',
+      'MATH 1710',
+      'PAM 2100',
+      'PSYCH 2500'
+    ),
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+  },
```
tcho6319, line 213: maybe for now, you could add `"disallowTransferCredit": true`.

src/requirements/data/majors/hd.ts
```diff
@@ -0,0 +1,263 @@
...
+      'https://www.human.cornell.edu/sites/default/files/Academics/Registrar/Curriculum%20sheets/2020-2021/HD%20Curriculum%20Sheet%202020-2021.pdf',
+    checker: [
+      (course: Course): boolean =>
+        ['PBS', 'BIOLS-AG', 'BIONLS-AG'].some(
+          distribution => course.catalogDistr?.includes(distribution) ?? false
+        ),
+    ],
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+  },
```
tcho6319, line 238: we could add `"disallowTransferCredit": true`.

src/requirements/data/majors/hd.ts
```diff
@@ -0,0 +1,263 @@
...
+  {
+    name: 'Additional Requirements',
+    description:
+      'Any course with the Course Distribution PBS, BIOLS-AG, BIONLS-AG, SBA, KCM, MQR, LA, CA, or HA. Language courses may count here.',
+    source:
+      'https://www.human.cornell.edu/sites/default/files/Academics/Registrar/Curriculum%20sheets/2020-2021/HD%20Curriculum%20Sheet%202020-2021.pdf',
+    checker: [
+      (course: Course): boolean =>
+        ['PBS', 'BIOLS-AG', 'BIONLS-AG', 'SBA', 'KCM', 'MQR', 'LA', 'CA', 'HA', 'FL'].some(
```
tcho6319, line 247: > One requirement (that is the same as DEA) specifically says "Language courses may count here," but the existing DEA checker for this just checked the 'FL' category. However, FL courses don't actually show up right? Should we update these checkers to use the list of language course categories?

Yes, we should use the list of language course categories. See my comment in #454

tcho6319 (changes requested):
Thanks Will! Requested some changes and left some comments.

> Having a problem with the minor, I think because of minor double counting. According to my roommate, you can count introductory courses towards the electives if you take more than one so I can't restrict them from the course elective checker, but it does mean that right now a course that someone uses to fulfill the introductory course will also count as an elective course despite both not being double countable. Is there a way to resolve this?

Not a fix that's within the scope of this PR. We see a similar problem with the CS minor's 3410/3110 and Four 3000+ level requirements. @einc, @SamChou19815, and I discussed how implementing compound requirements will remove how we're currently abusing the `allowCourseDoubleCounting` flag (i.e. how the Liberal Studies reqs are double countable) and then we can hopefully make these types of sub requirements (like the CS minor example and the issue you brought up) not double countable. 

> 2 of the requirements specifically mentioned that AP biology can be used to fulfill it, and another specifically mentioned AP statistics cannot be used to fulfill it. Do we have/want to have any way of handling this, and should I document this somewhere?

For now, you could add the `"disallowTransferCredit": true` flag, as I've indicated in my comments. We're going to eventually give AP/IB exams unique crseId values and have checkers explicitly include them if a certain exam could apply to the req. So we could probably handle these specific cases better a little later on.

tcho6319 (approved):
LGTM! Ty Will :) Make sure you ran the tests and generated the decorated requirements json with your latest changes and then feel free to squash & merge!

tcho6319 (approved):

PR #456, opened 2021-04-11
[easy] Remove simplified course json generator

SamChou19815:
### Summary 

We are always using the full json now, so let's remove this dead code.

### Test Plan 

Note that `src/assets/courses/courses.json` no longer exists in the repo, which shows that the generator code is indeed dead.


tcho6319 (approved):
LGTM ty Sam!

PR #459, opened 2021-04-14
Post launch identified bugs (Total credit denominator, edit semester dropdowns)

willespencer:
### Summary 

This pull request fixes two bugs identified by our users after the launch. [120 credits](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=01b596c8bca94d3cba14015f85ea0c65), [dropdown](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=9e3425f540cc4cdf83f33281f266501c)

- [x] Stop the `minCountRequired` from incrementing. (We seem to have identified the bug incorrectly, it was not courses with customizable credits that was the issue, but duplicate courses in general, which tend to be things with editable credits like INFO 4998).
- [x] Reset edit semester data when the modal is clicked outside, like new semester.

image

### Test Plan 

1. Add duplicate courses to your plan (some with editable credits like INFO 4998 and some without) and confirm the "120" for Total Academic Credits never changes
2.  Change the year or semester on the edit semester modal, close the modal without submitting, open it up, and submit. If the semester is unchanged (aka uses the dropdown defaults) then the bug is fixed.
3. Confirm edit semester, add semester, and add course from self check dropdown still works as expected.

### DIFF
--- a/src/requirements/requirement-frontend-computation.ts
+++ b/src/requirements/requirement-frontend-computation.ts
@@ -106,10 +106,7 @@ const getTotalCreditsFulfillmentStatistics = (
 
   let minCountFulfilled = 0;
   const minCountRequired = 120;
-  const eligibleCourses =
-    college === 'AG'
-      ? courses.filter(course => !course.code.startsWith('PE '))
-      : courses.filter(courseIsAllEligible);
+  const eligibleCourses = courses.filter(courseIsAllEligible);
 
   eligibleCourses.forEach(course => {
     minCountFulfilled += course.credits;
--- a/src/requirements/requirement-frontend-computation.ts
+++ b/src/requirements/requirement-frontend-computation.ts
@@ -105,8 +105,7 @@ const getTotalCreditsFulfillmentStatistics = (
   }
 
   let minCountFulfilled = 0;
-  let minCountRequired = 120;
-  const courseCodeSet = new Set<string>();
+  const minCountRequired = 120;
   const eligibleCourses =
     college === 'AG'
       ? courses.filter(course => !course.code.startsWith('PE '))
@@ -114,11 +113,6 @@ const getTotalCreditsFulfillmentStatistics = (
 
   eligibleCourses.forEach(course => {
     minCountFulfilled += course.credits;
-    if (courseCodeSet.has(course.code)) {
-      minCountRequired += course.credits;
-    } else {
-      courseCodeSet.add(course.code);
-    }
   });
 
   return {
--- a/src/requirements/requirement-frontend-computation.ts
+++ b/src/requirements/requirement-frontend-computation.ts
@@ -44,8 +44,8 @@ const getTotalCreditsFulfillmentStatistics = (
         id: 'College-AG-total-credits',
         description:
           '120 academic credits are required for graduation. ' +
-          'A minimum of 100 credits must be in courses for which a letter grade was recieved. ' +
-          'PE courses do not count.',
+          'PE courses and courses numbered 1000-1099 do not count towards the 120 credits. ' +
+          'Repeated courses may not apply to this requirement, but we do not check this.',
         source: 'http://courses.cornell.edu/content.php?catoid=41&navoid=11561',
       };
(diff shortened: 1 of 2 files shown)


SamChou19815 (approved):

src/requirements/requirement-frontend-computation.ts
```diff
@@ -105,20 +105,14 @@ const getTotalCreditsFulfillmentStatistics = (
   }
 
   let minCountFulfilled = 0;
-  let minCountRequired = 120;
-  const courseCodeSet = new Set<string>();
+  const minCountRequired = 120;
   const eligibleCourses =
     college === 'AG'
       ? courses.filter(course => !course.code.startsWith('PE '))
       : courses.filter(courseIsAllEligible);
```
tcho6319, line 112: I don't think we need to special case CALS here and we can just have `const eligibleCourses = courses.filter(courseIsAllEligible`  since it seems that courses that are 10** level or PE shouldn't count. 

Screenshot is of the [CALS graduation reqs](https://cals.cornell.edu/undergraduate-students/student-services/degree-requirements/graduation-requirements#:~:text=Minimum%20total%20credits%3A%20120%20academic,required%20for%20full%2Dtime%20status.)
Screen Shot 2021-04-14 at 3 41 11 PM

src/requirements/requirement-frontend-computation.ts
```diff
@@ -105,20 +105,14 @@ const getTotalCreditsFulfillmentStatistics = (
...
 
   eligibleCourses.forEach(course => {
     minCountFulfilled += course.credits;
-    if (courseCodeSet.has(course.code)) {
-      minCountRequired += course.credits;
-    } else {
-      courseCodeSet.add(course.code);
-    }
```
tcho6319, line 121: By removing this code, we do get rid of the issue where the 120 denominator does not increase for repeating courses. However, this means that repeated courses that aren't project teams or research (e.g. having CS 1110 more than once in your plan) would incorrectly get counted towards fulfilling the 120. For example, if I had CS 1110 in my plan twice, my total academic credits fulfillment would be 8/120 instead of 4/120.

@einc, @willespencer should we filter `eligibleCourses` further so that we only increase `minCountFulfilled` if it is a (dupe) project team/research course or if it's the first time we see the course (e.g. counting for the first CS 1110 we see)? Note this could get hairy for non research/project team courses like INFO 4940 where different sections are different courses. The question is how much inaccuracy are we comfortable with? Do we want to overcount/undercount?

tcho6319 (changes requested):
Thanks Will! I left a suggestion and brought up a discussion.

src/requirements/requirement-frontend-computation.ts
willespencer: Yeah I'm leaning toward leaving this as is so that we overcount? I was specifically looking at the [Arts Requirement](https://courses.cornell.edu/content.php?catoid=41&navoid=11570#credit-req) which states specifically 

"Students occasionally need to repeat courses. Some courses, such as independent study, some music and performance courses, and specific topical seminars, in which content is significantly different, do grant credit when the course is taken more than once. For all repeated courses, both grades appear on the transcript and are included in both the term and cumulative GPA.  For repeated courses that do not grant credit more than once, only one instance counts toward degree credits and requirements."

This seems to complicated for us to restrict, and I think we'd be likely to disallow something that does count. Let me know if you think otherwise though!

tcho6319: I would be ok with leaving as is and overcounting. In that case, I think we should add a sentence in the description (or a warning?) that says something like: "Repeated courses may not apply to the 120 Academic Credits requirement.". Just so that users are aware of the fact that we don't handle this.

tcho6319 (approved):
LGTM! Thanks Will

PR #460, opened 2021-04-14
Add and display slot names to every requirement

SamChou19815:
### Summary 

This PR adds a slot name to every requirement, and modified the frontend to display them.
The frontend display logic is as follows:

```
if requirement fulfilled by credit {
  maintain the old behavior of 'Course 1', 'Course 2' ...
  // that's why requirement fulfills by credit doesn't have to provide slot name
}
if requirement fulfilled by course {
  for each slot {
    if the minCount == 1 {
      name is just slotName from requirement data
      // e.g. if the slot minCount == 1, and slotName == "programming", then name is "programming"
    } else {
      generated names are `${slotName from requirement data} {1, 2, 3, ...}`
      // e.g. if slotName == "programming", then names are "programming 1", "programming 2", ...
    }
  }
}
```

Therefore, it will mostly replicate the old logic with names that make sense. For courses with only one slot, I just give it the generic name "Course", since it doesn't make much sense to give it a more specialized name.

### Test Plan 

**Credit requirement**
credits-requirement-slot

**Dummy course name**
dummy-course-name

**A named slot**
named-slot


tcho6319 (approved):
I left a small nit, but LGTM Sam! Thank you so much :) To confirm: only requirements fulfilled by courses should have the `slotNames` field right?

SamChou19815:
> To confirm: only requirements fulfilled by courses should have the `slotNames` field right?

Yes.

PR #461, opened 2021-04-15
Add Catalog Attribute course data

willespencer:
### Summary 

This pull request adds the catalog attribute field from the course roster to our course data to use in checkers

- [x] Add catalogAttribute field and re-run the course data fetcher
- [x] Re-work the linguistics major to support the restriction that only 1 additional course can have the CU-UGR attribute (and be 4 credits)

image

### Test Plan 

1. Complete the first additional course requirement with a CU-UGR course in linguistics (like LING 3390)
2. Try and fail to complete the second additional course requirement with a CU-UGR course in linguistics (like LING 4491)

### Notes 

This requirement is getting pretty complicated and would definitely benefit from compounded requirements. I had to make some weird choices like the following...

1. The CU-UGR course right now must be 3000+. This is not specified on the linguistics page, but I think it is ok because the only CU-UGR courses in linguistics happen to be 3000+. Source: https://classes.cornell.edu/search/roster/FA20?q=&subjects%5B%5D=LING&days-type=any&crseAttrs-type=any&crseAttrs%5B%5D=CU-UGR&breadthDistr-type=any&pi=
2. Because I split the 3000+ requirement, the names and descriptions are a bit long and wordy. Any suggestions to make them better, or do you think they are clear?


SamChou19815 (approved):
LGTM, just need to merge in my slot names changes.

tcho6319 (approved):
LGTM! Make sure you ran the test and everything checks out after merging in Sam's slotNames. Just checking, but do the linguistics minor requirements need to be updated like the major reqs regarding the check for the CU-UGR catalog attribute?

willespencer:
> Just checking, but do the linguistics minor requirements need to be updated like the major reqs regarding the check for the CU-UGR catalog attribute?

I don't think so! There is no mention of UGR [here](https://linguistics.cornell.edu/undergraduate) is for the major. Thanks for asking though!

PR #464, opened 2021-04-17
Add Material Science Engineering and PAM majors

willespencer:
### Summary 

This pull request adds the Material Science Engineering and PAM majors to CoursePlan

- [x] Added material science major
- [x] Added PAM major
- [x] Fixed a credit bug in the human ecology college (3 -> 9 credits)
- [x] Added TODO to ling major I missed, fixed a description in human development major

image

image

### Remaining TODOs

Put TODOs in the files where there were requirements we could not handle. Please confirm that these are all currently things that we cannot check.

### Test Plan 

1. Try to fulfill the entire Material Science Engineering Major and confirm the requirements match (especially the toggleable req and category req): https://www.mse.cornell.edu/mse/programs/undergraduate-programs/major/major-curriculum
2. Do the same for PAM here: https://www.human.cornell.edu/sites/default/files/Academics/Registrar/Curriculum%20sheets/2020-2021/PAM%20Curriculum%20Sheet%202020-2021.pdf
3. Confirm the credit bug in human ecology for the credits outside of major requirement is fixed.

### Notes 

Just had a few things to specifically mention
1. For the 2 categories requirement for MSE, I based the code off of the liberal studies code for engineering. However, I struggled with the ... operator and understanding what exactly the code was doing, so let me know if you have any suggestions on how to write that checker better!
2. The chart for the 'Natural Science 1' requirement in PAM is a little confusing. Do you think this means any course in this entire chart can count (like I coded)?

image

3. Been noticing a lot of the human development requirements are similar across human ecology majors (Natural Science I in human development is a subset of the one in PAM, Natural Science II, Additional Requirements, Electives is exactly the same as human development, etc.). Might be worth factoring these out in some way or adding common checkers?


src/requirements/data/majors/pam.ts
```diff
@@ -0,0 +1,226 @@
...
+    source:
+      'https://www.human.cornell.edu/sites/default/files/Academics/Registrar/Curriculum%20sheets/2020-2021/PAM%20Curriculum%20Sheet%202020-2021.pdf',
+    checker: includesWithSingleRequirement('ECON 1110'),
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+    slotNames: ['Course'],
+  },
+  {
+    name: 'Introduction to Sociology',
+    description: 'SOC 1101, DSOC 1101, or PAM 2250.',
```
tcho6319, line 81: Can you add this sentence to the description: "Note: PAM 2250 cannot count for both the Introduction to Sociology requirement AND the Additional PAM Electives requirement."

src/requirements/data/majors/pam.ts
```diff
@@ -0,0 +1,226 @@
...
+    description:
+      'Any 3-4 credit course with a Course Distribution PBS, BIOLS-AG, or BIONLS-AG. Course must be taken at Cornell (no AP credit allowed).',
+    source:
+      'https://www.human.cornell.edu/sites/default/files/Academics/Registrar/Curriculum%20sheets/2020-2021/PAM%20Curriculum%20Sheet%202020-2021.pdf',
+    checker: [
+      (course: Course): boolean =>
+        ['PBS', 'BIOLS-AG', 'BIONLS-AG'].some(
+          distribution => course.catalogDistr?.includes(distribution) ?? false
+        ),
+    ],
```
tcho6319, line 162: Can you also check that the credit min is 3?

tcho6319 (changes requested):
Thanks Will! I requested some smaller changes, but once those are addressed, this PR should be g2g.

> The chart for the 'Natural Science 1' requirement in PAM is a little confusing. Do you think this means any course in this entire chart can count (like I coded)?

Yes I think you're right!

> Been noticing a lot of the human development requirements are similar across human ecology majors (Natural Science I in human development is a subset of the one in PAM, Natural Science II, Additional Requirements, Electives is exactly the same as human development, etc.). Might be worth factoring these out in some way or adding common checkers?

Good eye! Can you make a Notion item (for week 4.26-5.2) and maybe assign to @einc for now since she'll be adding in more majors/minors later on?

src/requirements/data/majors/mse.ts
```diff
@@ -0,0 +1,305 @@
...
+      'https://www.mse.cornell.edu/mse/programs/undergraduate-programs/major/major-curriculum/electives',
+    checker: [
+      (course: Course): boolean =>
+        applicationElectives[0].includes(`${course.subject} ${course.catalogNbr}`) ?? false,
+      ...applicationElectives
+        .slice(1)
+        .map(categoryCourses => (course: Course): boolean =>
+          categoryCourses?.includes(`${course.subject} ${course.catalogNbr}`) ?? false
+        ),
+    ],
```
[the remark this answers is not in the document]
willespencer: Oh you're right, thanks for this! I was leaving the original `(course: Course): boolean =>` at the start of the checker which was what was not working, but I see now with your explanation that this operator produces the list of checkers so you wouldn't want that.

src/requirements/data/majors/pam.ts
willespencer: I didn't add this because neither requirement is double countable, so CoursePlan would not let users count PAM 2250 for both even if users tried. Do you still think I should add it?

tcho6319 (approved):
LGTM! Thanks Will :)

PR #465, opened 2021-04-17
Add physics and engineering physics majors

benjamin-shen:
### Summary 

This pull request adds the physics and engineering physics majors to CP.

- [x] added PHYS major according to https://courses.cornell.edu/preview_program.php?catoid=41&poid=19990
- [x] added EP major according to https://courses.cornell.edu/preview_program.php?catoid=41&poid=19775

Remaining TODOs:

- [ ] implement as compound requirements: PHYS Concentration, EP Major-approved Elective(s)

### Test Plan 

Change your major to phys (in A&S) and/or ep (in ENG) and confirm that the requirements are correct and as complete as possible.

### Notes 

1. I'm not sure why the physics core is inconsistent across these two sources: [1](https://physics.cornell.edu/physics-majors) and [2](https://courses.cornell.edu/preview_program.php?catoid=41&poid=19990). It looks like there are new courses (maybe required for newer students?). But I didn't find more information on when the different requirements took effect, so I omitted them altogether.
2. The following are not checked (also noted as comments):
    - "Students who do not take PHYS 1116 must also complete PHYS 2216." (not sure what this means)
    - "Inside concentrators should complete at least two semesters of advanced mathematics classes at the 3000+ level such as the AEP 3200 & AEP 4200 sequence, or other relevant classes." (can probably do with compound requirements)
3. The user must override in order to apply the following:
    - "Other courses approved by the director of undergraduate studies" and "If this activity is done as an independent project, PHYS 4490, up to 4 credits can be applied toward the concentration" to PHYS Concentration "inside" Physics
    - Any course to PHYS Concentration "outside" Physics
    - Any non-upper-level course to EP Major-approved Elective(s)


src/requirements/data/index.ts
```diff
@@ -160,6 +162,11 @@ const json: RequirementsJson = {
       schools: ['EN'],
       requirements: envEngineeringRequirements,
     },
+    EP: {
+      name: 'Engineering Physics',
+      schools: ['EN'],
+      requirements: epRequirements,
+    },
```
tcho6319, line 169: Do you mind renaming to AEP? I believe that's the more common name. The name would be 'Applied and Engineering Physics'. Instead of `epRequirements` for the file name, can we do `aepRequirements`?

src/requirements/data/majors/phys.ts
```diff
@@ -0,0 +1,121 @@
...
+        ],
+        perSlotMinCount: [15],
+      },
+      'Concentration "outside" Physics': {
+        description:
+          'The concentration outside physics ("outside concentration") provides more flexibility ' +
+          'for those want to develop skills in physics but whose career interests lie elsewhere.',
+        counting: 'credits',
+        // This is essentially self-check; courses must be manually overwritten by the user.
+        checker: [(): boolean => false],
```
tcho6319, line 114: You should make `checker: [(): boolean => true],` so that a course could be applied to it. Currently, I can't add any course

tcho6319 (changes requested):
Thanks Ben! I suggested a naming change and another change regarding the checker for the outside concentration, but once those comments are addressed, this PR should be g2g.

src/requirements/data/index.ts
benjamin-shen: I believe AEP is the department name, and EP is the undergraduate major.
image

src/requirements/data/majors/phys.ts
benjamin-shen: Then any course would be able to fulfill the requirement, right? I thought we wanted requirements to be as limiting as possible, so we wouldn't accidentally be telling the user that they finished a requirement when they really didn't.

My idea was to make no courses apply to it, and the user would have to override to apply anything to it. If any course fulfilled the requirement, then overriding wouldn't be useful here. I originally thought of explicitly classifying this as self-check, but didn't because we currently don't support self-check within toggleable requirements.

src/requirements/data/index.ts
tcho6319: Ah ok sounds good. I'm ok with leaving it as is then!

src/requirements/data/majors/phys.ts
tcho6319: Yeah I see your point. I wonder if we can update the checker so that it doesn't include courses from PHYS. Would that be a safe assumption to make for the outside concentration? Since we have a `checkerWarning` for this requirement, we're implicitly treating this concentration as having as weak but accurate of a checker as possible so that the user can essentially use whatever course to fulfill it. @einc what are your thoughts?

tcho6319: Yeah I see your point. I wonder if we can update the checker so that it doesn't include courses from PHYS. Would that be a safe assumption to make for the outside concentration? Since we have a `checkerWarning` for this requirement, we're implicitly treating this concentration as having as weak but accurate of a checker as possible so that the user can essentially use whatever course to fulfill it. @einc what are your thoughts?

einc: I think having any courses outside Physics would be good for now since yea we have the checkerWarning!

tcho6319 (approved):
LGTM! Thanks Ben

PR #470, opened 2021-04-19
Annotate emits on components so that they can be type-checked

SamChou19815:
### Summary 

Following the instructions here: https://v3.vuejs.org/guide/typescript-support.html#annotating-emits

### Test Plan 

Mostly type-only changes, and the CI still passes. Also played with the app to ensure there are no warnings.

### DIFF
--- a/src/components/Requirements/SubRequirement.vue
+++ b/src/components/Requirements/SubRequirement.vue
@@ -218,6 +218,18 @@ export default defineComponent({
     color: { type: String, required: true },
     tourStep: { type: Number, required: true },
   },
+  emits: {
+    'modal-open': (open: boolean) => typeof open === 'boolean',
+    changeToggleableRequirementChoice(id: string, option: string) {
+      return typeof id === 'string' && typeof option === 'string';
+    },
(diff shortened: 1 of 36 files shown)


src/components/Requirements/SubRequirement.vue
```diff
@@ -218,6 +218,18 @@ export default defineComponent({
     color: { type: String, required: true },
     tourStep: { type: Number, required: true },
   },
+  emits: {
+    'modal-open': (open: boolean) => typeof open === 'boolean',
+    changeToggleableRequirementChoice(id: string, option: string) {
```
tcho6319, line 223: why do we have `'modal-open'` as a string and `changeToggleableRequirementChoice(...)`?

tcho6319 (approved):
LGTM! Thanks Sam :) I left a question to clarify my own understanding.

src/components/Requirements/SubRequirement.vue
SamChou19815: They have the same runtime effect, the only difference is that `'modal-open'` without quote is not a legal JS identifier.

PR #473, opened 2021-04-21
Rename requirement components to reduce confusion

SamChou19815:
### Summary 

Since no one else is working on requirement stuff except me this week, I will do the renaming that I want to do for a while.

As discussed before subrequirement means two different things, depending on whether you are working on the algorithm or in the UI. This diff makes things more consistent. Since we finally have a consistent names for requirement stuff, I added them to the docs to reduce future confusion.

### Test Plan 

Site still runs.


tcho6319 (approved):
LGTM! Site seems to work as usual. Thanks Sam.

PR #474, opened 2021-04-21
Factor out requirement components in RequirementFulfillment (1/n)

SamChou19815:
### Summary 

Depends on #473.

Factor out some display component inside `RequirementFulfillment` to make it smaller and more manageable.

- The toggle part with progress now goes into `RequirementDisplayToggle`.
- The description + warnings now goes into `RequirementInformation`

### Test Plan 

The toggle and the basic information is still correctly displayed:

Screen Shot 2021-04-21 at 14 59 32
Screen Shot 2021-04-21 at 14 59 46

### DIFF
--- a/src/components/Requirements/RequirementFulfillment.vue
+++ b/src/components/Requirements/RequirementFulfillment.vue
@@ -1,63 +1,13 @@
 <template>
   <div class="subrequirement">
-    <button
-      @click="toggleDescription()"
-      class="dropdown row slightly-lower-opacity-on-hover"
-      aria-haspopup="true"
-      data-toggle="dropdown"
-    >
-      <div class="row depth-req">
-        <div class="btn">
-          <drop-down-arrow
-            :isFlipped="displayDescription"
-            :fillColor="getArrowColor()"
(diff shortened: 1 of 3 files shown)


src/components/Requirements/RequirementFulfillment.vue
```diff
@@ -1,63 +1,13 @@
 <template>
   <div class="subrequirement">
```
tcho6319, line 2: nit: should we rename this class because it's technically for the requirementfulfillment component now and subrequirements are the slots?

tcho6319 (approved):
Thanks Sam! Left a small nit but LGTM.

tcho6319:
Also, can you update your PR summary so that it includes the names of the components you factored out? Screenshots for each factored out component would be helpful!

src/components/Requirements/RequirementFulfillment.vue
SamChou19815: Turns out there is no css class for `subrequirement`, so i will just remove it

PR #479, opened 2021-04-28
Add more possible foreign langs in requirements

willespencer:
### Summary 

This pull request adds missing foreign languages codes to the list we have in requirements

image

### Test Plan 

1. Confirm that the list of course codes match the list here: https://lrc.cornell.edu/languages-cornell
2. Try to add a course with one of the new subjects to a requirement that uses the foreign language checker. For example, try to add an ASL course to fulfill the A&S foreign language requirement.

### Notes 

1. I was planning on just adding ASL as a small bug fix (because someone pointed out to me that they could not count their foreign language courses in ASL), but ended up consulting the above list and adding a bunch more. Please let me know if we don't want all of these added for some reason!
2. Some of the course subjects on that website are specifically marked with a hollow dot as not applying to the A&S Foreign Language requirement. Do we want to special case that in some way, or continue to be as open as possible?


SamChou19815 (approved):
LGTM

einc (approved):
Thank you Will! For Note #2, maybe we shouldn't add departments that don't count for the A&S FL to the list (maybe comment them out for now). We should probably be as strict as possible so we don't mislead users into thinking something can count. At least for being too strict, users can soon override.

tcho6319 (approved):
Test plan works. LGTM! Thanks Will. I agree with @einc; I think it would be better to be stricter especially if we'll allow requirement overriding.

tcho6319 (approved):
LGTM!

PR #481, opened 2021-04-29
Update Requirement or Slot Name Script

hahnbeelee:
### Summary

There are going to be 2 instances in the firestore where a requirement's name and slot name will appear:
1. user-onboarding-data
2. user-overridden-reqs (a new collection)
(read more about why [here](https://www.notion.so/courseplan/Discussion-Story-AP-IB-Infra-Update-to-handle-opt-out-and-requirement-overriding-ee2c3f234c6c44018531ed6ec34c56a1))

Whenever a dev changes a requirement or slot name, we must ensure that all instances of that name gets updated in the database too. This script was created as a tool for devs to run whenever a req or slot name gets updated.

[Related Notion page](https://www.notion.so/courseplan/Script-for-Changing-Requirement-or-Slot-Name-07894ee3f6cc4986ba84f1549b248714)

How we're envisioning it being used:

- Manually set these desired values in the code: `oldName`, `newName`, and `isChangingSlotName`

image

- Run the script

### Test Plan
I don't think you can actually test this out yet because this was pre-created for the new AP/IB opt in/out features and the collection doesn't exist in the firestore. So, just look over the code and make sure it logically makes sense.


functions/rename-requirement-and-slot-name.js
```diff
@@ -0,0 +1,109 @@
...
+/** updates the old name of the requirement or slot name with the new name in
+ * userOverriddenReqsCollection
+ */
+const updateNames = doc => {
+  const newDoc = {};
+  for (const [key, value] of Object.entries(doc)) {
+    const nameInDB = isChangingSlotName ? key[1] : key[0];
+    if (nameInDB === oldName) {
+      const newKey = isChangingSlotName ? [key[0], newName] : [newName, key[1]];
+      newDoc[newKey] = value;
```
SamChou19815, line 33: Not sure whether it will work. You are indexing a array with an array.

functions/rename-requirement-and-slot-name.js
```diff
@@ -0,0 +1,109 @@
...
+const fs = require('fs');
+const serviceAccount = require('./service-account.json');
+
+admin.initializeApp({
+  credential: admin.credential.cert(serviceAccount),
+  databaseURL: 'https://cornelldti-courseplan-dev.firebaseio.com',
+});
+
+const db = admin.firestore();
+const userOverriddenReqsCollection = db.collection('user-overridden-reqs');
```
tcho6319, line 13: To be consistent with the naming scheme (i.e. `'user-selectable-requirements'`), can we do `'user-overriden-requirements'`?

functions/rename-requirement-and-slot-name.js
tcho6319: @SamChou19815 I believe in this script we're assuming that the key for `newDoc` is an array `[requirement name, slot name]`, so the idea is to set the value of `newDoc` (an object) at `newKey`.

functions/rename-requirement-and-slot-name.js
```diff
@@ -0,0 +1,109 @@
...
+      const newKey = isChangingSlotName ? [key[0], newName] : [newName, key[1]];
+      newDoc[newKey] = value;
+    } else {
+      newDoc[key] = value;
+    }
+  }
+  return newDoc;
+};
+
+const updateNamesInArray = arr => {
```
tcho6319, line 41: Can you add documentation for the functions?

tcho6319 (changes requested):
Thanks Hahnbee! Requested some minor changes but overall this LGTM once they are addressed. I would note that this script would have to change a little bit if we change the structure of the user-overriden-requirements collection, but given the current assumption of the structure (where we're storing `[requirement name, slot name]` as an array) this looks good.

functions/rename-requirement-and-slot-name.js
SamChou19815: I don't think that will work on JS. You can only use string or number to index things.

tcho6319:
Also given our slack conversation, this script would need to change to handle the new structure for our data. See the updated Notion page [here](https://www.notion.so/courseplan/Discussion-Story-AP-IB-Infra-Update-to-handle-opt-out-and-requirement-overriding-ee2c3f234c6c44018531ed6ec34c56a1).

functions/rename-requirement-and-slot-name.js
```diff
@@ -17,43 +17,77 @@ const userOnboardingData = db.collection('user-onboarding-data');
...
+        : updateOnboardingSlotName(exam.optOut),
+    };
   });
 };
 
-const transformOverriddenReqsCollection = doc => updateNames(doc);
+const updateOnboardingSlotName = doc => {
+  const newDoc = {};
+  for (const [reqName, value] of Object.entries(doc)) {
+    if (value.includes(oldName)) {
```
hahnbeelee, line 76: I have a concern for if the same slot name exists and it isn't actually the one we want to update.

hahnbeelee: Do you think a case like this will exist / needs to be accounted for?

hahnbeelee: sounds good. ill make the change.

tcho6319 (approved):
LGTM! Once we start using the new collections for requirement overriding, this script should be tested with the dry run to see how it handles requirement name changes and/or slot name changes. I'm good with merging this in. Thanks Hahnbee!

PR #482, opened 2021-04-29
Fuse search implementation

hahnbeelee:
### Summary
This PR implements the search optimization library [Fuse.js](https://fusejs.io/) for our course search.

### Test Plan
Search some strings and inspect if the search results align to your expectations.
Take note of weird behavior

#### Improvements
- Searching without spaces works (ex. CS1110)
- Searching where you misspell things works

#### Drawbacks
- Feels slower

#### ???
The order that courses show up seem more intuitive at times and at other times it's questionable. 
For example, `Hist fws` doesn't load all the courses hist fws courses.

#### Questions
What does CUReviews do for their search? Can't we just copy them?


src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -25,35 +25,44 @@
...
 import { PropType, defineComponent } from 'vue';
+import Fuse from 'fuse.js';
 import { fullCoursesArray } from '@/assets/courses/typed-full-courses';
 
 const getMatchingCourses = (
   searchText: string,
   filter?: (course: CornellCourseRosterCourse) => boolean
 ): readonly CornellCourseRosterCourse[] => {
-  // search after value length of 2 to reduce search times of courses
-  if (!searchText || searchText.length < 2) return [];
```
SamChou19815, line 35: This needs to be added back to improve speed.

src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -25,35 +25,44 @@
...
-  // Sort both results by title
-  code.sort((first, second) => first.titleLong.localeCompare(second.titleLong));
-  title.sort((first, second) => first.titleLong.localeCompare(second.titleLong));
+  const courses: readonly SearchableCourse[] =
+    filter != null ? fullCoursesArray.filter(filter) : fullCoursesArray;
+  courses.map((course: SearchableCourse) => {
+    course.courseCode = `${course.subject} ${course.catalogNbr}`;
+    course.fullCourseString = `${course.subject} ${course.catalogNbr} ${course.titleLong}`;
+    return course;
+  });
```
SamChou19815, line 46: Do this computation on all courses only once.

src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -25,35 +25,44 @@
...
+  const options = {
+    keys: [
+      { name: 'courseCode', weight: 2 },
```
SamChou19815, line 49: Probably not needed since it's already covered by subject and catalogNbr?

SamChou19815 (commented):
Some more suggestions on optimization

src/components/Modals/NewCourse/CourseSelector.vue
tcho6319: Agreed. Could the same be also said for fullCourseString if we can derive it from subject, catalogNbr, and titleLong? Did you create a new field because there isn't another way of weighting them all together?

tcho6319 (changes requested):
Thanks Hahnbee! I agree with @SamChou19815's suggestions for optimization. I would experiment with the following cases (especially regarding the `fullCourseString` and `courseCode`) and see which works better:
1. Adding `courseCode` and `fullCourseString` for all courses beforehand.
2. Not having `courseCode` and `fullCourseString` as added fields and instead adjusting the weights of `subject`, `catalogNbr`, and `titleLong`. 

If after these explorations, the search time and results are improved, it would be great to have this included in our push to prod on May 4. We can improve on this again later with insights from the search functionality on CU Reviews.

src/components/Modals/NewCourse/CourseSelector.vue
hahnbeelee: It's funky. Without the fullCourseString it doesn't match substrings of the entry with each property. I added all these things to get it to act the way I want it to. But, after thinking about it, it might just be better to create an array of strings of only fullCourseString and strictly match to that. I think that's the most simple implementation

src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -80,7 +57,32 @@ export default defineComponent({
...
+      return result;
+    },
+    searchableCourses(): readonly SearchableCourse[] {
+      const courses: readonly SearchableCourse[] =
+        this.courseFilter != null ? fullCoursesArray.filter(this.courseFilter) : fullCoursesArray;
+      courses.map((course: SearchableCourse) => {
+        course.courseCode = `${course.subject} ${course.catalogNbr}`;
+        course.fullCourseString = `${course.subject} ${course.catalogNbr} ${course.titleLong}`;
+        return course;
+      });
```
tcho6319, line 84: Could we define a const like `fullSearchableCoursesArray` outside of computed props such that
```
fullSearchableCoursesArray = fullCoursesArray.map((course: SearchableCourse) => {
        course.courseCode = `${course.subject} ${course.catalogNbr}`;
        course.fullCourseString = `${course.subject} ${course.catalogNbr} ${course.titleLong}`;
        return course;
      });
```
and then the computed prop`searchableCourses()` would be `this.courseFilter != null ? fullSearchableCoursesArray.filter(this.courseFilter) : fullSearchableCoursesArray;`.

Then, we wouldn't have to add the `courseCode` and `fullCourseString` field every time we search for courses because it would precomputed. Hopefully this will speed things up a bit!

tcho6319 (changes requested):
Thanks Hahnbee! The search seems slower since the new changes. I made a suggestion that could hopefully speed things up! Let me know if that does/doesn't help.

tcho6319 (approved):
LGTM!

PR #484, opened 2021-04-30
Remove Semester Caution Modal

willespencer:
### Summary 

This pull request removes the outdated semester caution modal that used to pop up when trying to add duplicate courses when we disallowed that action. Since we don't ever prevent users from adding courses to their plan in this way anymore, the modal can be removed.

### Test Plan 

Make sure the site still works, and to be extra careful, confirm the regular confirmation modal still pops up.

### Notes 

This is part of the larger modal refactor happening right now, but made in a separate PR, so that it could easily be referenced in case this modal ever needed to be added back.


tcho6319 (approved):
Test plan works. LGTM!

PR #485, opened 2021-04-30
Run courses-gen for FA21 data

willespencer:
### Summary 

This pull request runs courses-gen to retrieve FA21 data before the next push to prod

### Test Plan 

Site works and has updated course data. Confirm a new course such as INFO 5101 can be added, and that the bottom bar shows up with expected information for it.

### Notes 

Looks like we're getting an error with CUReviews? when selecting courses that are completely new this semester, and the bottom bar does not pop up. Should I investigate resolving this as part of this PR?

image


SamChou19815 (commented):
JSON changes look good to me. I think it errors because CUReviews do not have the class yet. I think we need a way to gracefully handle this instead of let it crash.

tcho6319 (approved):
LGTM! We should handle the case where CU Reviews doesn't have a new course (e.g. INFO 5101) in another PR. 

https://github.com/cornell-dti/course-plan/blob/3e457c50db4bda05166810ed3ebe471e2b066cb5/src/components/BottomBar/BottomBarState.ts#L77-L81
In the case where we can't fetch CU Reviews data, we should set the `bottomBarCourse.overallRating`,  `bottomBarCourse.difficulty`, `bottomBarCourse.workload` to `None` so that we show 'N/A'.

See  https://github.com/cornell-dti/course-plan/blob/3e457c50db4bda05166810ed3ebe471e2b066cb5/src/components/BottomBar/BottomBarCourseReview.vue#L121-L135

PR #486, opened 2021-05-01
Set CUReviews data to N/A for new FA21 courses

willespencer:
### Summary 

This pull request resolves the remaining issue in #485 - that the bottom bar could not be opened for new FA21 courses because there is no data for them yet on CUReviews. Also resolves related bottom bar bugs

- [x] Set CUReviews data to null when the course does not exist on CUReviews yet
- [x] Fix bug where courses with no descriptions in the bottom bar (like some of the new courses) are styled the same
- [x] Explore bottom bar tabs as buttons 

image

### Test Plan 

1. Ensure that clicking a new FA21 course like INFO 5101 or MUSIC 6270 opens up the bottom bar with "N/A" for each CUReviews field, and that the CUReviews link goes to their 404 page.
2. Ensure clicking a course with data such as CS 2110 still displays CUReviews data as expected, and switching tabs between that course and the new course still preserves the right data in the bottom bar.

### Notes 

Per the TPM of CUReviews, course data is usually added at the beginning of the semester (although they might do so soon because of a bug they are having), so this fix should be useful in future semesters as we add future semester data.

### DIFF
--- a/src/components/BottomBar/BottomBarCourseReview.vue
+++ b/src/components/BottomBar/BottomBarCourseReview.vue
@@ -142,6 +142,7 @@ export default defineComponent({
 
 .details {
   padding: 20px;
+  width: 100%;
 
   &-head {
     margin-top: 10px;
--- a/src/components/BottomBar/BottomBarState.ts
+++ b/src/components/BottomBar/BottomBarState.ts
@@ -54,7 +54,15 @@ const getReviews = (
     method: 'POST',
     headers: { 'Content-Type': 'application/json' },
     body: JSON.stringify({ subject: subject.toLowerCase(), number }),
-  }).then(res => res.json().then(reviews => reviews.result));
+  }).then(res =>
+    res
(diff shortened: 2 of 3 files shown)


SamChou19815 (approved):

tcho6319 (changes requested):
Thanks Will! There seems to be a styling discrepancy for the courses that don't exist on CU Reviews (e.g. INFO 5101). 


https://user-images.githubusercontent.com/54298311/116800963-b5cf1180-aad3-11eb-9cc4-473f5f09cbb9.mov

The bars for the three ratings should be wider as the screen width decreases. I compare INFO 5101 to DSOC 6110 (a course that exists on CU Reviews).

Once this discrepancy is resolved, this PR should be g2g!

tcho6319 (approved):
LGTM!

tcho6319 (approved):
LGTM!

PR #487, opened 2021-05-02
Onboarding Required Fields Bug

tcho6319:
### Summary 
Relevant Notion item: [Onboarding Required Fields bug](https://www.notion.so/Onboarding-Required-Fields-bug-d9aed9bd517a43ceb7b8a1b2066aa939)

This pull request resolves the onboarding required fields bug where the user would see the error message for any missing required fields only when they try to finish onboarding. 

For example: If I don't input my last name, I get the error message when I'm on the last page of onboarding and click "Finish":
Screen Shot 2021-05-02 at 12 37 58 PM

Now, the user cannot move past the first page of onboarding if any required fields are missing:
Screen Shot 2021-05-02 at 12 47 33 PM

Also, in the case where an added major/minor is on a branch but not on master, an error message is displayed and the user cannot proceed to the transfer page:
Screen Shot 2021-05-02 at 12 49 09 PM

- [x] Prevent user from moving past first page if required fields are not completed
- [x] Prevent user from moving past first page if they have an invalid major/minor.
- [x] Modify `getCollegeFullName()`, `getMajorFullName()`, `getMinorFullName()` so that if a college, major, or minor doesn't exist in the requirementsJSON, it returns an empty string. 

### Test Plan 
1. Delete your onboarding document in `user-onboarding-data` on Firebase (simulate new user)
2. Make sure you can't move past the first page without completing required fields (i.e. first name, last name, college, entrance/grad years)
3. Delete the entrance year field (simulate existing user without entrance year) and make sure you can't move past the first page without completing the field.
4. Add a bogus major. Return to onboarding modal and make sure you can't move onto next page unless you delete the placeholder for the bogus major. Same for a bogus minor.
5. Ensure onboarding data saves as usual.

### DIFF
--- a/src/components/Modals/Onboarding/Onboarding.vue
+++ b/src/components/Modals/Onboarding/Onboarding.vue
@@ -118,11 +118,32 @@ export default defineComponent({
   data() {
     return {
       currentPage: 1,
-      isError: false,
-      isInvalidMajorOrMinorError: false,
       name: { ...this.userName },
       onboarding: { ...this.onboardingData },
     };
+  },
+  computed: {
+    // Display error if a required field is empty
+    isError(): boolean {
+      return (
+        this.name.firstName === '' ||
+        this.name.lastName === '' ||
+        this.onboarding.college === '' ||
+        this.onboarding.gradYear === '' ||
+        this.onboarding.entranceYear === ''
+      );
+    },
+    // Display error if onboarding data includes a major or minor that doesn't exist in requirementsJSON
+    isInvalidMajorOrMinorError(): boolean {
+      return (
+        this.onboarding.major
+          .map(getMajorFullName)
+          .some((majorFullName: string) => majorFullName === '') ||
+        this.onboarding.minor
+          .map(getMinorFullName)
+          .some((minorFullName: string) => minorFullName === '')
+      );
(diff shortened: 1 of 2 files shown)


src/components/Modals/Onboarding/Onboarding.vue
```diff
@@ -155,7 +152,29 @@ export default defineComponent({
...
+          this.onboarding.gradYear === '' ||
+          this.onboarding.entranceYear === '';
+
+        this.isInvalidMajorOrMinorError =
+          this.onboarding.major
+            .map(getMajorFullName)
+            .some((majorFullName: string) => majorFullName === '') ||
+          this.onboarding.minor
+            .map(getMinorFullName)
+            .some((minorFullName: string) => minorFullName === '');
```
SamChou19815, line 172: Why don't we put this as computed property, so that the warning can auto-showup in button before clicking next

SamChou19815 (approved):
The flow looks good. Just one nit.

src/components/Modals/Onboarding/Onboarding.vue
tcho6319: Yes good call! I will push those changes

PR #488, opened 2021-05-02
Create tests for computeFulfillmentCoursesAndStatistics

SamChou19815:
### Summary 

I'm about to add in more logic in computeFulfillmentCoursesAndStatistics to compute progress for compound requirements,
so I think it's good to test the existing code before we introduce more complexity.

I commented what each test is doing in the comments. 

### Test Plan 

`npm run test`


tcho6319 (approved):
LGTM! Thanks for the documentation too, Sam.

PR #489, opened 2021-05-02
AP/IB Overriden Requirements - Data Store

benjamin-shen:
### Summary 

This pull request is the first step towards implementing AP/IB overriding. [Notion](https://www.notion.so/courseplan/Discussion-Story-AP-IB-Infra-Update-to-handle-opt-out-and-requirement-overriding-ee2c3f234c6c44018531ed6ec34c56a1)

- [x] refactored onboarding data submission to `global-firestore-data.ts`
- [x] updated onboarding data exam types and tweaked `setOnboardingData` in `store.ts`
- [x] added functions in `global-firestore-data.ts` to add or clear AP/IB overriden requirements
- [x] implemented clearing AP/IB overriden requirements on college change
- [x] wrote tests for `global-firestore-data.ts` updater functions

Remaining TODOs:

- [ ] integrate in requirement graph
- [ ] attach frontend to `addOverridenRequirementAPIB` function for overriding (opt in) and opting out
- [ ] add gtag analytics (?)

### Test Plan 

1. I refactored the onboarding data submission, so it should be tested to behave as expected.
1. I wrote a test suite for the new updater functions in `global-firestore-data.ts`. You can run it via `npm run test override-apib`.
1. The `clearOverridenRequirementsAPIB` function has been fully integrated in the web app and is called when the user changes colleges. Feel free to test this as well -- you can manually add `optIn` and `optOut` in `user-onboarding-data` according to the schema on Notion (or see below), and ensure that changing colleges empties those objects for each user exam.
image

### Notes

1. In the `global-firestore-data.ts` functions, I maintain the invariant that any [`requirementName`,`slotName`] cannot be in both the `optIn` and `optOut` collections. If this invariant is broken (for example, through manually editing Firestore), the requirement graph may have undefined behavior regarding AP/IB overriding. I might add a check for this in the graph computation, but no guarantees.
1. `optIn` and `optOut` are currently optional fields, as per this brief conversation:
image


src/user-data.d.ts
```diff
@@ -32,12 +32,14 @@ type FirestoreSemester = {
   readonly courses: readonly FirestoreSemesterCourse[];
 };
 
-type FirestoreCollegeOrMajorOrMinor = { readonly acronym: string };
 type FirestoreAPIBExam = {
   readonly type: 'AP' | 'IB';
   readonly score: number;
   readonly subject: string;
+  readonly optIn?: Record<string, string[]>;
```
SamChou19815, line 39: Two nits:
- Make the record and array readonly to be consistent with the rest of the datatypes
- Name the record key to make it self-documenting

So something like:
```ts
type APIBOptInOut = {
  // Values are slot names
  readonly [requirementName: string]: readonly string[];
};
```

SamChou19815 (approved):

src/global-firestore-data.ts
```diff
@@ -198,6 +202,113 @@ export const deleteCourseFromSelectableRequirements = (courseUniqueID: number):
...
+      class: onboarding.transferCourse,
+      tookSwim: onboarding.tookSwim,
+    })
+    .then(() => {
+      const oldCollege = store.state.onboardingData.college;
+      store.commit('setOnboardingData', onboarding);
+      const newCollege = store.state.onboardingData.college;
+      if (oldCollege !== newCollege) {
+        clearOverridenRequirementsAPIB();
+      }
```
tcho6319, line 232: I'm not sure that the `optOut` and `optIn` fields are being cleared when the college is changed. I wonder if `oldCollege` and `newCollege` might have the same value even in cases where I'm changing from CALS to A&S. Doesn't the store listen to changes on Firestore? So maybe by setting the `onboardingDataCollection` first, `oldCollege` would actually be the value of the college that you're changing to. Could you move `const oldCollege = store.state.onboardingData.college;` to before you set the onboarding doc?

https://user-images.githubusercontent.com/54298311/117287936-ef3bb000-ae38-11eb-8afe-fe7eb15df456.mov

tcho6319, line 232: I'm not sure that the `optOut` and `optIn` fields are being cleared when the college is changed. I wonder if `oldCollege` and `newCollege` might have the same value even in cases where I'm changing from CALS to A&S. Doesn't the store listen to changes on Firestore? So maybe by setting the `onboardingDataCollection` first, `oldCollege` would actually be the value of the college that you're changing to. Could you move `const oldCollege = store.state.onboardingData.college;` to before you set the onboarding doc?

https://user-images.githubusercontent.com/54298311/117287936-ef3bb000-ae38-11eb-8afe-fe7eb15df456.mov

tcho6319 (changes requested):
Found a bug related to clearing the `optOut` and `optIn` but everything else looks good. Thanks Ben!

src/global-firestore-data.ts
benjamin-shen: Oh that's so funny, I actually changed it in [a later commit](https://github.com/cornell-dti/course-plan/pull/489/commits/d58445441f3684219b5544466c403acbd04d269f) but didn't realize it would break it. Now that I think about it, because of the Firestore listeners, none of the `store.commit(...)` calls need to happen except for in `editSemesters()`, right? I think right now they're probably redundant.

src/global-firestore-data.ts
```diff
@@ -40,8 +42,10 @@ const editSemesters = (
   updater: (oldSemesters: readonly FirestoreSemester[]) => readonly FirestoreSemester[]
 ): void => {
   const newSemesters = updater(store.state.semesters);
-  store.commit('setSemesters', newSemesters);
-  semestersCollection.doc(store.state.currentFirebaseUser.email).set({ semesters: newSemesters });
+  semestersCollection
+    .doc(store.state.currentFirebaseUser.email)
+    .set({ semesters: newSemesters })
+    .then(() => store.commit('setSemesters', newSemesters));
```
tcho6319, line 48: Is this change necessary? I think it might be causing the two bugs:

Dragging a course to a different semester deletes the course. First tab is this PR, second tab is dev staging:

https://user-images.githubusercontent.com/54298311/117380056-ebdf0d80-aea6-11eb-854b-a124a3304bae.mov

When I add a course to a requirement with a checkerWarning (like Technical Electives) the course doesn't show. First tab is this PR, second is dev:

https://user-images.githubusercontent.com/54298311/117380370-8b9c9b80-aea7-11eb-98e8-458cf5fc86f1.mov

tcho6319 (changes requested):
Very close! I found a couple bugs related to updating semesters but other than that, the test plan seems to work. Thanks Ben!

src/global-firestore-data.ts
benjamin-shen: Oooh yeah thanks for finding the bug and also finding the relevant code!! Nah the code didn't need to change, it's an artifact of my brain fart regarding the listeners.

tcho6319 (approved):
LGTM!

PR #490, opened 2021-05-03
Added Spanish major and minor, and Food Science major

vaishnavi17:
### Summary 

In this PR, I added the Spanish major and minor, as well as the Food Science major (with concentrations). The details are given here:
https://www.notion.so/courseplan/Add-a-major-minor-3100891bc8274ba59d42900118d2f5f2

Remaining TODOs:

- [ ] Add Food Science minor (requirements not available on website)

### Test Plan 

Test in staging.


src/requirements/data/majors/foodsci.ts
```diff
@@ -0,0 +1,278 @@
...
+        ),
+        perSlotMinCount: [1, 1, 1, 1, 1, 1, 1, 1, 2],
+        slotNames: [
+          'FDSC 4230',
+          'FDSC 5970',
+          'Economics: ECON 1110 or ECON 3030 or PAM 2000',
+          'Management: AEM 1200 or HADM 1810 or ILRID 1700 or AEM 2200 or ILROB 1220',
+          'Marketing: AEM 2400 or HADM 2410 or NCC 5530 or AEM 2420',
+          'Accounting: AEM 2210 or HADM 2230 or NCC 5500 or AEM 2225',
+          'Finance: AEM 2241 or NCC 5560 or HADM 2250 or AEM 2240',
```
SamChou19815, line 269: Since we already have sensible names for these categories, we don't need to list course code here. The frontend will auto show the eligible courses it's not fulfilled.

SamChou19815 (approved):
LGTM with a nit.

src/requirements/data/majors/foodsci.ts
vaishnavi17: Thanks, fixed that!

SamChou19815 (approved):

src/requirements/data/majors/foodsci.ts
```diff
@@ -0,0 +1,276 @@
...
+            'FDSC 4880',
+            'FDSC 5970',
+            'FDSC 6950',
+            'FDSC 4010',
+            'FDSC 4040',
+            'FDSC 4050',
+            'FDSC 4220',
+            'FDSC 4250',
+            'FDSC 4400',
+          ]
```
tcho6319, line 174: If the user needs 3 >4000 FDSC electives, would we need another subrequirement slot that is "Group 1 or Group 2 >4000 electives" where it includes courses in both of these arrays?

src/requirements/data/majors/foodsci.ts
```diff
@@ -0,0 +1,276 @@
...
+            'FDSC 5400',
+          ]
+        ),
+        perSlotMinCount: [1, 1, 1, 1, 1, 1, 1],
+        slotNames: [
+          'FDSC 3950',
+          'FDSC 4230 or FDSC 4250',
+          'FDSC 5970',
+          'NS 3600 or PAM 3280 or ENTOM 4250 (in preferred order)',
+          'PAM 4240 or DSOC 4080 or ORIE 2380 (in preferred order)',
```
tcho6319, line 219: Does "in preferred order" mean that they can take those requirements in any order? If so, I don't think "(in preferred order)" is necessary to have in the slot name.

src/requirements/data/majors/foodsci.ts
```diff
@@ -0,0 +1,276 @@
...
+          'BIOMS 4040 or BIOMS 4090 or BIOMS 4150 or BIOMS 4340 or BIOMG 4390 (in preferred order)',
+          '2 credits of >4000 FDSC electives',
```
tcho6319, line 221: Nit: To be consistent with other subrequirements that mention credit minimums, can you say 4000+ instead of >4000? This is for here and other places as well.

src/requirements/data/majors/spanish.ts
```diff
@@ -0,0 +1,55 @@
...
+    allowCourseDoubleCounting: true,
+    checker: [
+      (course: Course): boolean =>
+        (ifCodeMatch(course.subject, 'SPAN') || ifCodeMatch(course.subject, 'PORT')) &&
+        ifCodeMatch(course.catalogNbr, '4***'),
+    ],
+    checkerWarning: 'We do not check that the course is a senior seminar.',
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+    slotNames: ['Course'],
```
tcho6319, line 51: If the Senior Seminar is supposed to be included in the 15 credits for Electives, I believe you'll have to make this subrequirement allow double counting. `allowCourseDoubleCounting: true` should be added. I would add a comment that says we have to allow double counting here so that the Senior Seminar req can be double counted with Electives.

src/requirements/data/minors/spanish.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  includesWithSingleRequirement,
+  ifCodeMatch,
+  courseMeetsCreditMinimum,
+} from '../checkers-common';
+
+const spanishMinorRequirements: readonly CollegeOrMajorRequirement[] = [
+  {
+    name: 'Advanced Language Course',
+    description: 'SPAN 2130 or SPAN 2180',
```
tcho6319, line 11: Nit: Make sure descriptions end with periods here and in your other data files.

tcho6319 (changes requested):
Thanks Vaish! Requested some smaller changes but once those are addressed, this PR should be g2g!

src/requirements/data/majors/foodsci.ts
vaishnavi17: No, it means that these are the courses that they can take to fulfill the requirement, but NS 3600 is preferred over PAM 3280, for e.g.

src/requirements/data/majors/spanish.ts
vaishnavi17: Wait, I have added allowCourseDoubleCounting: true to the senior seminar requirement. Do I also need to add it for the 15 credit requirement?

src/requirements/data/majors/foodsci.ts
vaishnavi17: Right, that makes sense. Done!

src/requirements/data/majors/spanish.ts
tcho6319: Oh sorry I missed that! Actually I think it works when adding from the Add Modal. I was trying to do this by dragging from the sidebar which is a pre-existing bug that we have to resolve. Disregard this requested change!

src/requirements/data/majors/foodsci.ts
tcho6319: Oh I see. I wonder if by saying something like `'NS 3600 (or PAM 3280 or ENTOM 4250)'` it kind of implies a course is more recommended than the other.

tcho6319 (approved):
LGTM! I left a suggestion for the slot names with "in preferred order" if you want to try it!

PR #494, opened 2021-05-05
Reinstate Search with Exact Course Code and Title Match without Fuse.js

tcho6319:
### Summary 

This pull request reinstates the old course search implemented by @SamChou19815 in #351 for the May 4/5 push to prod.

The Fuse search implementation (#482) does improve results but suffers from slow performance. 

Remaining TODOs:

- [ ] Find a way to optimize search without compromising too much on speed

### Test Plan 

Try looking up a course (e.g. CS 1110). The course should appear in the search results and not be too slow.

### DIFF
--- a/src/components/Modals/NewCourse/CourseSelector.vue
+++ b/src/components/Modals/NewCourse/CourseSelector.vue
@@ -25,19 +25,34 @@
 
 <script lang="ts">
 import { PropType, defineComponent } from 'vue';
-import Fuse from 'fuse.js';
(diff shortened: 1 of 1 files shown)


einc (approved):
Test plan works! Do we want to comment out the new additions for now?

tcho6319:
> Test plan works! Do we want to comment out the new additions for now?

I think it'd be cleaner to remove the new additions instead of commenting them out. We can go back to the original Fuse PR when we want to try Fuse.js again

PR #5, opened 2019-11-18
Initial Requirements Menu Setup

hanywang2:
### Summary 

Developed the frontend for the requirements menu and made it adaptable with Vue's template binding. And made minor style changes to dashboard.

- [x] Added Bootstrap and JQuery

### Test Plan 

@willespencer @tcho6319 @ayesha4698 please test the functionality of basic buttons and provide feedback on the data setup for the requirements menu.

PR #496, opened 2021-05-06
Styling Bug Fix: Remove Bootstrap

hahnbeelee:
### Summary
Fixes this styling bug:
image

We are assuming this happened because the bootstrap updated post our push to prod.

Initially I tried to override it, but then I realized we always make our own styles and never refer to bootstrap styling. So, I just got rid of it because it seemed like a very outdated addition and after getting rid of it, nothing seems to change on the frontend styling.

### Test Plan
- Inspect all aspects of the frontend and make sure none of the styling looks off (bottom bar, req bar, onboarding, modals)
- Inspect mobile view

### Notes
[The PR where the Bootstrap was added](https://github.com/cornell-dti/course-plan/pull/5/files#diff-fc211eca36b0b8e521eb5d0f4347adbcac3ca210ba294cd0330e07a3d16e8308)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -14,11 +14,6 @@
     <link
       type="text/css"
       rel="stylesheet"
-      href="//unpkg.com/bootstrap/dist/css/bootstrap.min.css"
-    />
-    <link
-      type="text/css"
-      rel="stylesheet"
       href="//unpkg.com/bootstrap-vue@latest/dist/bootstrap-vue.min.css"
     />
 
--- a/src/assets/scss/_global.scss
+++ b/src/assets/scss/_global.scss
@@ -262,3 +262,9 @@ form {
     opacity: 0.8;
   }
 }
+
+.row > * {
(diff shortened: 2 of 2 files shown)


willespencer (approved):
Followed the test plan and it all looks good! Also checked the landing page to confirm it looks like normal.

Do you want to make a notion item in the backlog for removing any leftover bootstrap classes we might have in the codebase somewhere now that it isn't being imported?

tcho6319 (approved):
LGTM as well! Thanks Hahnbee. I like @willespencer's suggestion above to create a notion item. Can you do that and put it in the backlog?

SamChou19815 (commented):
As a note, we do use bootstrap:
https://github.com/cornell-dti/course-plan/blob/f2d7f1c2daeea7f3508527678a831aab47aae3ad/src/main.ts#L5

However, that one is version-managed by npm so it won't be accidentally updated.

PR #499, opened 2021-05-09
Modal Refactor Cleanup

willespencer:
### Summary 

This pull request cleans up some issues from the modal refactor (#491) found in [this Notion item](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b14db3e236ba41f4b733aeeade94239d) in the backlog

- [x] Remove all warnings found from classes put on modals that have no affect
- [x] Deleted unused refs, unused modal styling outside of TeleportModal
- [x] Remove some of the z-index toggles remaining from @hahnbeelee's PR #430
- [x] Remove now unneeded v-model from all modals 

image

Remaining TODOs:

- See other items on the backlog Notion item.

### Test Plan 

1. Open the console, and use every modal listed as a TeleportModal on [this writeup](https://www.notion.so/courseplan/Modal-Refactor-fbbea3dfda52468db4719f53df7d9464). Confirm using each results in no console warnings (including TourWindow) and the styling looks as expected.
2. Use the requirements bar on mobile, and confirm reset and self check course modals work as expected with their z-indices.


tcho6319 (approved):
Thanks Will!  LGTM overall except for a small nit. Also, can you link this PR to the completed checkbox on the [Notion item](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b14db3e236ba41f4b733aeeade94239d) with the remaining TODOs so that devs know which PR addressed those issues?

Nit: This is on master as well but I noticed that the confirmation modal isn't really centered, especially on mobile. If you can't get to this bug, can you add it to the [same Notion Story](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b14db3e236ba41f4b733aeeade94239d) with the other modal refactor finishing touches?

https://user-images.githubusercontent.com/54298311/117581989-e0b9f680-b0cd-11eb-88bb-a1e906f687a2.mov

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -1,13 +1,11 @@
 <template>
   <div class="incompleteselfcheck">
     <new-self-check-course-modal
-      class="incompleteselfcheck-modal"
       v-model="isCourseModalOpen"
       v-if="isCourseModalOpen"
```
neketka, line 5: Should've mentioned this on the older PR, but this form of opening a modal kind of counteracts the whole point of v-model. I get that the v-if will reset the data inside the modal, but maybe we need to look towards an older way of implementing the body of the modal. That is, make the body of the modal as a component and instead use TeleportModal where the current modal components are defined right now. Or, you could instead have two components per modal (a little much, which is why I sort of dislike having Single File Components) which forward props to the modal body component.  The two-component option might be better if the modal has functionality that requires the modal body component to need to listen to events from modal buttons for purposes other than closing an opening. If we end up going with the v-if solution, the v-model serves technically no purpose and the modal may as well be always open. This is quick so you can do it in this PR if you want, but you can also add it in the todo list.

neketka (changes requested):
Should've mentioned this on the older PR, but this form of opening a modal kind of counteracts the whole point of v-model. I get that the v-if will reset the data inside the modal, but maybe we need to look towards an older way of implementing the body of the modal. That is, make the body of the modal as a component and instead use TeleportModal where the current modal components are defined right now. Or, you could instead have two components per modal (a little much, which is why I sort of dislike having Single File Components) which forward props to the modal body component.  The two-component option might be better if the modal has functionality that requires the modal body component to need to listen to events from modal buttons for purposes other than closing an opening. If we end up going with the v-if solution, the v-model serves technically no purpose and the modal may as well be always open. This is quick so you can do it in this PR if you want, but you can also add it in the todo list. Lmk which one you do so I can resolve this.

willespencer:
> Thanks Will! LGTM overall except for a small nit. Also, can you link this PR to the completed checkbox on the [Notion item](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b14db3e236ba41f4b733aeeade94239d) with the remaining TODOs so that devs know which PR addressed those issues?

Done!
 
> Nit: This is on master as well but I noticed that the confirmation modal isn't really centered, especially on mobile. If you can't get to this bug, can you add it to the [same Notion Story](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b14db3e236ba41f4b733aeeade94239d) with the other modal refactor finishing touches?

Fixed! However, I am noticing that the confirmation modal on certain breakpoints covers the new semester button, since it is the center of the page instead of the center of the semesterview like it used to be. Is this a concern? I think if we wanted the previous behavior, we'd not be able to use TeleportModal?

image

src/components/Modals/TeleportModal.vue
```diff
@@ -178,6 +178,9 @@ export default defineComponent({
     min-height: 0;
     margin-left: auto;
     margin-right: auto;
+    display: flex;
+    align-items: center;
+    pointer-events: none;
```
tcho6319, line 183: This might have influenced how the Reset Requirement modal looks. Also the Reset requirement modal doesn't seem to reset.

https://user-images.githubusercontent.com/54298311/118169063-833be780-b3f6-11eb-9afd-fd4056c929db.mov

tcho6319 (changes requested):
Thanks Will! I found a small bug and answered your question below.

> However, I am noticing that the confirmation modal on certain breakpoints covers the new semester button, since it is the center of the page instead of the center of the semesterview like it used to be. Is this a concern? I think if we wanted the previous behavior, we'd not be able to use TeleportModal?

I think it's ok since the confirmation modal is only supposed to appear for a couple seconds. I'd be ok with cutting down the amount of time the confirmation modal appears because I think it is displayed for a little too long. Just a suggestion!

src/components/Modals/TeleportModal.vue
tcho6319: Other modals look ok to me

neketka (approved):

src/components/Modals/TeleportModal.vue
```diff
@@ -43,7 +43,6 @@ import { defineComponent, ref } from 'vue';
 
 export default defineComponent({
   props: {
-    modelValue: { required: true, type: Boolean }, // true if the modal is visible, false otherwise
```
tcho6319, line 46: So to confirm: Do we not need `modelValue` because we're using v-if?

tcho6319 (approved):
Left a quick clarification question but LGTM!

src/components/Modals/TeleportModal.vue
willespencer: Yeah. Originally we had the v-if instead of TeleportModal itself, which is why we needed v-model, so that both TeleportModal and the parent of the modal had an updated display boolean. However, when we switched the v-if to be on the parent itself (for proper data resetting), there was no reason to use v-model or pass the boolean down through the children per @neketka. So this commit shouldn't functionally change anything, just replace the v-models with emits so the parents can set the display booleans true/false.

PR #500, opened 2021-05-09
Graduate Program Support: Onboarding and Firestore

willespencer:
### Summary 

This pull request adds graduate programs to Onboarding and our Firestore database, so that users can now select a graduate program (with or without undergrad). After being integrated with @vaishnavi17's requirements side code, it should be easy to get a new section on the requirements sidebar. See [Notion](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b248521fc9464869a61df28e5a352c16)

- [x] Added "Undergraduate" and "Graduate" sections on the first and last page of Onboarding.
- [x] Create more specific errors. The first page can be passed only if all required fields are selected, and at least one program as well. Error text now tells users exactly what they are missing. Modal can only be cancelled/clicked outside if they have a program and/or college on firebase.
- [x] Skip the transfer page if only a graduate program is selected, and delete transfer data if submitted. Update progress bar svgs to reflect this.
- [x] Uses temp graduate programs to test that everything works in Onbaording and Firebase.
- [x] Update styling on first and last Onboarding pages to reflect designs
- [x] Removed outdated Safari check from @hahnbeelee's temporary modal fix.  

image

image

Remaining TODOs:

- [ ] Integrate with the requirements changes at the 3 locations I marked with TODO comments (and delete temp data)
- [ ] Make a separate PR or update this one after requirements changes are done to actually add a graduate program section to the requirements sidebar (@tcho6319 thoughts?).

### Test Plan 

1. Completely delete your firebase data, and confirm the error message that shows up mentions all missing required fields (and a degree).
2. Add an undergraduate college, add transfer credit, say you've taken the swim test, and submit the modal, and confirm everything works as before.
3. Remove the undergraduate college, select a graduate program. Confirm the transfer credit page cannot be accessed and submit. The requirements bar should be blank.
4. Open the onboarding model again, add back an undergraduate college, confirm the program you selected is still there, but that the transfer data/swim test has been deleted.
5. Submit and confirm you can have both a college and program selected at once.
6. Mess around with the required fields and confirm all error messages look good. 
7. Confirm that the x on the modal can be clicked, and the modal can be clicked outside, only when a user has a college or program on firebase.
8. Confirm that setting the firebase program to some invalid acronym (that is not TEMP, TEMP2, or TEMP3) leads to the error message @tcho6319 created about data on firebase that does not exist on this branch.
9. Check that all the new styling looks good on mobile.

### Notes 

I have a bunch of questions based on what I noticed when making these changes. Let me know what I should address!
1. Should we avoid using an image as the progress bar? This seems bad for accessibility because the text is part of the image, and not actual text that could be read by a screen reader
2. Should I update the `FirestoreCollegeOrMajorOrMinor` type name to include Program?
3. We use the style `onboarding-hidden` quite frequently in onboarding instead of using `v-if`. Should I change this like I did with the modal refactor?
4. Should the button on the first page be disabled if there is an error, so users don't think they can click it?
5. I sense an issue with the walkthrough coming up when the requirements side is added. The walkthrough depends on the user having a PE requirement, but this will not be the case for pure graduate students anymore. What do you think @einc ?

### Blockers 

- The requirements-side update coming soon
- Actual graduate program data

### Breaking Changes  

- There is now a `programs` field in `user-onboarding-data` like the `colleges` field. This should not break anything, but worth mentioning.


src/components/Modals/Onboarding/Onboarding.vue
```diff
@@ -130,38 +128,92 @@ export default defineComponent({
...
-        this.onboarding.college === '' ||
         this.onboarding.gradYear === '' ||
-        this.onboarding.entranceYear === ''
+        this.onboarding.entranceYear === '' ||
+        (this.onboarding.college === '' && this.onboarding.program === '')
       );
     },
-    // Display error if onboarding data includes a major or minor that doesn't exist in requirementsJSON
-    isInvalidMajorOrMinorError(): boolean {
+    // Display error if onboarding data includes a major, minor, or graduate program that doesn't exist in requirementsJSON
```
tcho6319, line 136: I wonder if we could make this into a docstring? Same for `errorText`

src/components/Modals/Onboarding/OnboardingReview.vue
```diff
@@ -54,13 +54,11 @@
...
+          <span class="onboarding-subHeader2-review"> Your Undergraduate Degree</span>
         </div>
         <div class="onboarding-selectWrapper">
           <div class="onboarding-selectWrapper-review">
-            <label class="onboarding-label"
-              >College<span class="onboarding-required-star">*</span></label
-            >
+            <label class="onboarding-label">College</label>
             <div>
               <label class="onboarding-label--review">{{ collegeText }}</label>
```
tcho6319, line 63: The placeholder text for College shows if I'm just doing a grad program. It seems for major and minor we use `getMajorFullName` and `getMinorFullName` so I would take a look at those to show an empty string for college. 

Screen Shot 2021-05-11 at 9 08 15 AM

tcho6319 (changes requested):
Thanks Will! Excited for grad programs :) I found some bugs and answered your questions:

From Test Plan:
> 3. Remove the undergraduate college, select a graduate program. Confirm the transfer credit page cannot be accessed and submit. The requirements bar should be blank.

My requirements bar isn't blank. Also I noticed that when we delete the college, the majors/minors aren't deleted which could be the reason why we see Biological Sciences still.

https://user-images.githubusercontent.com/54298311/117815170-34922000-b233-11eb-9f02-2fe44121361f.mov

Also, on a related note. When we change from undergrad program only to grad program only, we still see College Requirements and Major Requirements (including their dropdowns). :

https://user-images.githubusercontent.com/54298311/117817018-2c3ae480-b235-11eb-98b1-4d7728a529c9.mov

For your questions:
> 1. Should we avoid using an image as the progress bar? This seems bad for accessibility because the text is part of the image, and not actual text that could be read by a screen reader

Yes, I agree. This is not in the scope of this PR, so can you create an Accessibility Story on Notion backlog and add this checkbox to the description?

> 2. Should I update the FirestoreCollegeOrMajorOrMinor type name to include Program?

Yes I think that's a good call. This is in scope of the PR.

> 3. We use the style onboarding-hidden quite frequently in onboarding instead of using v-if. Should I change this like I did with the modal refactor?

Yes, v-if would be cleaner if we want to conditionally display something. This is in the scope of this PR.

> 4. Should the button on the first page be disabled if there is an error, so users don't think they can click it?

Yes, that would be a good idea. I think there was styling for disabling the Finish button when we waited until the last page to check for errors. In scope for this PR!

> 5. I sense an issue with the walkthrough coming up when the requirements side is added. The walkthrough depends on the user having a PE requirement, but this will not be the case for pure graduate students anymore. What do you think @einc ?

Maybe we could use PE requirement for users with undergrad programs and just the first requirement for users with only grad programs.

willespencer:
> Also, on a related note. When we change from undergrad program only to grad program only, we still see College Requirements and Major Requirements (including their dropdowns). :
> 

Ok, I fixed the related issues, but not this specific one. When console logging the functions used in RequirementSideBar, none of them were firing when college was deleted (leaving only the grad program). Thus, I think the requirements sidebar is just not updating at all. This is more evident when refreshing the page after clearing the college, as the requirements sidebar is then empty instead of containing the old info. If we're not merging this in yet, I think it would be easier to see if this would ~ just work ~ once Grad Programs are added on the requirements side and we add a section for them on the sidebar.

> Yes, I agree. This is not in the scope of this PR, so can you create an Accessibility Story on Notion backlog and add this checkbox to the description?

Done! [Here it is](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b398747d360540cbb7ee9d3f00a6c73e)

> Yes, v-if would be cleaner if we want to conditionally display something. This is in the scope of this PR.

I got rid of these, but I noticed a bug (I think) when doing so. The OnboardingBasicMultiDropdown had a check to set the class of the '+ another major / + another minor' button hidden if the dropdown was empty. However, the dropdown is never empty, as by default it has a single element of an empty string. I added the method 'showAddDropdown' to calculate this, but do we want this fixed behavior where the button only shows up when a major/minor has been selected?

image
image


> Maybe we could use PE requirement for users with undergrad programs and just the first requirement for users with only grad programs.

Going to save this for when requirements data is in for easier implementation and testing.

All other points you brought up in this long comment fixed!

willespencer:
Also had some more Onboarding questions come up while resolving the above ^ @tcho6319 

In OnboardingTransfer.vue, `transferClassSearchboxClassname() ` looks like it is supposed to set the dropdown text to the lighter gray when it is the placeholder. However, the function is never called, and the dropdown is not light gray when reset. Should I make an item to resolve this? Similarly 'getTransferClassSearchboxPlaceholder' is not used, not sure why it is even in the file tbh.

OnboardingTransfer.vue also has two style tags, and imports the Onboarding styling twice. Can one of them be removed, and if so, which?

src/store.ts
```diff
@@ -216,6 +217,7 @@ const createAppOnboardingData = (data: FirestoreOnboardingUserData): AppOnboardi
   college: data.colleges[0].acronym,
   major: data.majors.map(({ acronym }) => acronym),
   minor: data.minors.map(({ acronym }) => acronym),
+  program: data.programs[0].acronym,
```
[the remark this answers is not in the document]
willespencer: Wow thanks for catching this! Just added a proper null check to handle this so we don't have to do any sort of data migration.

src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -262,6 +300,16 @@ export default defineComponent({
...
     },
+    selectProgram(acronym: string) {
+      this.programAcronym = acronym;
+      this.updateBasic();
+    },
+    // remove college and all current majors
+    removeCollege() {
+      this.collegeAcronym = '';
+      this.majorAcronyms = [''];
+      this.updateBasic();
```
tcho6319, line 311: I would also include `this.minorAcronyms = [''];` because we want to clear out the minors as well when we delete college.

https://user-images.githubusercontent.com/54298311/118316701-252a0580-b4c5-11eb-8a2f-1104f6d97963.mov

tcho6319 (changes requested):
Thanks Will! I found a small unresolved issue and answered your questions.

> Thus, I think the requirements sidebar is just not updating at all. This is more evident when refreshing the page after clearing the college, as the requirements sidebar is then empty instead of containing the old info. If we're not merging this in yet, I think it would be easier to see if this would ~ just work ~ once Grad Programs are added on the requirements side and we add a section for them on the sidebar.

I'd be ok with that. Can you note this bug somewhere  (maybe a Notion item for Grad program finishing touches) so we know to actively test for this once we add a grad program?

> I added the method 'showAddDropdown' to calculate this, but do we want this fixed behavior where the button only shows up when a major/minor has been selected?

I think your fix makes sense. I have a feeling the implementation was due to how the designs looked for onboarding but I think your reasoning for showing the + another major/minor once a major/minor is selected would be more intuitive to the user. You can confirm with @einc or @yuxuanchenn if you want a second opinion!
Screen Shot 2021-05-14 at 3 03 32 PM

> In OnboardingTransfer.vue, transferClassSearchboxClassname() looks like it is supposed to set the dropdown text to the lighter gray when it is the placeholder. However, the function is never called, and the dropdown is not light gray when reset.

I believe this is leftover code from when we were considering including adding equivalent Cornell courses for transfer credit pre-public launch. Can you make a Notion item documenting this fix we have to do?

> OnboardingTransfer.vue also has two style tags, and imports the Onboarding styling twice. Can one of them be removed, and if so, which?

Yes, that's strange. I don't think we need both style tags. Can you remove the set of style tags that don't contain any styling and confirm the styling does not change?

src/components/Modals/Onboarding/OnboardingBasic.vue
willespencer: Discussed over Slack. Going to make this change so that if a user removes their college, they remove their minors, as we do not want pure grad students to have minors. Will also disallow the dropdown from being opened if a college is not selected, similarly to how major dropdowns work now.

willespencer:
> I'd be ok with that. Can you note this bug somewhere (maybe a Notion item for Grad program finishing touches) so we know to actively test for this once we add a grad program?

Yep! Created [an item](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=b5e0e799ac7e427e92a8898eeeb1a8d7) with this and something I noted on @vaishnavi17 's PR! 

> I think your fix makes sense. I have a feeling the implementation was due to how the designs looked for onboarding but I think your reasoning for showing the + another major/minor once a major/minor is selected would be more intuitive to the user. You can confirm with @einc or @yuxuanchenn if you want a second opinion!

Confirmed with @einc that this behavior is what we want!
 
> I believe this is leftover code from when we were considering including adding equivalent Cornell courses for transfer credit pre-public launch. Can you make a Notion item documenting this fix we have to do?

Yep! Notion item [here](https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=cc4d71ef69134ee9b2be16d493220acd).
 
> Yes, that's strange. I don't think we need both style tags. Can you remove the set of style tags that don't contain any styling and confirm the styling does not change?

Fixed and works!

tcho6319 (commented):
The changes look good to me, but they aren't reflected on the staging site. I believe you have to resolve the conflicts so that the build can complete.

tcho6319 (approved):
LGTM! I would hold off on merging in until Vaish's PR is merged.

src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -191,7 +191,7 @@ export default defineComponent({
       collegeAcronym: this.onboardingData.college ? this.onboardingData.college : '',
       majorAcronyms,
       minorAcronyms,
-      programAcronym: this.onboardingData.program,
+      gradAcronym: this.onboardingData.grad,
```
tcho6319, line 194: Nit: You could prob do
```
gradAcronym: this.onboardingData.grad ? this.onboardingData.grad : ''
```
like `collegeAcronym` in line 191. Then you probably don't have to do this check in line 262.

src/requirements/__test__/exam-credit.test.ts
```diff
@@ -164,7 +164,6 @@ it('Exam is counted correctly for one major', () => {
     gradYear: '2020',
     entranceYear: '2016',
     college: 'EN',
-    program: '',
```
tcho6319: To confirm: Why don't we include `program` here again?

src/requirements/requirement-frontend-computation.ts
```diff
@@ -286,7 +286,7 @@ export default function computeGroupedRequirementFulfillmentReports(
 
   // college may be undefined if the user has only selected a grad program
   if (college) {
-    groupedRequirementFulfillmentReport.push({
+    groupedRequirementFulfillmentReport.unshift({
```
tcho6319, line 289: I remember you said that order in the `groupedRequirementFulfillmentReport` mattered. Can you add a comment here to briefly explain why we unshift and not push?

tcho6319 (approved):
LGTM! I left a few nits but staging site behaves normally. 

We can either wait for the MPA program PR to be merged in before merging this in (which includes showing the grad program on the requirements bar) or we can merge this in first and then adapt the code to account for MPA afterward. It's up to you!

tcho6319 (commented):
Actually can you push this PR when you're ready? @hahnbeelee's PR (#512) has the bug where the requirements bar doesn't show. I'm thinking it might be easier if after you squash and merge your PR, she just merges in latest master so she doesn't have to resolve the same bug. Does that sound ok?

src/requirements/__test__/exam-credit.test.ts
willespencer: Two reasons. @vaishnavi17 used the prop 'grad' instead of 'program' so I switched to that in all my files. And then she also set the field as optional, so we don't need to include the string here anymore for the type check to pass

src/components/Modals/Onboarding/OnboardingBasic.vue
willespencer: Good call! This got rid of a few extra null checks too, thank you

PR #501, opened 2021-05-09
Compound requirements frontend rendering

SamChou19815:
### Summary 

This PR setups rendering of compound requirements. These changes are made:

- Render radio choices instead of descriptions for compound requirements
- Render the relevant slots for the chosen nested requirement inside a compound requirement
- Change the math minor requirement to use compound requirement for testing

### Test Plan 

Note that this PR does changes the snapshot test. However, right now minors requirements are auto-allow-double-counting, so users never make choices for them, so we don't need to write script to migrate any user choices.

https://user-images.githubusercontent.com/4290500/117584068-a86be580-b0d8-11eb-8bc4-fab2d7ecb21e.mp4


src/components/Requirements/RequirementDescription.vue
```diff
@@ -0,0 +1,102 @@
...
+    <a :style="{ color: `#${color}` }" :href="requirement.source" target="_blank">
+      <strong>Learn More</strong></a
+    >
+  </div>
+</template>
+
+<script lang="ts">
+import { PropType, defineComponent } from 'vue';
+
+/** Show compound requirement for toggle or description */
```
tcho6319, line 43: Thanks for this documentation! I think it would be good practice for devs to add a docstring of what the component is/does either above the `export default defineComponent` line or at the top of this file

src/components/Requirements/RequirementFulfillmentSlots.vue
```diff
@@ -103,16 +104,25 @@ export default defineComponent({
         }
       );
       if (requirementFulfillmentSpec === null) return [];
-      const requirementFulfillmentEligibleCourses = requirementFulfillmentSpec.eligibleCourses;
+      const appliedRequirementFulfillmentSpec =
+        (requirementFulfillmentSpec.additionalRequirements || {})[this.compoundRequirementChoice] ||
+        requirementFulfillmentSpec;
+      const requirementFulfillmentEligibleCourses =
+        appliedRequirementFulfillmentSpec.eligibleCourses;
+      const matchedCourses = (
```
tcho6319, line 112: Can you add a comment for what `matchedCourses` means? Is it supposed to be the taken courses?

tcho6319 (changes requested):
Thanks Sam! Excited for compound requirements :)

1. I found a bug where the requirement is marked complete even though only 3/4 requirements are fulfilled. It seems that if the main requirement is fulfilled (in this case the "Four 3000 or 4000 level MATH courses" requirement), the entire requirement is marked complete:

https://user-images.githubusercontent.com/54298311/117856558-c3199800-b259-11eb-9a44-998d05db34db.mov

src/components/Requirements/RequirementGroup.vue
```diff
@@ -146,6 +146,13 @@ export default defineComponent({
       this.req.reqs.forEach(req => {
         if (req.minCountFulfilled < req.minCountRequired) {
           ongoing.push(req);
+        } else if (
+          req.additionalRequirements == null ||
+          Object.values(req.additionalRequirements).every(
+            it => it.minCountFulfilled >= req.minCountRequired
+          )
+        ) {
+          ongoing.push(req);
```
tcho6319, line 155: Confused on whether the req should be pushed to `completed` in this case. 

If we do want to push this req to `ongoing`, shouldn't we negate the condition because the req would be ongoing if even one of the additional requirements is not above the minCountRequired?
```
!(Object.values(req.additionalRequirements).every(
            it => it.minCountFulfilled >= req.minCountRequired
))
```

I have the same bug - completing the main requirement makes the entire requirement complete:

https://user-images.githubusercontent.com/54298311/117983789-4643f800-b305-11eb-93a9-4c2113634fbe.mov

tcho6319 (changes requested):
So close! I still have the bug with the compound requirement being marked complete too early, but once that is resolved, this PR should be good to go!

src/components/Requirements/RequirementGroup.vue
SamChou19815: Should be fixed now!

tcho6319 (approved):
LGTM yay :D Thank you Sam!

PR #506, opened 2021-05-11
AP/IB Overriden Requirements - Graph & Frontend Computation

benjamin-shen:
### Summary 

This pull request is the second step towards implementing AP/IB overriding. [Notion](https://www.notion.so/courseplan/Discussion-Story-AP-IB-Infra-Update-to-handle-opt-out-and-requirement-overriding-ee2c3f234c6c44018531ed6ec34c56a1)  
It also sets up a lot of the architecture for handling general override functionality.

Regarding AP/IB equivalent courses:
- [x] added AP/IB equivalent course data to global data store (i.e., uniqueId mappings)
- [x] fixed repeated uniqueId bug for AP/IB equivalent courses
- [x] added dummy course for all exams taken that have no equivalent courses

Regarding the requirements algorithm/infrastructure:
- [x] added (AP/IB and general) user override choices to global data store
- [x] modified graph builder to add edges based on user's optIn override choices
- [x] implemented opting in/out of slot fulfillment during the frontend computation

Remaining TODOs:

- [ ] attach frontend modals/buttons
- [ ] add gtag analytics
- [ ] write additional unit tests for requirement graph builder and frontend util functions
- [ ] finish implementing general overriding next semester! :D

Depends on #489

### Test Plan 

1. Add an AP/IB exam normally.
1. Go to Firestore and add optIn/optOut options to the exam you want you opt in/out for. You may have to look through `requirements/data/` to find the right requirement names and slot names. For example:
image
1. Check the requirements bar to ensure the opt in/out was successful.
1. Repeat for all the edge cases / combinations you can think of!

### Notes 

The schema for `AppOverridenRequirementChoices` is currently:
```
type AppOverridenRequirementChoices = Readonly<
  Record<
    number, // uniqueId
    {
      readonly optIn: Record<string, Set<string>>; // { 'requirementName': new Set(['slotName1', 'slotName2']) }
      readonly optOut: Record<string, Set<string>>; // { 'requirementName': new Set(['slotName1', 'slotName2']) }
    }
  >
>;
```
This may influence how we want to design the Firestore collection schema for general overrides.


hahnbeelee:
I tried out the example you showed in your screenshot and it worked as desired!
Question: Does this work with credit requirements?

hahnbeelee:
What is the definition of a special course?

benjamin-shen:
> Question: Does this work with credit requirements?

This is a good point, I thought it would but it seems that credits-fulfilled reqs don't have slotNames (or at least it's an empty array). So, I added a clause to special-case it. They should work now!

src/store.ts
```diff
@@ -124,6 +149,15 @@ const store: TypedVuexStore = new TypedVuexStore({
     ) {
       state.selectableRequirementChoices = selectableRequirementChoices;
     },
+    setOverridenRequirementChoices(
+      state: VuexStoreState,
+      overridenRequirementChoices: AppOverridenRequirementChoices
+    ) {
+      state.overridenRequirementChoices = {
+        ...overridenRequirementChoices,
+        ...computeAPIBOverridenRequirements(state),
```
tcho6319, line 158: Can you add a comment about how this adds overridden requirements from AP/IB data on Firestore?

src/requirements/data/constants.ts
```diff
@@ -1,12 +1,18 @@
...
- * Course id for special course that satisfies total academic credits requirement.
- */
-export const CREDITS_COURSE_ID = 10;
-
-/**
- * Course id for special course that satisfies FWS requirement.
- */
-export const FWS_COURSE_ID = 11;
+/** Course ID for special course that satisfies no requirements. */
+export const NO_EQUIVALENT_COURSES_COURSE_ID = 10;
```
tcho6319, line 2: Just curious: what's an example of a special course that satisfies no req?

src/requirements/requirement-frontend-utils.ts
```diff
@@ -261,30 +261,45 @@ export function getMatchedRequirementFulfillmentSpecification(
...
+    if (!(disallowTransferCredit && courseIsAPIB(courseTaken)) || optInSlotNames) {
       for (
         let subRequirementIndex = 0;
         subRequirementIndex < eligibleCourses.length;
         subRequirementIndex += 1
       ) {
-        if (
-          eligibleCourses[subRequirementIndex].includes(courseId) &&
-          subRequirementProgress[subRequirementIndex] < perSlotMinCount[subRequirementIndex]
+        const slotName = fulfilledBy === 'courses' ? slotNames[subRequirementIndex] : 'Course';
```
tcho6319, line 290: Does this current infra support the case where I want to use my AP Physics C for a requirement with `checkerWarning` (i.e. Technical Electives)? 
Screen Shot 2021-05-14 at 4 06 11 PM

src/requirements/requirement-graph-builder-from-user-data.ts
```diff
@@ -12,20 +12,26 @@ import buildRequirementFulfillmentGraph from './requirement-graph-builder';
...
   const equivalentCourses = coursesTaken.filter(course => course.courseId !== CREDITS_COURSE_ID);
+
+  // generate set for all forfeited equivalent course ids
   const equivalentCourseIds = new Set(equivalentCourses.map(({ courseId }) => courseId));
-  let transferCreditCourses = coursesTaken.filter(course => course.courseId === CREDITS_COURSE_ID);
-  transferCreditCourses = transferCreditCourses.filter(
-    ({ courseId }) => !equivalentCourseIds.has(courseId)
+  // filter out any credits-only courses generated from AP/IB exams
+  const transferCreditCourses = coursesTaken.filter(
+    course => course.courseId === CREDITS_COURSE_ID && !equivalentCourseIds.has(course.courseId)
```
tcho6319, line 22: A little confused here. What do you mean by credits-only courses? Are those courses that don't apply to any other req except the 120 Academic Credit req? Are you saying that courses with courseid CREDITS_COURSE_ID are transfer credit?

src/requirements/data/exams/ExamCredit.ts
```diff
@@ -389,12 +389,14 @@ function userDataToCourses(
...
   userData: ExamsTaken,
-  examType: 'AP' | 'IB'
+  examType: 'AP' | 'IB',
+  uniqueIdDecrementer: number
 ): CourseTaken[] {
   const userExams = userData[examType];
   const exams = examData[examType];
   const courses: CourseTaken[] = [];
-  userExams.forEach((userExam, examIndex) => {
+  let uniqueId = -2 - uniqueIdDecrementer;
```
tcho6319, line 398: Why do we initialize `uniqueId` like this? A clarification comment would be helpful!

src/requirements/data/exams/ExamCredit.ts
```diff
@@ -416,33 +419,46 @@ function userDataToCourses(
       const excludedMajor =
         exam.fulfillment.majorsExcluded && exam.fulfillment.majorsExcluded.includes(major);
       if (!excludedMajor) {
```
tcho6319, line 421: Can you add a comment that this case is where the AP/IB credit can be used for the major requirements

tcho6319 (changes requested):
Thanks Ben! Test Plan worked for me. I left some documentation requests and noted an opt-in situation regarding requirements with the `checkerWarning` (e.g. Technical Electives).

src/requirements/data/constants.ts
benjamin-shen: If there are no equivalent courses for an AP/IB exam, but the user wants to use the exam to override, there needs to be a dummy course + uniqueId. so, if an AP/IB exam has 0 equivalent courses, this dummy course is pushed.

src/requirements/requirement-graph-builder-from-user-data.ts
benjamin-shen: I didn't actually change anything in this code, I just slightly simplified it. However, now that I'm looking at it more closely, I'm also slightly confused.

There is a definitely a mistake in there. `equivalentCourseIds` is all the courseIds that are not CREDITS_COURSE_ID. `transferCreditsCourses` as it is right now is all the courseids that are CREDITS_COURSE_ID (the second filter condition is always true). So, this function just returns `coursesTaken`.

I sat here for a while but I'm not satisfied with any of my attempts at an explanation for why this function is/was necessary. I think it has something to do with the nuances of AP/IB credit. Should I take it out for now? Maybe we should take another look next semester (since it's not the topic of this PR)?

src/requirements/requirement-frontend-utils.ts
benjamin-shen: I don't see why not. Does it break when you test it out? It has a slotName "Course", so that'll need to be added as a slotName.

tcho6319: Yes it does seem to work! My only remaining question then is do we have to know which slot (e.g. Course 3) that we want the AP/IB to apply to? 

When I just used slotName "Course" on Firestore, it seemed to work fine:
Screen Shot 2021-05-15 at 4 19 07 PM

src/requirements/requirement-graph-builder-from-user-data.ts
tcho6319: I think we should take another look next semester. Can you create a Notion Story item on the backlog?

tcho6319 (approved):
LGTM! I replied to your comments (see above). Once you respond to those, this PR should be good to merge.

src/requirements/requirement-frontend-utils.ts
benjamin-shen: > Do we have to know which slot (e.g. Course 3) that we want the AP/IB to apply to?

We don't need to know (nor have control over) the rendered slot name, but we do need to know which data slot name we want the exam to apply to. The rendered slot name assignment order is automatic.

> When I just used slotName "Course" on Firestore, it seemed to work fine.

It actually won't work if we store it as "Course X", only "Course".

It's similar to reqs fulfilled by credits in that the rendered slot name "Course X" is not actually the data slot name. In this case, "Course" is the data slot name (so that's what should be stored on Firestore / needs to be modified if the data slot name changes).

src/requirements/requirement-graph-builder-from-user-data.ts
benjamin-shen: https://www.notion.so/courseplan/Explore-forfeitTransferCredit-function-33203aea99b84f03a19c34d4031c4831

PR #507, opened 2021-05-13
Migrate engineering liberal arts to use compound requirements

SamChou19815:
### Summary 

Title

### Test Plan 

Screen Shot 2021-05-13 at 13 54 43

### Notes 

It has some changes in the snapshot tests, but the deleted ones are the ones that allow double counting, so it won't appear in user choices records. Therefore, no data migration is needed.

However, this PR also changes the requirement to stop allow double counting, since that is only added to avoid user confusion. User would have to readd their courses.


src/requirements/data/colleges/en.ts
```diff
@@ -103,9 +103,7 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
   },
   {
     name: 'Liberal Studies: 6 courses',
```
einc, line 105: nit: can we take out the ": 6 courses"

einc (approved):
Looks good! Thank you Sam!!! Super exciting to see liberal studies as a compounded req 🤩

tcho6319:
@SamChou19815 can you update your PR summary to note that you added allowCourseDoubleCounting so that users wouldn't have to readd their courses?

PR #509, opened 2021-05-13
Allow typing in dropdowns

neketka:
### Summary 

Added the ability to type inside dropdowns and filter out dropdown results in Onboarding

- [x] Added textbox to dropdown
- [x] Filter dropdown based on textbox
- [x] Allow usage of Enter/Return to select
- [x] Reset dropdown upon selecting again

Remaining TODOs:

- [ ] Possibly add autocomplete style feature (tried to make this happen but it ended up janky)
- [ ] Allow usage of arrow keys to select while typing

### Test Plan 

Test that searchable dropdowns work correctly.


src/components/Modals/Onboarding/OnboardingBasicSingleDropdown.vue
```diff
@@ -61,32 +65,54 @@ export default defineComponent({
...
+      prevQuery: '',
       stopClose: false,
       boxBorder: '',
       arrowColor: '',
       placeholderColor: this.choice !== '' ? lightPlaceholderGray : '',
     };
   },
+  computed: {
+    foundChoices(): [string, string][] {
+      return Object.entries(this.availableChoices).filter(v => v[1].startsWith(this.curQuery));
```
tcho6319, line 89: Can we modify this to allow lowercase?

Right now, it seems like it has to be the exact match:

https://user-images.githubusercontent.com/54298311/118371588-3df4f280-b57b-11eb-9d13-b106e765a89b.mov

src/components/Modals/Onboarding/OnboardingBasicSingleDropdown.vue
```diff
@@ -9,24 +9,28 @@
...
-        <div
+        <input
+          type="text"
           class="onboarding-dropdown-placeholder college-major-minor-placeholder"
-          :style="{ color: placeholderColor }"
-        >
-          {{ availableChoices[choice] || 'Select one' }}
-        </div>
+          :style="{ color: placeholderColor, border: 'none', outline: 'none' }"
+          :placeholder="prevQuery === '' ? 'Select one' : prevQuery"
```
tcho6319, line 16: When I click + another major/minor, the Select One placeholder doesn't seem to appear and we have a blank box.
Screen Shot 2021-05-15 at 12 44 55 PM

tcho6319 (changes requested):
Thanks Nikita! I requested a couple changes, but once those are addressed, this PR should be good to go!

 >  Allow usage of arrow keys to select while typing

This would be especially useful. Can you make a Notion Story item and include this on the backlog?

tcho6319 (approved):
LGTM! I realized that the dropdowns for the exams do not allow typing (e.g. I can't type in 'Physics' for AP; I have to select it). Can you add a notion item or add to the one that I mentioned earlier (where you noted allowing keyboard functionality) for allow typing in dropdowns for AP/IB exams?

PR #510, opened 2021-05-13
Grad program backend infra

vaishnavi17:
### Summary 

In this PR, I made modifications to the backend infrastructure files to support graduate programs. The process of adding any new grad programs is the same as it was for majors and minors - the RequirementsJson object in the requirements/data/index.ts file now contains a (currently) empty field called 'grad'. Any new requirements should be added to the key for this field.

TODO:
- [ ] Add tests to \_test\_ folder that test grad program requirement functionality.
- [ ] Add code so that the colour for the grad bar is appropriate: blue if there are college, major, minor and grad requirements, otherwise green.

Depends on #{500}

### Test Plan 
Test the staging version that combines the frontend and backend changes.

Make sure that university requirements are not added when only a grad program is selected

### Notes 
I was wondering, in the requirement-frontend-computation.ts file, lines 229-260 have some code that looks pretty similar across majors, minors and grad programs. Could this be made into a function?


src/requirements/requirement-frontend-utils.ts
```diff
@@ -147,6 +150,21 @@ export function getUserRequirements({
...
+          it =>
+            ({
+              ...it,
+              id: `Grad-${gradProg}-${it.name}`,
+              sourceType: 'Grad',
+              sourceSpecificName: gradProg,
+            } as const)
+        );
+      })
+      .flat(),
```
SamChou19815, line 167: We have an existing test that asserts all the requirement ID is stable: https://github.com/cornell-dti/course-plan/blob/c458ca6892cf6b9b73700e4284453cf45f724c4c/src/requirements/data/requirement-data-consistency.test.ts#L3-L29
You should extend that array with all the grad IDs.

SamChou19815 (commented):
Just one issue about extending existing tests, otherwise looks good to me.

src/global-firestore-data.ts
```diff
@@ -215,6 +215,7 @@ export const setOnboardingData = (name: FirestoreUserName, onboarding: AppOnboar
       colleges: [{ acronym: onboarding.college }],
       majors: onboarding.major.map(acronym => ({ acronym })),
       minors: onboarding.minor.map(acronym => ({ acronym })),
+      grad: onboarding.grad.map(acronym => ({ acronym })),
```
willespencer, line 218: Should this look like the the college line, since we're only going to add one grad program? So `[{ acronym: onboarding.grad}]`

willespencer: Oh, I also think I called this "program" instead of grad in my PR, but when combining our code I can make sure we are consistent! (I prefer grad anyway)

src/requirements/requirement-json-generator.ts
```diff
@@ -120,12 +120,21 @@ const produceSatisfiableCoursesAttachedRequirementJson = (): DecoratedRequiremen
         readonly requirements: readonly DecoratedCollegeOrMajorRequirement[];
       };
     };
+    grad: {
+      [key: string]: {
+        readonly name: string;
+        // Unsure if grad programs can be offered by multiple schools, but allows flexibility.
+        readonly schools: readonly string[];
```
willespencer, line 127: Agree with your comment!

src/requirements/requirement-frontend-computation.ts
```diff
@@ -247,6 +248,15 @@ export default function computeGroupedRequirementFulfillmentReports(
         }
         break;
       }
+      case 'Grad': {
```
willespencer, line 251: Same comment here and throughout this file. Should this code more resemble the college code instead of the major/minor code?

src/utilities.ts
```diff
@@ -50,6 +50,12 @@ export function getMinorFullName(acronym: string): string {
   return minor ? minor.name : '';
 }
 
+export function getGradFullName(acronym: string): string {
```
willespencer, line 53: Haha yeah I thin I wrote this function too. Will resolve after this is merged!

willespencer (commented):
Looks great! Just left a few minor comments/notes, and tested the site and made sure courses could be added to fulfill reqs and ensure that nothing breaks!

src/utilities.ts
vaishnavi17: Sounds good!

src/store.ts
```diff
@@ -216,6 +217,7 @@ const createAppOnboardingData = (data: FirestoreOnboardingUserData): AppOnboardi
   college: data.colleges[0].acronym,
   major: data.majors.map(({ acronym }) => acronym),
   minor: data.minors.map(({ acronym }) => acronym),
+  grad: data.grad.map(({ acronym }) => acronym),
```
tcho6319, line 220: Our current users might not have the `grad` field and so we get a TypeError. 
Screen Shot 2021-05-15 at 1 23 56 PM

I would take a look at my comment on @willespencer's PR: https://github.com/cornell-dti/course-plan/pull/500#discussion_r630116250

If we merge Will's PR first and then yours, it should be ok because I believe he resolved this issue on his end.

src/components/Requirements/RequirementGroup.vue
```diff
@@ -92,6 +92,8 @@ const reqGroupColorMap = {
   College: ['4D7D92', 'sangBlue'],
   Major: ['148481', 'emGreen'],
   Minor: ['105351', 'chrisGreen'],
+  // TODO: make colors wrap around if less than 4 elements, else new colour
+  Grad: ['105351', 'sangBlue'],
```
tcho6319, line 96: If we want sangBlue should the first element be the same as that of College? So 4D7D92?

src/requirements/requirement-frontend-computation.ts
tcho6319: @willespencer Based on the implementation plan, it seemed that the idea was to treat the grad program as the same level as a major/minor since grad programs are technically under colleges. So I think this is ok? What are your thoughts, @willespencer and @vaishnavi17?

src/requirements/types.ts
```diff
@@ -28,6 +28,7 @@ type GenericRequirementsJson<R> = {
   readonly college: CollegeRequirements<R>;
   readonly major: MajorRequirements<R>;
   readonly minor: MajorRequirements<R>;
+  readonly grad: MajorRequirements<R>;
```
tcho6319, line 31: Maybe add a comment here that says we're treating grad programs at the same level as Major Requirements?

src/requirements/types.ts
```diff
@@ -37,4 +38,5 @@ export type DecoratedRequirementsJson = {
   readonly college: CollegeRequirements<DecoratedCollegeOrMajorRequirement>;
   readonly major: MajorRequirements<DecoratedCollegeOrMajorRequirement>;
   readonly minor: MajorRequirements<DecoratedCollegeOrMajorRequirement>;
+  readonly grad: MajorRequirements<DecoratedCollegeOrMajorRequirement>;
```
tcho6319, line 41: Here as well

tcho6319 (changes requested):
Overall LGTM! Thanks Vaish. I left a couple nits, but I would address/reply to the comments that @SamChou19815 and @willespencer made. Once those are done, this PR should be good to go!

src/components/Requirements/RequirementGroup.vue
willespencer: I was going to resolve this on my end when I make sure the requirements bar front-end works! This is just a placeholder we decided on at the dev meeting haha

src/store.ts
willespencer: My concern with merging mine first is mine actually changes things like Onboarding, which will break things until everything is done, while this PR should theoretically not change anything on the front-end. Might be better to just copy my solution over? but not sure

src/requirements/requirement-frontend-computation.ts
willespencer: That sounds good to me! Good to leave this as is.

src/store.ts
tcho6319: @willespencer yes, you have a good point. If @vaishnavi17 can copy your solution over and the staging site looks good, then we can merge hers in first!

src/components/Requirements/RequirementGroup.vue
tcho6319: Sounds good!

src/requirements/requirement-frontend-utils.ts
vaishnavi17: Thanks, I added grad programs to this test!

src/global-firestore-data.ts
vaishnavi17: So the onboarding.college type is a string, which is why it says [{ acronym: onboarding.college }]. If I made grad a string, then we would always have to assign it a value. If it's a string[], then it can be empty (the same way the minors array would be empty if they didn't have a minor).

willespencer (approved):
All the code changes look good to me!

src/store.ts
```diff
@@ -214,10 +214,10 @@ const createAppOnboardingData = (data: FirestoreOnboardingUserData): AppOnboardi
   // TODO: take into account multiple colleges
   gradYear: data.gradYear ? data.gradYear : '',
   entranceYear: data.entranceYear ? data.entranceYear : '',
-  college: data.colleges[0].acronym,
+  college: data.colleges.length !== 0 ? data.colleges[0].acronym : undefined,
   major: data.majors.map(({ acronym }) => acronym),
   minor: data.minors.map(({ acronym }) => acronym),
-  grad: data.grad.map(({ acronym }) => acronym),
+  grad: data.grad.length !== 0 ? data.grad[0].acronym : undefined,
```
tcho6319, line 220: data.grad does not seem to be defined when I test the staging site and don't have the `grad` field in Firestore (emulating our current users). Maybe checking if data.grad is undefined before checking its length will fix this?
Screen Shot 2021-05-16 at 9 55 12 PM

tcho6319 (changes requested):
So close! Found a TypeError bc data.grad doesn't seem to be undefined.

tcho6319 (approved):
LGTM!

PR #511, opened 2021-05-16
Slot settings menu

hahnbeelee:
### Summary

This PR implements the Menu that pops up when you click on the settings button in the slot.
image

- [x] Renamed some variables to cater to the from the deleteCourseModal from the resetConfirmationModal
- [x] Edited TeleportModal to be even more flexible (custom modal positioning)
- [x] Create SlotMenu 

Remaining TODOs:

- [ ] Implement edit functionality / make override requirement modal

### Test Plan
- Make sure delete functionality works
- Make sure frontend looks good

### Notes
- Had to use Teleport because it's embedded into the requirements bar


src/components/Modals/TeleportModal.vue
```diff
@@ -181,5 +208,13 @@ export default defineComponent({
     margin-left: auto;
     margin-right: auto;
   }
+
+  &-slotmenu {
+    border: 1px solid black;
```
tcho6319, line 213: I would remove this line. This is why the border color seems to be a little darker than the designs
Screen Shot 2021-05-16 at 6 22 13 AM

src/components/Modals/TeleportModal.vue
```diff
@@ -181,5 +208,13 @@ export default defineComponent({
...
+    background-color: white;
```
tcho6319, line 214: Use the scss variable `$white` here

src/components/Modals/SlotMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+  <teleport-modal
+    content-class="content-slotmenu"
+    :isSimpleModal="true"
+    :modelValue="modelValue"
+    @modal-closed="closeCurrentModal"
+    :hasClickableTransparentBackground="true"
+    :hasCustomPosition="true"
+    :position="position"
+  >
+    <button class="slotMenu-section" @click="openEditSlotModal">
```
tcho6319, line 11: The hover states are different from the course menu 
Screen Shot 2021-05-16 at 6 26 58 AM
Screen Shot 2021-05-16 at 6 27 15 AM

Try adding the class `'full-opacity-on-hover'` from global scss to these buttons and it should work.

src/components/Modals/SlotMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+  },
+});
+</script>
+
+<style scope lang="scss">
+@import '@/assets/scss/_variables.scss';
+
+.slotMenu {
+  background: $white;
+  border: 1px solid #acacac;
```
tcho6319, line 74: Nit: Do we have an scss variable? If not, can you add it? This applies to other places in this file.

src/components/Modals/TeleportModal.vue
```diff
@@ -53,6 +60,20 @@ export default defineComponent({
     rightButtonIsDisabled: { type: Boolean, default: false },
     isSimpleModal: { type: Boolean, default: false }, // true if the modal will set its own styling for its position
     hasNoBackground: { type: Boolean, default: false }, // true for modals without the gray overlay behind them
+    hasClickableTransparentBackground: { type: Boolean, default: false },
+    hasCustomPosition: { type: Boolean, default: false },
+    position: { type: Object as PropType<{ x: number; y: number }>, default: { x: 0, y: 0 } },
```
tcho6319, line 65: Nit: Like the other props, can you add comments for what these are for?

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -99,6 +102,20 @@ export default defineComponent({
         } else deleteCourseFromSemesters(this.courseTaken.uniqueId, this.$gtag);
       }
     },
+    openSlotMenu(e: MouseEvent) {
+      console.log('window', window);
```
tcho6319, line 106: Remove console log

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -83,10 +86,10 @@ export default defineComponent({
   },
   methods: {
     onReset(): void {
```
tcho6319, line 88: Nit: Might be worth renaming to onDelete() to be consistent with the other name changes from Reset to Delete

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -34,36 +31,42 @@
...
+  components: { ReqCourse, DeleteCourseModal, SlotMenu },
   props: {
     slotName: { type: String, required: true },
     courseTaken: { type: Object as PropType<CourseTaken>, required: true },
   },
   data: () => ({
-    resetConfirmVisible: false,
+    deleteModalVisible: false,
+    slotMenuOpen: false,
+    position: { x: 0, y: 0 },
```
tcho6319, line 64: Consider making position a computed prop so that as the window size changes, the position is recomputed


https://user-images.githubusercontent.com/54298311/118394375-5401c200-b612-11eb-880b-63981769dead.mov

tcho6319 (changes requested):
Thanks Hahnbee! I requested some changes but once those are addressed, this PR should be g2g.

src/components/Modals/DeleteCourseModal.vue
```diff
@@ -35,19 +35,19 @@ export default defineComponent({
...
   emits: {
     'update:modelValue': (value: boolean) => typeof value === 'boolean',
-    'close-reset-modal': (value: boolean) => typeof value === 'boolean',
+    'close-delete-course-modal': (value: boolean) => typeof value === 'boolean',
   },
   methods: {
     closeCurrentModal(): void {
       this.$emit('update:modelValue', false);
     },
-    resetClicked(): void {
```
willespencer, line 44: I think you're going to have some merge conflicts here because of my modal fixes PR. I changed some things from the presentation, so let me know if you need any help resolving!

src/components/Modals/SlotMenu.vue
```diff
@@ -0,0 +1,123 @@
+<template>
+  <teleport-modal
+    content-class="content-slotmenu"
+    :isSimpleModal="true"
+    :modelValue="modelValue"
```
willespencer, line 5: More specifically, we don't need to pass down modelValue anymore, and just use v-if on the parent component and emit up to it. Could you remove this and follow the new style from #499 ?

src/components/Modals/SlotMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+    position: { type: Object as PropType<{ x: number; y: number }>, required: true },
+  },
+  emits: {
+    'open-delete-slot-modal': () => true,
+    'open-edit-slot-modal': () => true,
+    'update:modelValue': (value: boolean) => typeof value === 'boolean',
+  },
+  methods: {
+    closeCurrentModal() {
+      this.$emit('update:modelValue', false);
```
willespencer, line 55: Instead of using an update:modelValue emit, can you emit something like 'close-edit-slot-menu'?

src/components/Modals/TeleportModal.vue
willespencer: Would also be useful to add them to the notion writeup! https://www.notion.so/courseplan/Modal-Refactor-fbbea3dfda52468db4719f53df7d9464

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -99,6 +102,20 @@ export default defineComponent({
         } else deleteCourseFromSemesters(this.courseTaken.uniqueId, this.$gtag);
       }
     },
+    openSlotMenu(e: MouseEvent) {
+      console.log('window', window);
+      this.position =
+        window.innerWidth > 863
```
willespencer, line 108: Is this what determines if the menu should go left or right? If so, could you make a notion item to apply this same code to our edit course modal? Right now the code that determines which way it should go doesn't work very well

willespencer (changes requested):
Left some small comments, mostly about modals since I just refactored them a little bit (sorry). Otherwise, this looks and works great! 😄

src/components/Modals/TeleportModal.vue
hahnbeelee: done!

src/components/Modals/SlotMenu.vue
```diff
@@ -42,25 +41,24 @@ import TeleportModal from '@/components/Modals/TeleportModal.vue';
...
   },
   methods: {
     closeCurrentModal() {
-      this.$emit('update:modelValue', false);
+      this.$emit('close-slot-menu', false);
     },
     openDeleteSlotModal() {
       this.$emit('open-delete-slot-modal');
-      this.$emit('update:modelValue', false);
+      this.$emit('close-slot-menu', false);
```
willespencer, line 57: One thing we do in a bunch of other modals is call closeCurrentModal() from other functions that need to close the modal instead of duplicating the emit. Nit but might be good to be consistent and avoid calling the emit so many times!

willespencer (approved):
All my comments are addressed, thank you! Also appreciate the notion updates 😀 Just left a small nit but good to go on my end!

tcho6319 (approved):
LGTM!

PR #512, opened 2021-05-17
WIP Override AP/IB frontend

hahnbeelee:
**Disclaimer:** Feel free to review although this is a WIP. It's missing the final functionality, but the code and frontend can all be critiqued.

### Summary
This PR implements the override requirement with AP/IB credit functionality.

image

- [x] The dropdown shows the user's AP/IB exams.
- [x] Updates firebase 
( The CS 2800 in the picture below was added by calling the function/ using the frontend)
image
- [ ] Accurately show the overridden requirement.
The function updates the firestore, but the requirements bar doesn't correctly update after pressing "Override". @benjamin-shen I was wondering if you could help look into what this issue is with this?

### Test Plan
- Nit @ frontend
- Lowkey nothing else till we figure out why the req bar isn't updating properly.

### Notes 
Questions for @einc :
- What happens if the course is already fulfilled by transfer credit?
- What happens if the user has no transfer credits?


hahnbeelee:
@benjamin-shen also for some reason it isn't recognizing `derivedAPIBEquivalentCourseData` as a property of `store.state` for me

image

PR #524, opened 2021-09-24
Standardize CI node version and use npm cache

SamChou19815:
### Summary 

Let's standardize the node version we use on CI. This updates the `setup-node` action to v2 and uses the node on CI by default (v14). `setup-node` action also allows you to cache `node_modules` now.

See https://github.com/actions/setup-node#caching-packages-dependencies

### Test Plan 

It's first time running it with cache, so of course cache doesn't exist yet... So check with 👀

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -10,7 +10,9 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - name: Set up Node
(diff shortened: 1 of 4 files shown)


SamChou19815:
Not sure why cypress tests are failing... Debugging.

hahnbeelee (approved):
LGTM

PR #527, opened 2021-09-27
Migrate build system to vite for improved build speed

SamChou19815:
### Summary 

This PR switches the build system to [Vite](vitejs.dev) for much faster speed of bundling and compiling source files. Vite uses native code to transpile JS/TS code so it's much faster than babel.

Changelist:

- environment variables now should be prefixed with `VITE_`, and referenced by `import.meta.env.VITE_xxx` (test transpiler needs to be slightly tweaked to support that)
- css importing assets should start with `@/` instead of `~@/`
- No `require` in JS code. `import` only
- [no support for scss exports](https://github.com/vitejs/vite/issues/1279). I have to create a separate TS file 😭 

### Test Plan 

CI and preview site. Nothing should change.


SamChou19815:
Compared to build job in #526, the `build` step' time drops from 1m to 45s.

src/shims-vue.d.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="vite/client" />
```
hahnbeelee, line 1: is this supposed to be here?

src/firebaseConfig.ts
```diff
@@ -5,7 +5,7 @@ import 'firebase/firestore';
 import 'firebase/functions';
 
 let config;
-if (process.env.VUE_APP_FIREBASE_MODE === 'prod') {
+if (import.meta.env.VITE_SOME_KEY === 'prod') {
```
hahnbeelee, line 8: is this supposed to be SOME_KEY ?

SamChou19815: Fixed

src/shims-vue.d.ts
SamChou19815: Yes, it's necessary for things like `import.meta.env`

SamChou19815:
After latest commit, build time drops to 18s!

hahnbeelee (approved):
LGTM

PR #529, opened 2021-09-28
Adding empty semesters from entrance to grad semester for first-time users

vaishnavi17:
### Summary 

This PR implements functionality to pre-add empty semesters for first-time users, based on their entrance and graduation periods. Currently, since we only collect entrance and grad years, the default assumption is that the user is entering in the Fall and graduating in Spring.

I use the "isEditingProfile" boolean in `Onboarding.vue` to detect whether this is a first-time onboarding. If the boolean is false, I call the populateSemesters() function in `global-firestore-data.ts` through `Onboarding.vue`. This is a clean way, since it doesn't involve adding an unnecessary boolean to `AppOnboardingData`, or making additional Firestore calls.

- [x] Updated `AppOnboardingData` to have optional entrance/grad sem fields
- [x] Added populateSemesters() function in `global-firestore-data.ts`

TODOs for future PR:

- [ ] Collect entrance/grad semester info in onboarding and store it in AppOnboardingData
- [ ] Store entrance/grad semester in Firestore onboarding data?

### Test Plan 
New user scenario (semesters auto-added):

Old semesters preserved when old user is going through onboarding again:

### Notes 
Currently, I only modified the AppOnboardingData type to store optional entrance and grad semesters. We could also modify the firestore onboarding data type to store this info in Firebase, but I don't think this information is being used anywhere else, so it might be unnecessary.

### DIFF
--- a/src/global-firestore-data.ts
+++ b/src/global-firestore-data.ts
@@ -80,6 +80,27 @@ export const addSemester = (
   );
 };
 
+// function to add all the semesters a user will be enrolled in based on entrance and graduation time
+export const populateSemesters = (onboarding: AppOnboardingData): void => {
+  const entranceYear = parseInt(onboarding.entranceYear, 10);
+  const gradYear = parseInt(onboarding.gradYear, 10);
+
+  const entranceSem: FirestoreSemesterType = onboarding.entranceSem
+    ? onboarding.entranceSem
+    : 'Fall';
+  const gradSem: FirestoreSemesterType = onboarding.gradSem ? onboarding.gradSem : 'Spring';
+
+  const sems = [createSemester('Fall', entranceYear, []), createSemester('Spring', gradYear, [])];
+  if (entranceSem === 'Spring') sems.push(createSemester('Spring', entranceYear, []));
+  if (gradSem === 'Fall') sems.push(createSemester('Fall', gradYear, []));
+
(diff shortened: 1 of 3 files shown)


src/global-firestore-data.ts
```diff
@@ -80,6 +80,27 @@ export const addSemester = (
...
+  const sems = [createSemester('Fall', entranceYear, []), createSemester('Spring', gradYear, [])];
+  if (entranceSem === 'Spring') sems.push(createSemester('Spring', entranceYear, []));
+  if (gradSem === 'Fall') sems.push(createSemester('Fall', gradYear, []));
+
+  for (let yr = entranceYear + 1; yr < gradYear; yr += 1) {
+    sems.push(createSemester('Spring', yr, []));
+    sems.push(createSemester('Fall', yr, []));
+  }
+  editSemesters(() => sems.sort(compareFirestoreSemesters));
+};
```
SamChou19815, line 102: Seems like there are a lot of edge case logic here. Probably good to factor out all but the last line into a pure function and unit test that.

hahnbeelee:
I like this a lot! Looks good to me.
image

src/global-firestore-data.ts
vaishnavi17: Did that, thanks!

PR #531, opened 2021-09-29
Setup typed firebase admin database collections

SamChou19815:
### Summary 

To make the infrastructure to support requirement-graph-aware data migrations, we first need a way to access the database by `firebase-admin`, ideally in a typed way. This PR sets up such files using firestore API's `typeConverter`. I will explain the details in comments below.

### Test Plan 

Everything compiles. Things run fine on CI.

### DIFF
--- a/src/firebase-config-common.ts
+++ b/src/firebase-config-common.ts
@@ -1,0 +1,19 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type CommonDocumentData = { [field: string]: any };
+/** An interface for `FirestoreDataConverter` common to both frontend firebase and admin firebase. */
+export interface CommonFirestoreDataConverter<T> {
+  toFirestore(modelObject: T): CommonDocumentData;
+  fromFirestore(snapshot: { data(): CommonDocumentData }): T;
+}
+
+export const getTypedFirestoreDataConverter = <T>(): CommonFirestoreDataConverter<T> => ({
+  fromFirestore(snapshot) {
+    return snapshot.data() as T;
+  },
+  toFirestore(userData) {
+    return userData;
+  },
+});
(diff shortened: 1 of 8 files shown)


src/firebase-config-common.ts
```diff
@@ -0,0 +1,19 @@
...
+  fromFirestore(snapshot: { data(): CommonDocumentData }): T;
+}
+
+export const getTypedFirestoreDataConverter = <T>(): CommonFirestoreDataConverter<T> => ({
+  fromFirestore(snapshot) {
+    return snapshot.data() as T;
+  },
+  toFirestore(userData) {
+    return userData;
+  },
```
SamChou19815, line 15: If you look at the old code, all the type converters look exactly the same. In fact, they are completely the same when the TS code is type erased to JS. Therefore, we can factor them out into a helper constructor, and use the passed-in generic type parameter to figure the type out!

src/firebase-config-common.ts
```diff
@@ -0,0 +1,19 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type CommonDocumentData = { [field: string]: any };
+/** An interface for `FirestoreDataConverter` common to both frontend firebase and admin firebase. */
+export interface CommonFirestoreDataConverter<T> {
+  toFirestore(modelObject: T): CommonDocumentData;
+  fromFirestore(snapshot: { data(): CommonDocumentData }): T;
+}
```
SamChou19815, line 7: Firestore has almost the exact same TS typedef on frontend and in admin, but they are slightly different so it's not structurally the same. Therefore, I have to exact this minimal interface that satisfy both.

hahnbeelee (approved):
LGTM

Issue #522, opened 2021-09-17
Rethinking Requirement Computation and UX


SamChou19815:
_Paradigm Shift: From pre-computation invariant enforcement to post-computation violation fixing_.

## Rationale

In a recent team meeting, we discussed how requirement computation's complexity has exploded. We have several pressing flaws in our current system:

1. lack of opt-in/out-out mechanism;
2. broken double-counting mechanism for minors.

Given our current design/UX philosophy, we need to force user to make some choices so that these problems are impossible to occur after a course is added. This can be confusing for the user, since they might just want to mass add a lot of courses to see what's going on. This is also difficult for us, since we have to design modals for all these choices in all possible flows.

This design document will present an alternative way to solve the problem: we shift the burden of producing a valid requirement graph from the pre-computation stage to the post-computation stage. Intuitively, it can be described by the following graph:

<img width="606" alt="intuition" src="https://user-images.githubusercontent.com/4290500/133834873-a67d32b0-d656-4d6c-8ef3-07d31975fb89.png">

I believe this new general model of thinking can help both users and developer understand what's going on:

- For developers, instead of predicting what problems can be caused by adding a class before adding it, we can directly compute the list of problems on the computed requirement graph.
- For users, instead of making choices they have no idea why they have to be made, they can see the problem and have the problem in mind to make their choices.

Finally, it can also unblock another planned feature: mass adding courses, since users don't have to make choices at adding time any more!

### Correctness Theorem

The correctness of the new flow rests upon a single theorem:

> For a graph `G = (V, E)` produced from user input with possible constraint violations, there exists a graph `G_correct = (V, E')` such that `E'` is a subset of `E`, and `G_correct` does not have any constraint violations.

In plain English, it states that starting from any bad graph, we can remove edges until it becomes good. (In the current system, being good means no illegal double-counting.)

To prove the theorem, we need to first define constraint:

> A constraint is a tuple `(c, R)` where `c` is a course and `R` is a set of requirement. The constraint mandates that there can be at most one edge `(r, c)` in the graph such that `r in R`.

With this definition of constraint, the theorem can be easily proven: for every violation of constraint, we kept only one edge. So the resulting graph will be valid, has the same set of originasl vertices but a subset of original edges.

The proof of the theorem also tells us how to guide user to have a valid requirement graph, which I will talk about more in the design/UX section.

### Double Counting as Constraint Violations

[...]

PR #532, opened 2021-09-30
Assign string uniqueId to AP/IB equivalent courses

benjamin-shen:
### Summary 

This pull request adds a stable (string) uniqueId to AP/IB equivalent courses. Rationale (optional reading after the first bullet point):
- Constraints are defined by `(c,R)` tuples. The course `c` refers to the `uniqueId`, not the `courseId` (i.e., different instances of the same course can have different constraint violations). Therefore, for AP/IB credit to be compatible with constraints, they must have a stable `uniqueId`.
- Why don't we just assign `uniqueId` regularly (using and incrementing the numeric `uniqueId`)?
  - The user can't have a single exam multiple times (unlike courses), thus a concept of an exam-specific unique ID is not as important as a course-specific unique ID. In other words, the exam name *can be* the exam's unique ID.
  - If we change the AP/IB data (eg. an exam can fulfill an extra course), or if the users modify their AP/IB exams, the equivalent course unique ids don't need to change. If they did, we would need to store them in the database (as special courses or maybe alongside the exams) and update them every time -- but at that point, we can just use the exam name to the same effect.
  - Multiple plans would be easier to implement (AP/IB exams should transcend all the plans, so a special non-numeric `uniqueId` makes sense).
  - With the numeric approach, the equivalent course unique IDs become as stable as regular courses. With the string approach, the exam unique IDs become as stable as requirement IDs (which is stable enough for user choice and the requirements algorithm with constraints).
- Why don't we just use the current method of assigning `uniqueId` (decrementing numbers, `uniqueId < -1`) and store them in the database?
  - This is technically feasible, but making `uniqueId` stable is a bit more complex than that. It would require:
    - requirements algorithm changes: change AP/IB equivalent courses from being recomputed every time to being read from the database
    - frontend changes: compute the equivalent courses and store them every time AP/IB exams are modified
    - migration script: generate the data for everyone's current AP/IB equivalent courses
    - new `uniqueDecrementer` field and function: like `uniqueIncrementer`
  - There are then a couple of downsides (in my imagined implementation):
    - *Imagine taking exams with the following equivalent courses: `{e1: [(c1, -2), (c2, -3)], e2: [(c2, -4), (c3, -5)]}`. Then, modifying those exams (or even just clicking through the onboarding modal) would require a full comparison with all of the stored exams. If there is any difference, the differing exams would have all their equivalent courses updated. For example, if `e1` changed score:  `{e1: [(c1, -6), (c2, -7), (c3, -8)], e2: [(c2, -4), (c3, -5)]}`
    - Every AP/IB exam modification would be linear in time complexity, which is bad for Firestore quotas
    - The `uniqueDecrementer` can potentially blow up really quickly
    - Updating (`e1: [(c1, -2), (c2, -3)]` to `e1: [(c1, -6), (c2, -7), (c3, -8)]`) is not obvious and kind of defeats the purpose of `uniqueId`
    - AP/IB data changes would require migration scripts for the equivalent courses collection but *also* every single user choice collection.
    - Multiple plans would be harder to implement
- Why don't we just assign `uniqueId = -1` and update the requirements algorithm to check the course code / exam name if `uniqueId < 0`?
  - I would personally rather shift away from having `-1` be a special exception id, in case the data explodes in complexity (again) and we need the `uniqueId` to be distinguished from the course code / exam name.
  - The `uniqueId` wouldn't be very unique. A decent chunk of the requirements algorithm would need to change.
- Why don't we just assign `uniqueId = -2` and have a constant mapping from negative unique ID to special category of courses? Eg. `{-1: 'SWIM', -2: 'AP', -3: 'IB', -4: 'CASE', -5: 'TRANSFER'}`, etc. And then check the course code / exam name?
  - This has a reasonable thought process. But, as with the last suggestion, the `uniqueId` would inherently be *not* unique. If we want to do this, we should introduce an additional field.

🌵 

- [x] changed `uniqueId` type from `number` to `string | number` in entire codebase, kept current behavior the same
- [x] changed overridden requirements key type from `number` to `string`
- [x] assigned exam name (eg. `AP Computer Science A`) to exam's equivalent course(s)' `uniqueId`s. 

Remaining TODOs:

- [ ] change AP/IB underlying course equivalent infra to associate each AP/IB exam with a single AP/IB course
- [ ] snapshot test to make sure AP/IB exam names don't change in the data unexpectedly (analogous to requirement IDs)
- [ ] (possibly) move Swim Test logic to `uniqueId = 'Swim Test'` instead of `uniqueId = -1`

### Test Plan 

All existing behavior related to `uniqueId` (including adding courses and requirements / user choice) should be the same -- especially for AP/IB exam equivalent course fulfillments.

I tested AP/IB optIn (as it's currently implemented) and it worked. This is not strictly part of the test plan, but you can play with it if you'd like.
image

### Notes
To make the `uniqueId` truly unique, my first TODO point (change AP/IB underlying course equivalent infra) is important. Currently, multiple equivalent courses can have the same `uniqueId`, which would make it difficult for a user choice to split up the equivalent courses for a single exam. For example, if an exam `e` generates an equivalent courses list `[c1, c2, credits]`, we don't necessarily want a user choice `(e, r)` to apply all courses `[c1, c2, credits]` to requirement `r`. Instead, exam `e` should have a single equivalent course `c`, and a user choice `(e, r)` would apply a single course `c` to requirement `r`. This should ideally be implemented before overridden fulfillments and constraints are implemented; I will explore this topic next.

After changing the AP/IB underlying course equivalent infra, it may turn out to be better to assign numeric `uniqueId`s. Even if that is the case, I think it'd still be good to have string `uniqueId`s allowed in the code.


willespencer:
Rationale for this change makes sense to me! Appreciate the thorough summary and considering every other case.

SamChou19815 (changes requested):
Changes look reasonable to me, but it seems to cause all the AP credits to no longer fulfill requirements:

Screen Shot 2021-10-01 at 14 36 24

benjamin-shen:
@SamChou19815 thank you for the catch. After some investigation, I discovered some interesting bugs.

I fixed it so it has the previous behavior (don't remove edges from the graph if the course is swim test or AP/IB).

Staging:
image

Dev branch before:
image

Dev branch after (fixed):
image

However, I don't know if the current behavior is what we want to keep:  
Adding CS 1110 automatically removes the edges that violate the double counting constraints. In fact, it looks exactly like dev branch before. You can replicate this on staging.
image

This is because the Computing requirement (and the other two requirements besides the Intro Programming requirement for which the edges got removed) are not double countable. Related code is in [graph building phase 3](https://github.com/cornell-dti/course-plan/pull/532/files#diff-ce531e7a7ae3563410fc747e501e2c64cb1fcb9196cfd25468710df01ac50e9bR114-R131).

This is certainly part of the conversation of how we will want to handle constraint violations after #522 is implemented. Actually, with the introduction of constraints, graph building phase 3 will be completely changed; modifying edges for double counting will be constraint violation resolution user choices. I think truly targeting the root of the issue above (i.e., CS 1110 not counting for Computing, and APCS counting because of a hack) will be much easier after constraints are implemented.

SamChou19815 (approved):
LGTM

PR #533, opened 2021-09-30
Unbreak master branch

SamChou19815:
### Summary 

Fix a bad merge between #527 and #531.

### Test Plan 

Previously it breaks local development. Now it should be fixed.


benjamin-shen (approved):
Thanks 😀

PR #536, opened 2021-10-03
Minimize semester

bungaepyo:
### Summary 

This pull request implements the minimize/expand semester feature. Users can minimize the semesters they want to hide and are able to expand them later at will. Added a minimize/expand button next to the three dots button in order to control the behavior.

- [x] added svg assets
- [x] implemented minimize/hide feature

### Test Plan 

Screen Shot 2021-10-03 at 12 31 10 PM

Screen Shot 2021-10-03 at 12 32 28 PM

### Notes 

Might need to clarify if we want the minimize/hide to persist when users return later.


hahnbeelee (changes requested):
It seems that when you....
1. Minimize a semester
2. Delete it
It doesn't delete.

hahnbeelee:
omg i deleted a semester and it didn't delete but it did this 😳
image

hahnbeelee:
I don't think the "Edit Semester" button works either when it's minimized

benjamin-shen:
@hahnbeelee I can't reproduce your issues. Delete & Edit Semester works for me when minimized on the deployed site, maybe do you have something funky going on locally?

hahnbeelee:
@benjamin-shen thanks ben~! you right.

hahnbeelee:
Also make sure to name your PR appropriately to be more descriptive of what the PR is doing! You can look at the other PRs that are opened right now as an example.

SamChou19815:
Screen Shot 2021-10-03 at 21 54 54

The error is legit on dev. It's actually also silently failing in preview with the same error in console but somehow it does not block the operation.

The root cause is that the DOM nodes for droppable (courses) got nuked when semester is minimized. I think we can prevent that by using `display: none` to hide instead of eliminating them.

src/components/Semester/Semester.vue
```diff
@@ -45,12 +45,20 @@
...
+            <img v-else src="@/assets/images/expand.svg" alt="expand semester" />
+          </button>
           <button class="semester-dotRow" @click="openSemesterMenu" data-cyId="semesterMenu">
             <img src="@/assets/images/dots/threeDots.svg" alt="open menu for semester" />
           </button>
         </div>
       </div>
-      <div class="semester-courses">
+      <div v-if="!isSemesterMinimized" class="semester-courses">
         <draggable
```
benjamin-shen, line 62: Looks like this conditional rendering of `<draggable>` is causing `beforeUnmount()` to error out (check the console when deleting/editing a minimized semester). I think it's not showing up on the deployed site because the warnings are getting suppressed or smth. @hahnbeelee u weren't buggin 🐛

SamChou19815 (approved):
LGTM

willespencer (approved):
> omg i deleted a semester and it didn't delete but it did this 😳 (The New Semester button) image

I also tested this and it worked fine for me - but it could be useful to write a Cypress test to check that only one of the add semester buttons shows up at any time, even after minimizing and deleting a semester (since we have 2 different buttons that show depending on the # of semesters). Lmk if you would want to write one @bungaepyo and I could help, or I could add one myself!

PR #539, opened 2021-10-07
Use esbuild to bundle firebase functions

SamChou19815:
### Summary 

This PR adds a build step to firebase functions instead of using the old approach of using JS only without any build step. This allows us to eventually switch to typescript for writing these functions.

I decided to use [esbuild](https://esbuild.github.io/) to bundle it mostly for its speed and ease of configuration. You can also use webpack for bundling, but that requires a lot of config. 
In addition, I decided to _bundle_ instead of simply _transpile_ since I have future plan to organize our repo into workspaces (see [npm workspaces (v7 only)](https://classic.yarnpkg.com/lang/en/docs/workspaces/) or [yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/)). Unfortunately, [firebase tools do not work well with with these workspaces](https://github.com/firebase/firebase-tools/issues/653), so we have to bundle the code into a single, self-contained js file without any dependencies, so that firebase tools will not look into `node_modules`. The bundled code is put inside `functions/deployment-template` with a package.json without dependencies.

### Test Plan 

Function should be invoked fine (unfortunately I don't have access to the google cloud project)


hahnbeelee:
hmm looks like i don't have permissions either. @willespencer can you review this?
image

willespencer (commented):
image

When I run the deploy command, I get the following error? Not sure what could be causing this

SamChou19815:
> image
> 
> When I run the deploy command, I get the following error? Not sure what could be causing this

Looks like a windows issue. Are you running this in WSL or normal terminals?

hahnbeelee (changes requested):
OK, so I was able to run `npm run deploy` but when I tried invoking `TrackUsers` [here](https://console.cloud.google.com/functions/details/us-central1/TrackUsers?project=cornelldti-courseplan-dev&tab=testing) I got this error:
image

SamChou19815:
> OK, so I was able to run `npm run deploy` but when I tried invoking `TrackUsers` [here](https://console.cloud.google.com/functions/details/us-central1/TrackUsers?project=cornelldti-courseplan-dev&tab=testing) I got this error: image

I might know what's happening. I think bumping `firebase-admin` might help. It's issue with esbuild is only fixed recently: https://github.com/googleapis/nodejs-firestore/issues/1493

SamChou19815:
@hahnbeelee I bumped it and tried to deploy it, but still could not test. Can you help to run it?

PR #540, opened 2021-10-07
Implement clear semester feature

bungaepyo:
### Summary 

This pull request implements the clear semester feature. Users would be able to get rid of all the courses that were added to a specific semester by clicking with the "Clear Semester" button. The button is located within the three-dot menu, between the "Edit Semester" and the "Delete Semester" buttons. I've also added a confirmation modal and a message for optimal user flow.

- [x] added svg assets
- [x] implemented clear semester feature

### Test Plan 
Screen Shot 2021-10-07 at 1 20 23 AM
Screen Shot 2021-10-07 at 1 20 33 AM
Screen Shot 2021-10-07 at 1 20 42 AM
Screen Shot 2021-10-07 at 1 21 05 AM

### DIFF
--- a/src/components/Modals/SemesterMenu.vue
+++ b/src/components/Modals/SemesterMenu.vue
@@ -64,24 +64,23 @@
 import { defineComponent } from 'vue';
 
 export default defineComponent({
-  emits: ['open-delete-semester-modal', 'open-edit-semester-modal', 'open-clear-semester-modal'],
-  data() {
-    return {
-      isOpenModal: false,
-    };
+  emits: {
+    'open-delete-semester-modal': () => true,
+    'open-edit-semester-modal': () => true,
+    'open-clear-semester-modal': () => true,
+  },
+  props: {
+    isOpenModal: { type: Boolean, required: true },
   },
   methods: {
     openDeleteSemesterModal() {
       this.$emit('open-delete-semester-modal');
-      this.isOpenModal = true;
     },
     openEditSemesterModal() {
       this.$emit('open-edit-semester-modal');
-      this.isOpenModal = true;
     },
     openClearSemesterModal() {
       this.$emit('open-clear-semester-modal');
-      this.isOpenModal = true;
     },
   },
 });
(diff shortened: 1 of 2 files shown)


src/components/Modals/SemesterMenu.vue
```diff
@@ -19,6 +19,25 @@
...
+            <img
+              class="semesterMenu-icon"
+              src="@/assets/images/erase.svg"
+              alt="clear semester erase icon"
+            />
+            <span class="semesterMenu-edit">Clear Semester</span>
+          </div>
+        </div>
+      </div>
+    </button>
```
zachary0kent, line 40: Do you think it would be possible to factor out some of the repeated code for each modal into a separate component? The same button template seems to be similar for each of the edit, clear, and delete modals.

src/components/Modals/SemesterMenu.vue
```diff
@@ -45,7 +64,7 @@
 import { defineComponent } from 'vue';
 
 export default defineComponent({
-  emits: ['open-delete-semester-modal', 'open-edit-semester-modal'],
+  emits: ['open-delete-semester-modal', 'open-edit-semester-modal', 'open-clear-semester-modal'],
   data() {
     return {
       isOpenModal: false,
```
zachary0kent, line 70: I'm not sure if this data field is really needed--isDeleteSemesterOpen, isEditSemesterOpen, and isClearSemesterOpen are all stored in the data of the Semester component, so isOpenModal could be passed into SemesterMenu as as a prop such that isOpenModal = isDeleteSemesterOpen || isEditSemesterOpen || isClearSemesterOpen. It's probably best to minimize use of mutable state.

src/components/Modals/SemesterMenu.vue
```diff
@@ -118,5 +141,8 @@ export default defineComponent({
   &-edit {
     color: black;
   }
+  &-clear {
+    color: black;
+  }
```
zachary0kent, line 146: Could probably merge these into one rule.

PR #541, opened 2021-10-07
⚡️ Speed up type checking

SamChou19815:
### Summary 

This PR aims to speed up type checking on command line to greatly improve dev experience. On my machine, it reduces the total time from 50s to 13s.

It contains two main changes:
- Disable `resolveJsonModule` in tsconfig and replaces it with manual type definition `d.ts` files for those json. Our json is big and TS spends a lot of time type checking them. In the end, TypeScript still cannot correctly decides its type due to our usage of discriminated union in the json. Therefore, it does not provide any type safety. It's better to just declare their types manually and make typescript skip checking its content. This change speeds up TS files checking from 5s to 3s on my machine.
- Replace `vti` with `vue-tsc`. `vue-tsc` comes with vite's template, and it's much faster than `vti`. From now on, we should use [this plugin](https://github.com/johnsoncodehk/volar) instead of `Vetur`. This provides most of the speedup. It seems to be stricter than `vti` so I think it's a good thing.

### Test Plan 

CI.

I also tested manually to add some type error to both ts files and vue files like `const a: number = 'dddd'`. Both errors can be caught.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "req-graph-dump": "ts-node -T -P tsconfig.node.json src/requirements/requirement-graph-dumper.ts",
     "lint": "eslint . --cache",
     "tsc": "tsc",
-    "type-check": "tsc && vti diagnostics",
+    "type-check": "vue-tsc --noEmit",
     "build:staging": "vite build --mode staging",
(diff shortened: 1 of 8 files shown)


hahnbeelee (approved):
LGTM!

PR #542, opened 2021-10-07
Script to migrate engineering liberal studies data (dry-run-only)

SamChou19815:
### Summary 

This PR adds a script to migrate engineering liberal studies data so that we can switch it to not allow double counting. It works in the following flow:

1. Find all courses that are connected to this engineering liberal studies requirement.
2. For each course, find requirements that connects to it
  a. Check if there all requirements connected to it are either
    - liberal study requirement
    - allow double counting
  b. If the above check returns true, it means that user has not chosen other requirement, then we can bind it to liberal study requirement (sort of)
3. For all such courses found above, select liberal study as requirement, except that
  - The user might already made a choice on it. It happens maybe when the old requirement data is not the same as the new one, or maybe due to override???
  - In those cases, we skip and print a warning.

### Test Plan 

`npm run ts-node -- src/requirements/admin/migrate-engineering-liberal-arts-data.ts <your email>` to run on one user only.

`npm run ts-node -- src/requirements/admin/migrate-engineering-liberal-arts-data.ts` to run on all data.

In addition to the warning above, it also dies on some user on dev db that still has `AS` as their college while we only have `AS1` and `AS2` now. Not sure what's the story on prod data yet...


src/requirements/admin/migrate-engineering-liberal-arts-data.ts
```diff
@@ -0,0 +1,95 @@
+/* eslint-disable no-console */
+
+import { usernameCollection } from '../../firebase-admin-config';
+import { requirementAllowDoubleCounting } from '../requirement-frontend-utils';
+import getUserRequirementDataOnAdmin from './requirement-graph-admin-utils';
+
+const REQUIREMENT_ID = 'College-EN-Liberal Studies: 6 courses';
+
+async function runOnUser(userEmail: string) {
```
hahnbeelee, line 9: Can you add documentation for this function

src/requirements/admin/migrate-engineering-liberal-arts-data.ts
```diff
@@ -0,0 +1,95 @@
...
+    onboardingData,
+    userRequirements,
+    requirementFulfillmentGraph,
+  } = await getUserRequirementDataOnAdmin(userEmail);
+  const userRequirementMap = Object.fromEntries(userRequirements.map(it => [it.id, it]));
+
+  const connectedCourses = requirementFulfillmentGraph.getConnectedCoursesFromRequirement(
+    REQUIREMENT_ID
+  );
+  const coursesNotConnectedToOtherRequirementThatDoesNotAllowDoubleCounting = connectedCourses.filter(
```
hahnbeelee, line 21: Would a better name for this be `connectedCoursesOnlyConnectedToDoubleCountedReqs`?

src/requirements/admin/migrate-engineering-liberal-arts-data.ts
```diff
@@ -0,0 +1,95 @@
...
+    course => {
+      const connectedRequirementIDs = requirementFulfillmentGraph.getConnectedRequirementsFromCourse(
+        course
+      );
+      return connectedRequirementIDs.every(
+        id =>
+          id === REQUIREMENT_ID ||
```
hahnbeelee, line 28: Doesn't this mean that if the course is connected to `REQUIREMENT_ID` it'll get connected to it again?

hahnbeelee (changes requested):
Just requested some comments and asked some clarifying questions.

src/requirements/admin/migrate-engineering-liberal-arts-data.ts
SamChou19815: Added

PR #537, opened 2021-10-06
Sort by semester

zachary0kent:
### Summary 

This pull request implements a dropdown menu that allows the user to select whether their semesters are shown in order of newest to oldest or oldest to newest, persisting their choices in the backend. The dropdown menu also allows users to select whether semesters are displayed in compact or default layout, although this choice is not persistent. Implementing these changes required adding an orderByNewest field both to the user-semesters collection in Firebase and to the Vuex store. To minimize reads and writes to the Firebase, changing the ordering preference on the frontend only modifies the orderByNewest field in the Firebase. As a result, the semesters are no longer sorted in the backend; to ensure this does not introduce any breaking changes, when the orderByNewest field is changes, the semesters are resorted in the Vuex store--but not in the Firebase. Nevertheless, their still exists non-negligible latency when changing the ordering preference on the frontend, something to be explored in the future.

Remaining TODOs:

- [ ] Reduce order preference latency
- [ ] Fix down arrow asset to be more centered

### Test Plan 

Screen Shot 2021-10-05 at 10 57 39 PM

Screen Shot 2021-10-05 at 10 58 27 PM

Screen Shot 2021-10-05 at 10 58 48 PM

PR #314, opened 2021-02-21
Allow for clicks outside modal to close it

neketka:
### Summary 

Allowed for clicks outside the modals to be detected and close the modal. This was done through a basic rectangle-point intersection test using refs from the DOM elements.

- [x] Added modal closing to FlexibleModal
- [x] Added modal closing to Onboarding

Remaining TODOs:

- [ ] Make less hacky by reimplementing the modal system using vue-portal and maybe mixins for reuse of functionality (after launch)

### Test Plan 

Check that every modal has been covered by this functionality and that modals behave as expected (does not affect the ability to click other elements on the modal).

PR #123, opened 2020-05-09
Handle updated CUReviews json

willespencer:
### Summary 

Update JSON structure to handle new CUReviews data endpoints. All CUReviews data on courseplan.io currently displays as N/A.

- [x] updated CUReviews fetch to read from the slightly updated JSON

### Test Plan 

Verify that CUReviews data now loads in properly for a bunch of different classes.

PR #543, opened 2021-10-15
Cypress testing suite for hidden features / bugs

willespencer:
### Summary 

This pull request adds Cypress front-end tests to test a number of features & bugs that can pop up without devs noticing, especially ones that have happened in the past and could be repeated.

- [x] Added tests to run through Onboarding and Walkthrough modals as a new user (involves deleting the testing email's firestore data). The walkthrough modal has frequently been modified without noticing (#537) as it only shows up for new users, so having a test will ensure it does not break from a PR without knowing.
- [x] Added a test to check that CUReviews data is being returned and not 'N/A'. The API has changed in the past without notice, so having this test will let us know in the future if their data or API changes again (#123)
- [x] Added a test to confirm that the "+ New Semester" buttons are mutually exclusive. As brought up in #536, there are two different buttons depending on if there are semesters in a plan or not for styling reason. This test ensures that two buttons will never show up at once.
- [x] Added a test to ensure clicking outside modals works as expected, as this functionality often breaks. Also added a check to ensure onboarding cannot be clicked outside for new users (#314, #403)
- [x] Added a test to confirm the mobile navbar shows up on top of the plan, as there have been frequent z-index issues (#430)
- [x] Resolve tests failing because the CI does not have permissions to delete the test user's Firestore data. See solution explanation below

Remaining TODOs:

- [ ] Add more tests to address hidden features after discussion with other devs
- [ ] Look into doing something like configuring supportFile that run before every test for things that are re-used like login: https://docs.cypress.io/guides/references/configuration#Folders-Files

### Test Plan 

Confirm that the CI tests pass - this means that all the existing and newly written Cypress tests also pass

Additionally, change the URL used for accessing CUReviews data locally and confirm that Cypress test would fail were CUReviews API to break or change routes.
CUReviews Test

### Notes 

Resolving the Firestore permissions error took a long time - but the issue ended up being related to the service account secret on GitHub. @SamChou19815 suspects that it was in multiline form which could have been causing the issue of insufficient permissions on the CI. Resetting it fixed the issue, for both using Firestore calls in Cypress, and also when writing separate scripts to edit user data before running cypress (see individual commits).

This new spec file will delete the semester and onboarding collections for the test email, which could affect other tests if this new file does not run completely and these documents are not recreated.


SamChou19815 (approved):
O9vpKoN

cypress/integration/hidden-features.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  // confirm the final page is visible and the tour can be finished
+  cy.get('[data-cyId=tour]').should('be.visible');
+  cy.get('[data-cyId=tour-startButton]').click();
+});
+
+// Test to confirm that only one "+New Semester" button shows up at once
+// Due to front-end styling issues, there are two different buttons in different components
+// Only one should be visible at a time (from semesterView if there are 0 semesters, from semester otherwise)
+it('Test that only one semester button shows', () => {
```
benjamin-shen, line 118: Is it possible to have this test be perpetual (i.e. throughout the other tests, ensure the invariant holds) rather than for one specific set of steps?

cypress/integration/hidden-features.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+
+// Before running tests, delete Firestore data for the testing email (so the initial onboarding and walkthrough trigger)
+// Then start on landing page, log in to firebase, and visits the dashboard
+// Delete and log in occurs with TEST_UID and test email of the courseplan testing account using functions from the cypress-firebase package
+before('Delete test user data, then visit site and log in', () => {
+  // log the user in
+  cy.visit('localhost:8080/login');
+  cy.login(Cypress.env('TEST_UID'));
+
+  const TEST_EMAIL = 'courseplan.cornelldti.test@gmail.com';
```
benjamin-shen, line 14: 😍 
Should we refactor this Firestore reset and call it for every integration test?  
Or at least store `TEST_EMAIL` in a constants file?

benjamin-shen (approved):
YAY!!!!! amazing 🌟

hahnbeelee (approved):
This is truly amazing. I appreciate this so much! This is truly your baby. Looks amazing and I'm glad we have these reinsure these basic things don't break.

hahnbeelee:
^ also went through the entire test plan and everything worked as expected:
- tests ran perfectly
- changed the url for CUReviews & it broke (then i changed it back and it was back to normal)

cypress/integration/hidden-features.spec.ts
willespencer: Going to put TEST_EMAIL in a constants file now!

willespencer: I however do not think we want to do the Firestore reset in every integration test because then we have to go through Onboarding/walkthrough every time. It might make sense though to do a more simple data reset (instead of deleting data, set to only required/basic data) so that tests are not reliant on each other. Making an item for this!

cypress/integration/hidden-features.spec.ts
willespencer: Yes! I think I can add an `afterEach` hook to run it after every test that we have. It will require reworking things a bit, so the test only checks for buttons and does not actually click anything, and also needs to ensure that the site is in a state where this can be checked. Adding it to my other item!

benjamin-shen: > and also needs to ensure that the site is in a state where this can be checked.

I think this can just be done by making the invariant "there is max 1 button" instead of "there is exactly 1 button". So if no buttons exist, the test passes.

"There is exactly 1 button" is not really related to the bug itself and can be another test if we want.

PR #545, opened 2021-10-18
[easy] Add delete-semester-courses gtag event, alphabetize gtags

benjamin-shen:
### Summary 

This pull request adds the `delete-semester-courses` gtag event and alphabetizes the gtags in the code.

- [x] add `delete-semester-courses` gtag event and use it in `deleteAllCoursesFromSemester()`
- [x] alphabetize gtag ordering in `gtag.ts` to reduce merge conflicts

### Test Plan 

This is a low-risk change because it only touches analytics data. Devs can double check that I didn't miss any gtags and they are correctly ordered.

### Notes
`deleteAllCoursesFromSemester()` is currently not used.

### DIFF
--- a/src/global-firestore-data/semesters.ts
+++ b/src/global-firestore-data/semesters.ts
@@ -128,8 +128,12 @@ export const deleteCourseFromSemester = (
   deleteCourseFromSelectableRequirements(courseUniqueID);
 };
 
-export const deleteAllCoursesFromSemester = (season: FirestoreSemesterType, year: number): void => {
-  // TODO add gtag event
+export const deleteAllCoursesFromSemester = (
+  season: FirestoreSemesterType,
+  year: number,
+  gtag?: GTag
+): void => {
+  GTagEvent(gtag, 'delete-semester-courses');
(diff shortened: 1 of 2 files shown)


src/global-firestore-data/semesters.ts
```diff
@@ -128,8 +128,12 @@ export const deleteCourseFromSemester = (
   deleteCourseFromSelectableRequirements(courseUniqueID);
 };
 
-export const deleteAllCoursesFromSemester = (season: FirestoreSemesterType, year: number): void => {
-  // TODO add gtag event
+export const deleteAllCoursesFromSemester = (
```
willespencer, line 131: Since this function is not being used, can we make sure it gets called when this functionality is implemented (maybe by adding it to a notion item)?

willespencer (approved):
Looks good! Left 1 suggestion, but stuff is alphabetically ordered and I don't see any other changes! Is there a way to confirm this works before/after merge?

src/global-firestore-data/semesters.ts
benjamin-shen: It's currently being implemented by @bungaepyo, probably a notion item is unnecessary?

PR #546, opened 2021-10-19
Standardize semester type/season naming convention

benjamin-shen:
### Summary 

We used "semester type" before we were using TypeScript and it worked for a while, but now "type" can be very ambiguous -- meanwhile, "season" has no ambiguity and we've actually been using both terms for a while now. This pull request officially standardizes the naming convention by changing most instances of "semester type" to "semester season".

- [x] changed `FirestoreSemesterType` to `FirestoreSemesterSeason` in `user-data.d.ts`
- [x] changed as many instances of "semester type" to "semester season" that I could find and safely remove
- [x] standardized argument order (year then semester) in `global-firestore-data/semesters.ts`
- [x] simplified some functions in `global-firestore-data/semesters.ts`
- [x] cleaned up some util functions, clear semester functionality, and toggle semester order functionality

TODO
- [ ] [fully remove all instances of `type`](https://www.notion.so/courseplan/Remove-all-instances-of-semester-type-8788f8e840d04bc6b69f8f0cb0dfb070) after devs pull this PR. it needs to remain in some places for now; otherwise, it breaks staging

### Test Plan 

Verify that all semester-related functionality still works as intended.

### Notes 

If there are merge conflicts: accept your current changes, change "semester type" to "semester season" where appropriate, verify that the web app builds, and make sure to `npm run type-check`.

### DIFF
--- a/src/utilities.ts
+++ b/src/utilities.ts
@@ -19,7 +19,7 @@ export const sortedSemesters = (
   semesters: readonly FirestoreSemester[],
   orderByNewest = true
 ): readonly FirestoreSemester[] =>
-  [...semesters].sort((a, b) => {
+  semesters.slice().sort((a, b) => {
     // sort in increasing order iff orderByNewest is false, increasing otherwise
     const order = orderByNewest ? -1 : 1;
     const byYear = a.year - b.year;
--- a/src/store.ts
+++ b/src/store.ts
@@ -6,8 +6,8 @@ import computeGroupedRequirementFulfillmentReports from './requirements/requirem
 import RequirementFulfillmentGraph from './requirements/requirement-graph';
 import { createAppOnboardingData } from './user-data-converter';
 import {
-  getCurrentSeason,
   sortedSemesters,
+  getCurrentSeason,
   checkNotNull,
   getCurrentYear,
   allocateAllSubjectColor,


src/components/Modals/SelectSemester.vue
```diff
@@ -32,18 +32,18 @@
...
             >
               <img
-                :src="season[0]"
+                :src="s[0]"
                 class="selectSemester-dropdown-content-season"
-                :alt="`${season[1]} icon`"
+                :alt="`${s[1]} icon`"
               />
-              {{ season[1] }}
+              {{ s[1] }}
```
benjamin-shen, line 46: had to rename since `season` is now a prop name

src/components/Modals/SelectSemester.vue
```diff
@@ -215,20 +215,6 @@ export default defineComponent({
...
-        currentSeason = 'Winter';
-      } else if (currentMonth <= 4) {
-        currentSeason = 'Spring';
-      } else if (currentMonth <= 7) {
-        currentSeason = 'Summer';
-      } else {
-        currentSeason = 'Fall';
-      }
-      return currentSeason;
-    },
```
benjamin-shen, line 231: duplicate function in `@/utilities`

src/utilities.ts
```diff
@@ -6,22 +6,14 @@ export function checkNotNull<T>(value: T | null | undefined): T {
...
 }
 
-export function getCurrentType(): 'WI' | 'SP' | 'SU' | 'FA' {
-  const currentMonth = new Date().getMonth();
-  if (currentMonth === 0) return 'WI';
-  if (currentMonth <= 4) return 'SP';
-  if (currentMonth <= 7) return 'SU';
-  return 'FA';
-}
-
```
benjamin-shen, line 24: unused

hahnbeelee:
Just wanted to say - overall looks good to me. I want to check all the season semester functionality in depth but it's getting a little late so ill do it tmrw!

hahnbeelee (changes requested):
@vaishnavi17 's semester population doesn't seem to be working here?
All other semester functionality seems to be good.
image

benjamin-shen:
@hahnbeelee thanks for the catch. The bug existed because I hadn't merged Vaish's PR from master, it should be fixed now.

I intend on merging Toby's PR and then Zach's PR to my branch first and then pushing to master.

src/components/Modals/ClearSemester.vue
```diff
@@ -24,14 +24,9 @@ import eraseIcon from '@/assets/images/erase-white.svg';
...
   components: { TeleportModal },
-  props: {
-    clearSemType: { type: String, required: true },
-    clearSemYear: { type: Number, required: true },
-  },
   emits: {
     'close-clear-sem': () => true,
-    'clear-semester': (type: string, year: number) =>
-      typeof type === 'string' && typeof year === 'number',
+    'clear-semester': () => true,
```
benjamin-shen, line 29: These are redundant props because the `<clear-semester>` component is a child of the `<semester>` prop.

I think it was modeled after `delete-semester`, which does need a season & year passed in to delete the semester (because it is emitted to the grandparent `<semester-view>`). Instead, `clear-semester` is more similar to `add-course` and `delete-course`, despite what the naming suggests.

src/global-firestore-data/semesters.ts
```diff
@@ -1,31 +1,14 @@
...
-    return -1;
-  }
-  if (a.year < b.year) {
-    return 1;
-  }
-  if (SeasonOrdinal[a.type] < SeasonOrdinal[b.type]) {
-    return 1;
-  }
-  return -1;
-};
```
benjamin-shen, line 27: We can get rid of this; `sortedSemesters()` has the same purpose. The only difference is that we don't modify the array directly anymore (think `sorted()` vs `.sort()`, which is good to avoid side effects & when working with `readonly` arrays.

src/global-firestore-data/semesters.ts
```diff
@@ -37,61 +20,75 @@ const editSemesters = (
...
 /**
- * Toggles whether semesters are ordered by newest/oldest
- * @returns true iff semesters were previously ordered oldest -> newest,
- *          false otherwise
+ * Sets whether semesters are ordered by newest/oldest
  */
-export const toggleOrderByNewest = (): boolean => {
-  const toggled = !store.state.orderByNewest;
-  store.commit('setOrderByNewest', toggled);
+export const setOrderByNewest = (orderByNewest: boolean): void => {
```
benjamin-shen, line 25: Changed `toggle` (which set the negated boolean value) to `set` (which sets a specific boolean value) to simplify the code in the frontend component and unify the `global-firestore-data` functions (i.e. `setBlah(newValue)` or `editBlah(updater)`). It should have the same functionality.

Lmk if you have a different opinion and I can change this back!

src/utilities.ts
```diff
@@ -15,38 +15,30 @@ export const SeasonOrdinal = {
...
-    return order * (byYear === 0 ? SeasonOrdinal[a.type] - SeasonOrdinal[b.type] : byYear);
+    return order * (byYear === 0 ? SeasonOrdinal[a.season] - SeasonOrdinal[b.season] : byYear);
   });
 
 export function checkNotNull<T>(value: T | null | undefined): T {
   if (value == null) throw new Error();
   return value;
 }
 
-export default function getCurrentSeason(): FirestoreSemesterType {
```
benjamin-shen, line 34: No reason for this to be a default export

PR #548, opened 2021-10-19
[Easy] Rename to overridden fulfillments

benjamin-shen:
### Summary 

This pull request renames all instances of `overriden requirements`, `overridden reqs`, etc. to `overridden fulfillments`.

### Test Plan 

No functionality changed, just variable names. CI checks handle testing.

### DIFF
--- a/functions/rename-requirement-and-slot-name.js
+++ b/functions/rename-requirement-and-slot-name.js
@@ -10,7 +10,7 @@ admin.initializeApp({
 });
 
 const db = admin.firestore();
-const userOverriddenReqsCollection = db.collection('user-overridden-requirements');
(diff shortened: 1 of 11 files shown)


functions/rename-requirement-and-slot-name.js
```diff
@@ -10,7 +10,7 @@ admin.initializeApp({
 });
 
 const db = admin.firestore();
-const userOverriddenReqsCollection = db.collection('user-overridden-requirements');
+const userOverriddenFulfillmentsCollection = db.collection('user-overridden-fulfillments');
```
SamChou19815, line 13: This might be runtime change but if it's not released it should be OK

PR #550, opened 2021-10-19
Feature flag improvements and docs

SamChou19815:
### Summary 

Add documentation for feature flags and improve the code so that creating one is easy.

### Test Plan 

Feature flag still works.


src/feature-flags.ts
```diff
@@ -0,0 +1,36 @@
...
+export const isRequirementDebuggerEnabled = registerFeatureFlagChecker('RequirementDebugger');
+
+// Add your new feature flags above
+// Gate keeper runtime code below. Do not touch if you do not understand.
+
+/**
+ * @param flagName name of the feature flag. Use `PascalCase` to be callable in developer console.
+ * @returns a checker function to tell whether the feature flag is enabled.
+ */
+function registerFeatureFlagChecker(flagName: string): () => boolean {
```
benjamin-shen, line 15: Instead of `flagName: string` can we define a more specific type? (similar to how we do gtags)

For example:
```
type FlagType =
  | 'RequirementDebugger'; // shows or hides the Requirement Debugger button
...
function registerFeatureFlagChecker(flagName: FlagType): () => boolean { ...
```

src/feature-flags.ts
```diff
@@ -0,0 +1,36 @@
+/** All registered GK togglers. Do not manually add items to it. */
+const GateKeeperTogglers: Record<string, () => void> = {};
+
+// Add your new feature flags below
+
+export const isRequirementDebuggerEnabled = registerFeatureFlagChecker('RequirementDebugger');
```
benjamin-shen, line 6: I'm worried that assigning and using constants will be less scalable and searchable. Instead, would exporting a single function work?

For example, first register all the feature flag checkers. Then export something like `isFlagEnabled: flagName => boolean`, and in call it like `isFlagEnabled('RequirementDebugger')`.

src/feature-flags.ts
```diff
@@ -0,0 +1,45 @@
...
+const featureFlagCheckers: FeatureFlagCheckers = registerFeatureFlagChecker(
+  'RequirementDebugger'
+  /* 'AddYourFeatureFlagNameHere' */
+);
+export default featureFlagCheckers;
+
+// Add your new feature flags above
+// Gate keeper runtime code below. Do not touch if you do not understand.
+
+type FeatureFlagCheckers = Record<`is${FeatureFlagName}Enabled`, () => boolean>;
```
SamChou19815, line 15: Using the feature from TS 4.1: https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

PR #552, opened 2021-10-21
[RFC] Small release

SamChou19815:
### Summary 

Before we make a release that requires some data migration, let's batch together these small improvements to make the next release less risky and easier to review.

Only user visible changes:

- Minimize semesters.
- Clear semesters
- Sort semesters
- Disable scrolling for modals
- Generate semesters for new users.

### Test Plan 

👀


benjamin-shen (dismissed):

hahnbeelee (approved):
Ok other than this frontend issue I think everything functionality-wise is okay. I would rather push this and just maybe make a patch release for this bug independently because I want to get all the other features (as well as the courses) up there. I think it's a sacrifice I'm willing to take.
image

benjamin-shen (approved):

einc (approved):

willespencer (changes requested):
Most things work!

Tested a lot of new functionality: Clear Semester, Sort Semesters, Minimize Semesters, Scroll behind modals, Compound requirement progress bug, etc.

Tested other site features: Walkthrough for new users, Onboarding, Mobile, New Course modals, New semesters, Edit Semester, etc.

However, there seems to be a bug with AP/IB transfer credit counting for requirements. Multiple exams do not fulfill engineering requirements on this branch when they currently do on prod. @benjamin-shen was able to replicate this as well. Does anyone know what might have caused this?

image

SamChou19815:
> Most things work!
> 
> Tested a lot of new functionality: Clear Semester, Sort Semesters, Minimize Semesters, Scroll behind modals, Compound requirement progress bug, etc.
> 
> Tested other site features: Walkthrough for new users, Onboarding, Mobile, New Course modals, New semesters, Edit Semester, etc.
> 
> However, there seems to be a bug with AP/IB transfer credit counting for requirements. Multiple exams do not fulfill engineering requirements on this branch when they currently do on prod. @benjamin-shen was able to replicate this as well. Does anyone know what might have caused this?
> 
> image

I cannot repro this on my end. I'm originally suspecting it's caused by leftover of AP/IB opt-in/opt-out stuff (that's never shipped), but there isn't any data in the database that supports that hypothesis...

benjamin-shen:
@willespencer I think it's because the AP/IB code currently requires that you select a major -- I remember when we reproduced it, it was if the user only selected a college.

To debug this, I tried [`git-bisect`](https://git-scm.com/docs/git-bisect) but found that every commit (after & including the push to prod) was "bad". I then switched to the `beta-release` branch and found the same issue, so I had the same suspicion as Sam that it was a database issue. Since I also couldn't find any data that was out of place, I then suspected it must be reproducible on prod.

This is definitely not ideal for new users that are undecided in their major, but I think it's fine to push since it's the same behavior as prod (please double check though). I can make a bug fix PR later.

willespencer (approved):
It looks like @benjamin-shen is right - reproducible on prod if you do not have a major, which is probably why I never noticed the issue before starting onboarding from scratch on dev today. LGTM!

SamChou19815:
looks like we have enough approvals, so i will merge now

PR #553, opened 2021-10-22
[RFC][data-migration] `selectableRequirementChoices` -> `optOut`

SamChou19815:
### Summary 

This PR implements the data migration script that will migrate the `selectableRequirementChoices` into the new format where all the user choices are unified under the concept of override: opt-in and opt-out.

We originally think everything can be categorized into opt-out and opt-in, but the running result shown above indicates that we might need 3, which I propose below:

```ts
type OverrideChoiceForCourse = {
  /* unchanged */
  optOut: Requirement[];
  /**
   * same as opt-in as we discussed in previous meetings. 
   * It's for attaching completely unknown courses to a requirement 
   * (e.g. opt-in CS 2112 for history requirement)
   */
  arbitraryOptIn: Record<Requirement, slot[]>
  /* See discussion below */
  acknowledgedCheckerWarningOptIn: Requirement[];
}
```

`acknowledgedCheckerWarningOptIn` is for requirements that have a checker warning. e.g. tech elective in CS. In the old flow, they are considered more like self-checks, and never picked as default choice. For example, when you add CS 4120, the following thing will happen in the old system:

- CS elective chosen as default choice
- Allow you to choose tech elective, advisor approved elective, etc (those with broad, imprecise checker) as alternative choice, but never selected by default
- opt-in not implemented yet, so that's irrelevant

For the new system and the 3 category thing I proposed, the default choice would be:

```ts
choice = { optOut: [], arbitraryOptIn: {}, acknowledgedCheckerWarningOptIn: null };
```

which corresponds to `CS 4120` being connected to `CS elective` only and not to any broad requirements. It fits the general philosophy of the new flow: during add time, we don't make complicated computation to decide eligibility, and the default choice should be safe, non-confusing.

It's not only satisfying philosophically, but it's also necessary. It's important to distinguish between arbitrary opt-in and `acknowledgedCheckerWarningOptIn`, since the former is a complete escape hatch and the system knows nothing about where it should be without user intervention, the latter we do exactly know which slot it should go to (we are just unsure whether it can be safely assigned to a requirement)

- [x] Dry-run proof of concept
- [x] Dry-run consolidated
- [ ] Runnable on staging site
- [ ] Run on staging site
- [ ] Run on production site

### Test Plan 

TODO

### Notes 

Once the data migration script is completed, it should not be run and deployed to production immediately. We need a second part of this migration project, which makes the frontend saves data to both `selectableRequirementChoices` and `optOut` while still making `selectableRequirementChoices` the source of truth, so that we can safely deploy this to prod and the prod will start consistency generating data of the new format. Then in a future deploy to prod when we are ready, we can switch to use the new format to be the source of truth.

### DIFF
--- a/src/requirements/admin/opt-out-data-migration.ts
+++ b/src/requirements/admin/opt-out-data-migration.ts
@@ -1,0 +1,62 @@
+/* eslint-disable no-console */
+
+import { usernameCollection } from '../../firebase-admin-config';
+import { getFirestoreCourseOptInOptOutChoicesBuilder } from '../requirement-graph-builder-from-user-data';
+import { getUserDataOnAdmin } from './requirement-graph-admin-utils';
+
+/** Compute opt-out choices for given user using their existing choices. */
+async function runOnUser(userEmail: string) {
+  const {
+    courses,
+    onboardingData,
+    toggleableRequirementChoices,
+    selectableRequirementChoices,
+  } = await getUserDataOnAdmin(userEmail);
+
+  const builder = getFirestoreCourseOptInOptOutChoicesBuilder(
+    courses,
+    onboardingData,
+    toggleableRequirementChoices,
+    selectableRequirementChoices,
+    {}
+  );
+
+  console.log(userEmail);
+  courses.forEach(course => {
+    const choices = builder(course);
+
+    console.log(`${course.code} (${course.uniqueId}):`);
+    console.log(
+      `- selected requirement: ${selectableRequirementChoices[course.uniqueId] || 'None'}`
+    );
+    console.log(`- optOut = [${choices.optOut.join(', ')}]`);
+    console.log(
+      `- acknowledgedCheckerWarningOptIn = [${choices.acknowledgedCheckerWarningOptIn.join(', ')}]`
+    );
+  });
+
+  console.log('\n');
+}
+
+async function main() {
+  let userEmail = process.argv[2];
+  if (userEmail != null) {
+    await runOnUser(userEmail);
+    return;
+  }
+  const collection = await usernameCollection.get();
+  const userEmails = collection.docs.map(it => it.id);
+  for (userEmail of userEmails) {
+    console.group(`Running on ${userEmail}...`);
+    try {
+      // Intentionally await in a loop to have no interleaved console logs.
+      // eslint-disable-next-line no-await-in-loop
+      await runOnUser(userEmail);
+    } catch (e) {
+      console.log(e);
+    }
+    console.groupEnd();
+  }
+}
+
+main();
--- a/src/requirements/admin/opt-out-data-migration.ts
+++ b/src/requirements/admin/opt-out-data-migration.ts
@@ -39,14 +39,14 @@ async function runOnUser(userEmail: string) {
 }
 
 async function main() {
-  let userEmail = process.argv[2];
-  if (userEmail != null) {
-    await runOnUser(userEmail);
+  const userEmailFromArgument = process.argv[2];
+  if (userEmailFromArgument != null) {
+    await runOnUser(userEmailFromArgument);
     return;
   }
   const collection = await usernameCollection.get();
   const userEmails = collection.docs.map(it => it.id);
-  for (userEmail of userEmails) {
+  for (const userEmail of userEmails) {
     console.group(`Running on ${userEmail}...`);
     try {
       // Intentionally await in a loop to have no interleaved console logs.
--- a/src/requirements/requirement-graph-builder-from-user-data.ts
+++ b/src/requirements/requirement-graph-builder-from-user-data.ts
@@ -180,10 +180,10 @@ export function getFirestoreCourseOptInOptOutChoicesBuilder(
    * | (has warning)    |                                        |                                   |
    */
   return function builder(course) {
-    const connectedRequirementsWithDoubleCountingAccounted = graphWithDoubleCountingAccounted.getConnectedRequirementsFromCourse(
+    const requirementsWithDoubleCountingRemoved = graphWithDoubleCountingAccounted.getConnectedRequirementsFromCourse(
       course
     );
-    const connectedRequirementsWithoutDoubleCountingAccounted = graphWithoutDoubleCountingAccounted.getConnectedRequirementsFromCourse(
+    const allRelevantRequirements = graphWithoutDoubleCountingAccounted.getConnectedRequirementsFromCourse(
       course
     );
     /**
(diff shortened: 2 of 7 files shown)


hahnbeelee:
Can you explain this case to me? Why is the elective in optOut?
```
CS 4780 (67):
- selected requirement: Major-CS-Technical Electives
- optOut = [Major-CS-CS Electives]
- acknowledgedCheckerWarningOptIn = [Major-CS-Technical Electives]
```

hahnbeelee:
And when you say that optOut is the same as it is in the system are you referring to Ben's unused optOut code?

hahnbeelee:
Not sure if you need to fix this here but this type is deprecated.
image

hahnbeelee:
How do the decisions made here affect our choice of choosing optIn or out?

SamChou19815:
@hahnbeelee 

> And when you say that optOut is the same as it is in the system are you referring to Ben's unused optOut code?

No, I'm referring to the discussion when we first established that all user choices are modeled as opt-in/opt-out.

> Not sure if you need to fix this here but this type is deprecated. (in requirement-graph-builder-from-user-data)

I intentionally mark is as deprecated. The replacement type is `FirestoreCourseOptInOptOutChoices`. Full removal of that deprecated type is pending on the migration.

src/requirements/requirement-graph-builder-from-user-data.ts
```diff
@@ -99,7 +106,114 @@ export default function buildRequirementFulfillmentGraphFromUserData(
...
+   * |                  |                                        |                                   |
+   * | allow double     | True                                   | implicit (connected by default)   |
+   * | counting         |                                        |                                   |
+   * |                  |                                        |                                   |
+   * | not connected    | False                                  | `optOut`                          |
+   * | (no warning)     |                                        |                                   |
+   * |                  |                                        |                                   |
+   * | not connected    | False                                  | implicit (unconnected by default) |
+   * | (has warning)    |                                        |                                   |
+   */
```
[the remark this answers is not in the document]
hahnbeelee: Can you explain where the current requirement types go in this graph? Specifically, the non-double countable courses and requirements with checker warnings?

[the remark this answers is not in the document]
hahnbeelee: You don't have to push this but I'm trying to understand how this graph would look:
```
  /**
   * The table below summerizes the type of all possible requirement-course edges before
   * double-counting elimination, and where they will go in the new format.
   *
   * -------------------------------------------------------------------------------------------------
   * | Requirement Type | edge exists after double-counting elim | where is it in the new format     | Related Requirements |
   * | ---------------- | -------------------------------------- | --------------------------------- | -------------------- |
   * | selected         | True                                   | implicit (connected by default)   |                      |
   * | (no warning)     |                                        |                                   |                      |
   * |                  |                                        |                                   |                      |
   * | selected         | True                                   | `acknowledgedCheckerWarningOptIn` |                      |
   * | (has warning)    |                                        |                                   |                      |
   * |                  |                                        |                                   |                      |
   * | allow double     | True                                   | implicit (connected by default)   |                      |
   * | counting         |                                        |                                   |                      |
   * |                  |                                        |                                   |                      |
   * | not connected    | False                                  | `optOut`                          |  non-double count    |
   * | (no warning)     |                                        |                                   |                      |
   * |                  |                                        |                                   |                      |
   * | not connected    | False                                  | implicit (unconnected by default) |                      |
   * | (has warning)    |                                        |                                   |                      |
   */
```

hahnbeelee:
I feel we can rename `arbitraryOptIn` to `optIn` because the existence of `acknowledgedCheckerWarningOptIn` kinda implies that the other optIn is arbitrary.

SamChou19815:
> I feel we can rename `arbitraryOptIn` to `optIn` because the existence of `acknowledgedCheckerWarningOptIn` kinda implies that the other optIn is arbitrary.

This is still pending whether we go with default opt-in or opt-out. If we default by opt-out, there will be another kind of opt-in lol.

src/requirements/admin/opt-out-data-migration.ts
```diff
@@ -0,0 +1,62 @@
...
+      `- acknowledgedCheckerWarningOptIn = [${choices.acknowledgedCheckerWarningOptIn.join(', ')}]`
+    );
+  });
+
+  console.log('\n');
+}
+
+async function main() {
+  let userEmail = process.argv[2];
+  if (userEmail != null) {
```
zachary0kent, line 43: Is there reason you use `!=` instead of `!==` here?

src/requirements/admin/opt-out-data-migration.ts
```diff
@@ -0,0 +1,62 @@
...
+    await runOnUser(userEmail);
+    return;
+  }
+  const collection = await usernameCollection.get();
+  const userEmails = collection.docs.map(it => it.id);
+  for (userEmail of userEmails) {
```
zachary0kent, line 49: Maybe it would be better to change the name of the loop variable here and declare it `const`?

src/requirements/requirement-graph-builder-from-user-data.ts
```diff
@@ -99,7 +106,113 @@ export default function buildRequirementFulfillmentGraphFromUserData(
...
+  return function builder(course) {
+    const connectedRequirementsWithDoubleCountingAccounted = graphWithDoubleCountingAccounted.getConnectedRequirementsFromCourse(
+      course
+    );
+    const connectedRequirementsWithoutDoubleCountingAccounted = graphWithoutDoubleCountingAccounted.getConnectedRequirementsFromCourse(
+      course
+    );
```
zachary0kent, line 188: I think it might be more readable if some of these variable and function names were shorter

src/requirements/requirement-graph-builder-from-user-data.ts
```diff
@@ -99,7 +106,113 @@ export default function buildRequirementFulfillmentGraphFromUserData(
...
+     * ```
+     * union(
+     *  selectableRequirementChoices[course],
+     *  requirements that allow double counting and the given course can be used to satisfy,
+     *  complementary,
+     * ) == all requirements that course can be used to satisfy
+     * ```
+     */
+    const complementary = new Set(connectedRequirementsWithoutDoubleCountingAccounted);
+    connectedRequirementsWithDoubleCountingAccounted.forEach(r => complementary.delete(r));
```
zachary0kent, line 202: Could you change the callback for `forEach` to be `complementary.delete` instead of `r => complementary.delete(r)`?

zachary0kent (commented):
Honestly, I feel like I would need to look extensively through all of your previous PR's/talk with you more in person to really understand what's going on here. However, I was always very interested in the work you did on the requirement graph and wanted to take a look at what you've been working on! For now I just left some nitty suggestions/questions.

hahnbeelee:
How do the updates from #522 affect this?

SamChou19815:
> How do the updates from #522 affect this PR?

No change for now. The two graph approach will only be implemented after we have the design. Currently it's ok with one graph because we know existing data does not have any constraint violations.

hahnbeelee (approved):
Alright I believe the logic checks out. After you address Zak's comments let's merge.

src/requirements/admin/opt-out-data-migration.ts
SamChou19815: `!= null` will check for both `null` and `undefined`, so it's usually used and why it's allowed by linters.

src/requirements/requirement-graph-builder-from-user-data.ts
SamChou19815: That will actually crash the program due to JS `this` receiver unbinding. A simple example:

Screen Shot 2021-11-03 at 11 52 19

SamChou19815:
Planning to merge once the release PR is merged.

PR #558, opened 2021-10-25
Onboarding modal refactor

zachary0kent:
### Summary 

This pull request refactors the transfer credit onboarding modal by factoring out the frontend code for AP and IB score input into one component. I'm going to add another input for CASE exams, so I thought it would be better to get this refactor out of the way now (instead of having to duplicate the same code used for AP/IB). No changes are visible to the client, and the modal looks the same as it did before.

Screen Shot 2021-10-25 at 10 26 16 AM
Screen Shot 2021-10-25 at 10 27 29 AM


willespencer (approved):
LGTM!

PR #562, opened 2021-10-30
Add MEng CS program

benjamin-shen:
### Summary 

This pull request adds the Master of Engineering grad program in Computer Science.

- [x] added meng cs data
- [x] updated requirement id snapshot test to include grad programs
- [x] simplified A&S foreign language requirement checker

TODOs
- [ ] convert [python script ](https://gist.github.com/benjamin-shen/557bfa3e61443dfd2caf75414666a1d3)into firebase function (?) -- not sure if this is worth it
- [ ] add back meng cs after undergrad-grad constraints are implemented

### Test Plan 

**no visible frontend changes**

~~ no need to test! the following can be tested in a later pr

Requirements bar:
image

1. Switch to the MEng CS grad program in the onboarding modal.
2. Verify that the requirements summarize [the program requirements](https://www.cs.cornell.edu/masters/academics/degreerequirements) and that the text description / "Learn More" links are reasonable.
  a. Add an advanced (5000+) non-seminar/project CS class (it should apply to the 30 total credits and 15 cs credits)
  b. Add an [advanced elective class](https://www.cs.cornell.edu/masters/academics/preapprovedcourses) (it should apply to the 30 total credits)
  c. Add CS 5999 (it should apply to the 30 total credits and project requirement)
3. Switch back to another program when you're done testing so staging isn't mad at you.

~~ do not test ^

### Notes 

I would appreciate if @SamChou19815 and anyone else interested in the MEng CS program would take a look.


src/requirements/data/index.ts
```diff
@@ -377,6 +378,11 @@ const json: RequirementsJson = {
     },
   },
   grad: {
+    MENGCS: {
+      name: 'Master of Engineering (MEng) in Computer Science',
```
benjamin-shen, line 382: Any preference between `Master of Engineering (MEng) in Computer Science` vs `Master of Engineering (MEng) Program in Computer Science`? Or another option?

src/requirements/requirement-frontend-utils.ts
```diff
@@ -78,8 +78,6 @@ export function requirementAllowDoubleCounting(
 ): boolean {
   // All minor requirements are automatically double-countable.
   if (requirement.sourceType === 'Minor') return true;
-  // All grad program requirements are automatically not double-countable.
-  if (requirement.sourceType === 'Grad') return false;
```
benjamin-shen, line 82: This is already implied for all requirements. Having an early `return false` means grad requirements cannot allow double counting, even with the `allowCourseDoubleCounting` flag.

src/requirements/data/index.ts
SamChou19815: What about just CS M.Eng to save some space

SamChou19815 (commented):
It's important to note that the current system does not have the ability to prevent you from using class already used for undergrad to also fulfill the 30 credit requirement in MEng. (e.g. I use CS 5414 to count toward some undergrad requirement.) 

This is a fundamental limitation of the current system that is easily solvable by the constraint violation system (by introducing another constraint that forbidden any double counting between undergrad and grad).

The data here looks good, but the above-mentioned problem makes me wonder whether we should ship this before we have better system in place.

SamChou19815:
Might also be relevant to opt-in vs opt-out discussion since I'm proposing a new type of constraint violation above. I feel it's more easily understandable for both users and us with the opt-out model.

src/requirements/requirement-frontend-utils.ts
SamChou19815: See the general comment

SamChou19815: See the general review.

SamChou19815 (approved):

PR #564, opened 2021-11-02
Add a type and sample CS-Eng file to support Templates

willespencer:
### Summary 

This pull request is the first step towards adding Templates to CoursePlan

- [x] Add type `PlaceholdersForRequirement` to represent a set of placeholders for a requirement in a major, college, or university.
- [x] Create `cs-eng.ts` with a list of `PlaceholdersForRequirement` representing all placeholders in the CS Engineering template.

More information about the data format and template semesters can be found on [this Notion page](https://www.notion.so/courseplan/Templates-ddc4980bde08420591530945edf6ce48)

Remaining TODOs:

- [x] Remove the computing placeholder if people agree it is redundant
- [ ] Add more data files and support for templates within CP

### Test Plan 

Compare the data file with the CS FA2021 template to ensure all requirements are correctly matched to where placeholders should be.

image

Furthermore, confirm the site and `npm run req-gen` are unchanged

### Notes 

The data format was based on engineering CS, so this is subject to change if other templates are more complicated, or the questions on the notion doc above are resolved.


src/requirements/data/templates/cs-eng.ts
```diff
@@ -0,0 +1,87 @@
+import { PlaceholdersForRequirement } from '@/requirements/types';
+
+const csEngTemplate: readonly PlaceholdersForRequirement[] = [
```
benjamin-shen, line 3: Maybe defining something like `type Template = readonly PlaceholdersForRequirement[]` and `const csEngTemplate: Template` will make this more clear?

benjamin-shen (approved):
The data looks good!

That's a good point about the `Computing` requirement. The two requirement slots have the same checker, which means there would be two identical placeholders in the same semester -- potentially very confusing (eg. the user might add CS111x twice rather than deleting the second placeholder). I think there should be a one-to-one relation from PlaceholdersForRequirements (i.e. slots) to placeholders.

However, I think directly removing PlaceholdersForRequirements from the data when we notice an overlap may be less systematic and more prone to error, especially for majors we're not as familiar with. Another possible approach would be to intentionally check for "duplicate" placeholders (comparing the checkers and seeing if the requirements are not in a constraint) before adding them. This might also work better if we want to support simultaneous templates in the future; for example, two templates for a double major in the same college would necessarily have overlapping placeholders for the university and college.

This might be a point to add to the "Questions to Resolve" on Notion. We can also discuss more during dev meeting.

hahnbeelee (approved):
LGTM!

Aryan77:
Looks good!

PR #565, opened 2021-11-02
Fix deployment of firebase functions to work for windows machines

willespencer:
### Summary 

Fix deployment of firebase functions to work for windows machines, broken in #539 with the error `'..' is not recognized as an internal or external command,`

- [x] `npm install esbuild` in the firebase functions folder so it can be called without filepaths (see below notes for rationale)
- [x] Add `engines: node 10` to package to resolve the below error

image

### Test Plan 

Confirm TrackUsers can be deployed to the dev env from the `functions` folder with `npm run deploy` and then successfully run and tested [here](https://console.cloud.google.com/functions/details/us-central1/TrackUsers?authuser=0&project=cornelldti-courseplan-dev&tab=testing) on both Windows and MacOS machines 

Windows:
image

Mac: Could @SamChou19815 or @hahnbeelee test this?

### Notes 

Notes on why the above changes were made instead of alternatives...

1. Trying to use the `cd` command to go into the existing directory with `esbuild` before calling it directly led to the deployment working, but the function failed to run on the cloud console - still not sure why.
2. Replacing the forward slashes with escaped back slashes worked locally, but doubt that this would work on Mac machines
3. Installing `esbuild` globally with `npm` worked as well, but this would require all other developers to do so or it would break on their machines.

### DIFF
--- a/functions/deployment-template/package.json
+++ b/functions/deployment-template/package.json
@@ -5,5 +5,8 @@
   "main": "index.js",
   "scripts": {
     "deploy": "firebase deploy --only functions"
+  },
+  "engines": {
+    "node": "10"
   }
 }
--- a/functions/package-lock.json
+++ b/functions/package-lock.json
@@ -662,6 +662,132 @@
       "integrity": "sha1-6WQhkyWiHQX0RGai9obtbOX13R0=",
       "optional": true
     },
+    "esbuild": {
+      "version": "0.13.12",
+      "resolved": "https://registry.npmjs.org/esbuild/-/esbuild-0.13.12.tgz",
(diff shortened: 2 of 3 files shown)


SamChou19815 (approved):
LGTM

hahnbeelee (approved):
Ran `npm run deploy` on the dev instance and ran trackUsers [here](https://console.cloud.google.com/functions/details/us-central1/TrackUsers?project=cornelldti-courseplan-dev&tab=testing) Works well for me!

hahnbeelee (changes requested):
waitttt when I try to run it locally this happens:
image

SamChou19815:
> waitttt when I try to run it locally this happens: image

This is also pre-existing on master, and it's caused by my bundling change. So i think it's better to let this one go through first and we can think how to deal with that later.

On that note, I'm not sure what's the benefit of keeping this as a firebase function instead of a normal script, so maybe we don't need to solve that problem after all...

hahnbeelee (approved):
Ok per Sam's last comment - let's push! :)

PR #567, opened 2021-11-02
Compound Requirements Progress Bug

willespencer:
### Summary 

Fix bug where individual compound requirements do not count toward requirements fulfillment.

In prod, a compound requirement like engineering liberal studies will count as "fulfilled" on the progress bar text if the main requirement is complete, but now, each additional requirement will also increase fulfillment counter by 1. The progress bar and the denominator has always counted each part of the compounded requirement, leading to the screenshotted issue.

image

### Test Plan 

1. Confirm that completely fulfilling engineering requirements leads to the progress bar saying "15/15"
image

2. Confirm that fulfilling only some of "Liberal Studies" leads to the right fulfillment count, and that the count and progress bar update if a part of the compounded requirement is then fulfilled.
image

### DIFF
--- a/src/components/Requirements/RequirementHeader.vue
+++ b/src/components/Requirements/RequirementHeader.vue
@@ -217,7 +217,9 @@ export default defineComponent({
     requirementFulfilled(): number {
       let fulfilled = 0;
       this.req.reqs.forEach(req => {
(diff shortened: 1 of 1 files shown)


hahnbeelee (approved):
Did the test plan! works for me

PR #572, opened 2021-11-04
Use `CYPRESS_TEST_UID` from the repo

SamChou19815:
### Summary 

Looks like we are running on different accounts for cypress tests, which causes the tests to fail on CI but not locally... Not sure how that happens. Let's all use the `CYPRESS_TEST_UID` from the repo.

### Test Plan 

A side by side comparison to prove my point above (left CI, right local):
Screen Shot 2021-11-04 at 15 51 17


willespencer (approved):
Frontend tests pass! Thanks so much Sam for looking into this and resolving.

PR #573, opened 2021-11-04
Implement new user flow for course/subject color update feature

bungaepyo:
### Summary 

This pull request implements a new user flow and design for the course/subject color update feature. Previously, a user was only able to change the color for a single course through the course menu. Now, a user is able to (1) choose among a wider variety of colors (2) choose to apply the change either to a single course or to the entire academic subject (3) confirm their change both visually and with a checkmark.

- [x] add new colors
- [x] implement new design for CourseMenu
- [x] add EditColor modal
- [x] implement course/subject color update function

### Test Plan 
- Default with 4 semesters
Screen Shot 2021-11-04 at 5 42 32 PM
- New design & colors for edit color
Screen Shot 2021-11-04 at 5 42 53 PM
- Tooltips
Screen Shot 2021-11-04 at 5 44 05 PM
- Edit color modal
Screen Shot 2021-11-04 at 5 44 29 PM 
- Confirmation message & Single course color change result
Screen Shot 2021-11-04 at 5 44 57 PM
- Academic subject (CS) change to green
Screen Shot 2021-11-04 at 5 47 25 PM
- Academic subject color change result
Screen Shot 2021-11-04 at 5 47 57 PM


src/store.ts
```diff
@@ -7,6 +7,7 @@ import RequirementFulfillmentGraph from './requirements/requirement-graph';
 import { createAppOnboardingData } from './user-data-converter';
 import {
   allocateAllSubjectColor,
+  udpateSubjectColor,
```
hahnbeelee, line 10:
```suggestion
  updateSubjectColor,
```

hahnbeelee: And everywhere else it's spelled wrong

src/assets/constants/colors.ts
```diff
@@ -5,32 +5,52 @@ export const reviewColors = {
 };
 
 export const coursesColorSet = [
```
hahnbeelee, line 7: Are both `coursesColorSet` and `SUBJECT_COLORS` necessary? Can we make them 1 thing?

src/components/BottomBar/BottomBarState.ts
```diff
@@ -27,6 +27,12 @@ export const reportCourseColorChange = (courseUniqueID: number, color: string):
   );
 };
 
+export const reportSubjectColorChange = (code: string, color: string): void => {
```
hahnbeelee, line 30: nice

hahnbeelee:
I added a new course and the selected color is not checked.
image

hahnbeelee:
is there anyway that the top radio button can be automatically chosen?
image

src/components/Modals/CourseMenu.vue
```diff
@@ -158,20 +149,26 @@ export default defineComponent({
...
+    'open-edit-color-modal': (color: string) => typeof color === 'string',
     'edit-course-credit': (credit: number) => typeof credit === 'number',
   },
   methods: {
     deleteCourse() {
       this.$emit('delete-course');
     },
-    colorCourse(color: { hex: string }) {
-      this.$emit('color-course', color.hex.substring(1));
-      GTagEvent(this.$gtag, 'course-edit-color');
```
hahnbeelee, line 170: Can we preserve this gtag?

hahnbeelee: and also create a new one for `course-subject-edit-color`?

src/components/Semester/Semester.vue
```diff
@@ -347,6 +350,19 @@ export default defineComponent({
           ),
         })
       );
+      this.openConfirmationModal(`Changed color for ${code}`);
+    },
+    colorSubject(color: string, targetCourse: string) {
+      const code = targetCourse.split(' ')[0];
```
hahnbeelee, line 356: Can you rename either this variable or the one inside the `colorCourse` function? I don't even know which one is right, but it's confusing that they're both `code` since I know we're looking at 2 different properties of a course right now.

hahnbeelee (changes requested):
Okieee Left some comments about the code - general functionality looks good to me!

bungaepyo:
> I added a new course and the selected color is not checked. The checkmark appears after the first time that the color is changed. image
@hahnbeelee 

- All subject colors are initially assigned and stored during `initializeFirestoreListeners` within `store.ts`.
- New colors are entirely different from previous ones (e.g. current orange and previous orange has different hex values).
- Users signing up after this feature released will have the new colors initially assigned as subject colors, and thus will have a color checkmarked from the beginning.
- Users who signed up before this feature released (all existing users) will have the old colors assigned as subject colors, and will not have any color checkmarked unless they explicitly change color for a course/subject using the edit color modal.

I believe the reason why you aren't seeing any checkmarks after adding a new course is because CS2110 has the **OLD ORANGE** as the subject color. Since the hex values & variety is very different for current and previous colors, I thought it was natural to allow users to first use previous colors as long as they want, and give them options to edit to one of the new colors.

hahnbeelee (approved):
LGTM! I understand your reasoning for preserving old color types.

src/assets/constants/colors.ts
```diff
@@ -5,32 +5,52 @@ export const reviewColors = {
...
-  {
-    text: 'Gray',
```
willespencer, line 9: Do we want users to still have gray courses in their plan, even though we do not offer the color anymore?

src/components/Modals/CourseMenu.vue
```diff
@@ -35,15 +28,23 @@
...
           >
             <div class="courseMenu-left">
-              <div
-                class="courseMenu-icon courseMenu-icon--color"
-                :style="{ backgroundColor: color.hex }"
-              ></div>
-              <span class="courseMenu-text">{{ color.text }}</span>
+              <div class="courseMenu-color--icon" :style="{ backgroundColor: color.hex }">
+                <img
+                  v-if="`#${courseColor}` === color.hex"
```
willespencer, line 39: Instead of checking for hex value here, could we check for color name? This would resolve the issue with checkmarks not showing for existing users, and since the hex value changes are not really noticeable, I think it is ok to "lie" to the user and show that they are using the same color.

willespencer (changes requested):
Looks really good, and very excited to finally see updated colors! In general, it seems to work in all cases, including changing bottom bar and requirements bar course colors when changing academic subject colors too :D 

I have a couple remaining functionality questions, but good to approve after you respond!
- I also noticed the discrepancy with arrows not showing up that Hahnbee did (before seeing her review). Also understand your reasoning, but I might argue that we should still show the checkmark anyway? See my comment in the code
- I'm noticing the color tooltip names are covered by the cursor when hovering, at least on my windows machine, which makes it hard to read the names (especially the 2nd rows Cornell themed ones), or hover over other colors (if hovering over one in the top row, the one directly below gets covered by the tooltip). Could they be adjusted at all?

src/assets/constants/colors.ts
einc: Maybe we can comment out gray for now, because I think we might want to reserve it for placeholder for templates?

einc:
Screen Shot 2021-11-08 at 8 26 54 PM

changing colors for individual courses/subject works awesome!! this is so cool toby!! 🥳
i am noticing that the colors get cut off on the right side, can we have the color picker face the other way in those cases?
Screen Shot 2021-11-08 at 8 29 50 PM

hahnbeelee:
^ @einc that's weird for me it shows up on the left
image

kkkehui:
Toby!!! this is so cool!!! thanks for implementing colors! 
For the selected color, I think the selected color block is one or two pixels below compared to other color blocks. Can you make them all on the same level/height? I think the checkmark is currently centered, which is perfect! 
Screen Shot 2021-11-09 at 6 11 10 PM

src/assets/constants/colors.ts
bungaepyo: Even if we do not offer gray color anymore, subject colors for each user are saved in firestore so existing users would be able to use gray color unless they make a change by themselves.

src/components/Modals/CourseMenu.vue
bungaepyo: I looked into this and I don't know if we can match with color name at all, since we are only storing the hex value in each course object. For example:

`code: "CS 1110"
color: "FF3B30"
creditRange: (2) [4, 4]
credits: 4
crseId: 358526
lastRoster: "FA21"
name: "Introduction to Computing Using Python"
semesters: (3) ['Fall', 'Spring', 'Summer']
uniqueID: 200`

bungaepyo: > I'm noticing the color tooltip names are covered by the cursor when hovering, at least on my windows machine, which makes it hard to read the names (especially the 2nd rows Cornell themed ones), or hover over other colors (if hovering over one in the top row, the one directly below gets covered by the tooltip). Could they be adjusted at all?
I adjusted this issue.

bungaepyo:
> Toby!!! this is so cool!!! thanks for implementing colors! For the selected color, I think the selected color block is one or two pixels below compared to other color blocks. Can you make them all on the same level/height? I think the checkmark is currently centered, which is perfect! Screen Shot 2021-11-09 at 6 11 10 PM

Thanks for the feedback!!! I adjusted this height issue.

src/components/Modals/CourseMenu.vue
willespencer: > I looked into this and I don't know if we can match with color name at all, since we are only storing the hex value in each course object. 

Shoot yeah I think you're right. If we really want to, I think instead we could run a script on the Firestore (at any time) that would change a user's hex codes from the old values to the new values in the semesters collection. Not sure if that is worth it but it would probably resolve the arrow issue. This is definitely good for now though!

willespencer (approved):
Agree with your point about colors, and tooltips are looking better! Approved 😄

bungaepyo:
> Screen Shot 2021-11-08 at 8 26 54 PM
> 
> changing colors for individual courses/subject works awesome!! this is so cool toby!! 🥳 i am noticing that the colors get cut off on the right side, can we have the color picker face the other way in those cases? Screen Shot 2021-11-08 at 8 29 50 PM

I adjusted this issue!

PR #574, opened 2021-11-05
Record user choice in both old and new data format for compatibility

SamChou19815:
### Summary 

This PR adds a compatibility layer for migration of user choice format. In short, this PR:

- does not change how requirement is computed
- does not change how user choice _that affects requirement computation_ is stored
- instead, it only modifies the new format in `user-overridden-requirement-choices` in parallel.

With this setup, this diff can go into prod at any time. New user data can be recorded once we push, which gives us a lot of flexibility of when we want to run the migration script: run it after release to prod, then there will be no need to run the migration script in the future (since with this PR prod data will self-correct), we can replace the entire old choice infra with new choice infra.

### Test Plan 

After added MATH 4710 (won't change choice since it only connects to reqs that allow double counting), new `user-overridden-requirement-choices` appear in the DB:

Screen Shot 2021-11-05 at 17 21 37

### DIFF
--- a/src/components/Modals/NewCourse/NewCourseModal.vue
+++ b/src/components/Modals/NewCourse/NewCourseModal.vue
@@ -98,7 +98,7 @@ export default defineComponent({
         selectedCourse,
         store.state.groupedRequirementFulfillmentReport,
         store.state.toggleableRequirementChoices,
-        store.state.overriddenFulfillmentChoices
+        /* deprecated AppOverriddenFulfillmentChoices */ {}
       );
 
       const requirementsThatAllowDoubleCounting: string[] = [];
--- a/src/global-firestore-data/selectable-requirement-choices.ts
+++ b/src/global-firestore-data/selectable-requirement-choices.ts
@@ -1,23 +1,53 @@
-import { selectableRequirementChoicesCollection } from '../firebase-frontend-config';
+import { getCourseCodesArray } from '../requirements/requirement-frontend-computation';
+import { getFirestoreCourseOptInOptOutChoicesBuilder } from '../requirements/requirement-graph-builder-from-user-data';
+import {
+  db,
+  selectableRequirementChoicesCollection,
+  overriddenFulfillmentChoicesCollection,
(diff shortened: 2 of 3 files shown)


src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -98,7 +98,7 @@ export default defineComponent({
         selectedCourse,
         store.state.groupedRequirementFulfillmentReport,
         store.state.toggleableRequirementChoices,
-        store.state.overriddenFulfillmentChoices
+        /* deprecated AppOverriddenFulfillmentChoices */ {}
```
SamChou19815, line 101: `overriddenFulfillmentChoices` is all empty. Since it collides with the new name I choose, I decide just to replace it with `{}` to avoid confusion. The old `overriddenFulfillmentChoices` are not completely removed yet, to do so involves quite a lot of changes that will make this PR big and unfocused.

hahnbeelee (changes requested):
I deleted my document and it wont appear again whereas if i delete a document from user-onboarding-data or user-semesters it generates again

hahnbeelee:
Oh wait jk does it only generate when specific courses are added?

hahnbeelee:
- i know we cant test `arbitraryOptIn` rn
- I was able to get a successful `acknowledgedCheckerWarningOptIn` by choosing from the add modal
- How do I replicate an `optOut`?

SamChou19815:
> Oh wait jk does it only generate when specific courses are added?

@hahnbeelee 
Yes, this is triggered only when you are making a choice about requirement. i.e. when you are changing `user-selectable-requirement-choices` through the courseplan UI.

SamChou19815:
> How do I replicate an `optOut`?

You can try to find a course that can satisfy two requirements (without checker warnings and do not allow double counting) at once.

hahnbeelee:
I was able to replicate optOut by choosing "None" here :)
image

hahnbeelee (approved):
looks good to meee

PR #554, opened 2021-10-22
Finish script to migrate engineering liberal studies data

SamChou19815:
### Summary 

This PR finishes the script initially added in #542. The only real effective change is to add a more that actually modifies the database. I also tweaked some other logs produce by this script since I determined that they are all safe.

### Test Plan 

`npm run ts-node -- src/requirements/admin/migrate-engineering-liberal-arts-data.ts --run-on-db` if you want to run the script.

### Notes

I have local changes that removes the `allowDoubleCounting` flag. However, it can't be included in this PR, since removing the flag will break the assumption of the script that engineering liberal studies is double-countable.

We would need a coordinated release done in short time after this is merged. The sequence would be:

1. Release everything on master to prod before this PR is approved and merged.
2. Approve and merge this PR.
3. Run this on prod
4. PR for removing `allowDoubleCounting` is created, approved and merge to master.
5. (Optional) Another PR to delete this script
5. Create another release that contains the PR in step 4 only.

PR #575, opened 2021-11-06
[easy] No longer mark engineering liberal studies data as allow double counting

SamChou19815:
### Summary 

To be merged after #554 is run on prod data.

### Test Plan 

👀


hahnbeelee:
technically can't this be merged to master and then later when we push to prod we should run the script on prod?

SamChou19815:
> technically can't this be merged to master and then later when we push to prod we should run the script on prod?

yes that's right

hahnbeelee (approved):
LGTM. i ran the script on prod! I'll have to remember to run it again when we push to prod though.

PR #569, opened 2021-11-03
fix 13-inch screen issue

bungaepyo:
### Summary 

This pull request fixes the issue of 13-inch screens having only 2 columns in the default compact view.

### Test Plan 

@hahnbeelee

PR #577, opened 2021-11-07
Desktop toggle requirements btn + unlock semester grid

Enochen:
### Summary 

This pull request adds a button to the navbar on desktop that allows for toggling the display of the requirements pane as to provide more potential visual space for the semester view.  I just really wanted to see more semesters on my screen lol

The diff also converts SemesterView to CSS grid (from flexbox) for listing the semester cards. The old implementation using flexbox had 2 columns hardcoded in (4 for compact mode). Using grid, we can have variable number of columns. allowing us to take advantage of the extra space with the requirements pane hidden.

Finally, this diff converts the desktop navbar style `position` from `fixed` to `sticky`. A `fixed` position means that no space is created for the element in the document. The position of requirements bar was implemented with a hacky `margin-left: 4.5rem`, which is not needed with the `sticky` positioning, which does create space for the element. This also allows the sidebar to not overlap the semester view when the requirements bar is hidden.
The mobile styling doesn't require changes since it sits on the top of the screen, not the side, where overlapping is acceptable behavior.

- [x] add button in navbar to toggle requirements bar
- [x] show requirements on page load for desktop, but hide by default on page load for mobile/tablet
- [x] convert semesterview card layout from flexbox to css grid
- [x] use `position: sticky` for sidebar (desktop navbar only)

### Remaining TODOs:

Currently the icon for the button is just an existing hamburger svg, but we can definitely come up with something better that communicates its purpose more effective.

The button is currently located in the lower end of the nav bar right above the logout button. This location may not be ideal, for example allowing for accidental clicks on the logout button. We may want to relocate this button if that is the case.

~~Also there is a bug where the button stays blue after being clicked, only returning to gray once you click somewhere else. This is demonstrated in the Toggling GIF. This issue should/will be addressed along with an actual icon.~~

- [ ] finalize location of button in navbar
- [ ] finalize button icon
- [x] fix button icon focus bug

### Test Plan 

Toggling (updated design, images below don't reflect this toggle btn change)
image

Default view on ultrawide (21:9) screen
image
image

CSS Grid implementation shows 3 columns in compact mode at 1400px - 13" laptop (referenced in #569)
image

### Notes 

If the requirements bar is set to be shown on desktop and the device switches to mobile while on the same page, the requirements page will be shown rather than the semester vue. This is because both desktop/mobile share the same state variable (idk what it's called in Vue) in Dashboard.vue for showing/hiding the requirements bar/page (namely `isOpeningRequirements`).

If we don't want this behavior we'll just have to separate them into different state variables, though this situation is extremely rare imo

Also this is behind a feature flag, call `GK.enableToggleRequirementsBarBtn()` in the console to enable.


functions/package-lock.json
```diff
@@ -6,6 +6,7 @@
     "": {
```
hahnbeelee, line 6: If there isn't a reason for this `package-lock.json` to be pushed I think it should be deleted

hahnbeelee (commented):
Scroll isn't working :,(
https://user-images.githubusercontent.com/55263191/141197871-806f1db5-b1a7-408c-8510-d0e048e8c1ce.mov

src/components/NavBar.vue
```diff
@@ -155,6 +156,16 @@ $mobile-navbar-height: 4.5rem;
...
 
+  .toggle-requirements-icon {
+    background-image: url('@/assets/images/navbar/hamburger-gray.svg');
+
+    &:hover,
+    &:focus,
+    &:active {
+      background-image: url('@/assets/images/navbar/hamburger-blue.svg');
+    }
+  }
```
hahnbeelee, line 167: delete? i dont think this code is being used rn

hahnbeelee:
Bottom bar is not looking so hot :,(
image

src/components/Requirements/RequirementSideBar.vue
```diff
@@ -1,113 +1,131 @@
...
-      data-step="1"
-      data-tooltipClass="tooltipCenter tourStep1"
+  <div>
+    <button
+      v-if="!isExpanded && showToggleRequirementsBtn"
+      class="requirement-sidebar-btn-open"
+      @click="setExpanded(true)"
     >
-      <!-- loop through reqs array of req objects -->
+      →
```
hahnbeelee, line 8: omg lol the arrow in text is so cute

src/components/Requirements/RequirementSideBar.vue
```diff
@@ -165,10 +184,10 @@ export default defineComponent({
   },
   props: {
     startTour: { type: Boolean, required: true },
+    isOpeningRequirements: { type: Boolean, required: true },
+    isDesktop: { type: Boolean, required: true },
   },
-  emits: {
-    showTourEndWindow: () => true,
-  },
+  emits: ['showTourEndWindow'],
```
hahnbeelee, line 190: I think there's a reason why it was the way it was formatted in the previous way @SamChou19815 Can you provide insight?

src/containers/Dashboard.vue
```diff
@@ -120,14 +122,17 @@ export default defineComponent({
     TourWindow,
   },
   data() {
+    const isTablet = window.innerWidth <= mediumBreakpointPixels;
+    const isMobile = window.innerWidth <= smallBreakpointPixels;
```
hahnbeelee, line 126: I think it's better to put these as `computed` values.

hahnbeelee (changes requested):
omggg enoch! this is so awesome!!! im obsessed. you have such great frontend knowledge. respect. sorry for spamming u with requests lmk if ur too busy for them lmfao.

src/components/Semester/SemesterView.vue
```diff
@@ -56,23 +55,6 @@
...
-          v-if="compact"
-          class="semesterView-empty semesterView-empty--compact"
-          aria-hidden="true"
-        ></div>
-        <div
-          v-if="compact"
-          class="semesterView-empty semesterView-empty--compact"
-          aria-hidden="true"
-        ></div>
-        <div v-if="compact"><div v-if="compact"></div></div>
```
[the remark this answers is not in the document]
Enochen: I think these were added to make the flexbox implementation work correctly

src/components/Requirements/RequirementSideBar.vue
Enochen: We probably want actual icons to replace this later on lol

src/components/Requirements/RequirementSideBar.vue
Enochen: There was a second function after `showTourEndWindow` previously, which had a more complex function signature than `() => true`. That existing forced it to be in the more verbose format, but that function has been since removed. Not sure what makes the most sense stylistically but there are a bunch of other places in the code that do `emits: [...]`.

hahnbeelee (changes requested):
Left nav bar is scrolling too
https://user-images.githubusercontent.com/55263191/141856269-094d826a-4671-4b4e-b95d-5c2f052eca77.mov

Enochen:
Left nav bar + bottom bar + scroll lockup should all be fixed now

hahnbeelee (approved):
it's so good 😭 tysm enoch! LGTM. let's ship it :3

PR #579, opened 2021-11-08
Case Exams frontend

zachary0kent:
### Summary 

This pull request implements (most) of the frontend features for CASE exams, including being able to add CASE exams onto a dropdown menu and remove then as you currently are for AP/IB exams. I decided to hold off making the CASE exams appear in the "Onboarding Review" section, as that would require several in depth changes to the AP/IB types that I think should come after the AP/IB infrastructure change. This functionality is gated behind the "Case" feature flag, so you won't initially see it until enabling it in the developer's console.

Remaining TODOs:

- [ ] Show added CASE exams on Onboarding Review
- [ ] Implement backend support

### Test Plan 

Screen Shot 2021-11-07 at 11 32 26 PM

### Blockers 
Possibly the AP/IB infrastructure change


src/user-data.d.ts
```diff
@@ -45,6 +45,18 @@ type FirestoreAPIBExam = {
   readonly optIn?: FirestoreAPIBOverriddenFulfillments;
   readonly optOut?: FirestoreAPIBOverriddenFulfillments;
 };
+
+/** Represents the name of an exam a student can take for transfer credit */
+type TransferExam = 'AP' | 'IB' | 'CASE';
+
+type FirestoreTransferExam = {
+  readonly name: TransferExam;
+  readonly score: number;
```
benjamin-shen, line 54: Should `score` be optional since we're using this type for CASE exams?

src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -59,16 +70,30 @@ type TransferClassWithOptionalCourse = {
   course?: CornellCourseRosterCourse;
 };
 
+type TransferExamInfo = {
+  exams: FirestoreTransferExam[];
+  scores?: readonly number[];
+};
```
benjamin-shen, line 76: Out of curiosity, what is the benefit of putting `scores` here and using a new type `TransferExamInfo`?

benjamin-shen (commented):
Thanks for merging in my AP/IB refactor branch. I agree with implementing CASE up to this point and good call on the feature flag 💯 

Left a few comments and questions!

src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -59,16 +70,30 @@ type TransferClassWithOptionalCourse = {
...
+  CASE: TransferExamInfo;
   classes: TransferClassWithOptionalCourse[];
 };
 
+const asAPIB = (exam: FirestoreTransferExam) => {
+  const { name } = exam;
+  if (name === 'CASE') {
+    throw new TypeError('Cannot fetch credit from CASE exam');
+  }
+  return { ...exam, type: name };
```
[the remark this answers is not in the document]
zachary0kent: Yeah I agree. Hopefully, this should only be temporary once we resolve the APIB and TransferExam types.

src/components/Modals/Onboarding/OnboardingTransfer.vue
zachary0kent: Just reduced a bit of code duplication between "examsAP, examsIB" and "scoresAP, scoresIB." Before, we had functions that were very similar for "examsAP" and "examsIB" etc

benjamin-shen: Despite the similarity in naming, `scoresAB` and `scoresIB` are very different from `examsAP` and `examsIB`. They are just global constants -- I don't think they need to be part of a new type. I was actually confused at first because I thought `scores` referred to user scores.

IMO, if we just have `FirestoreTransferExam[]` (like for case exams / if we keep `scoresAB` and `scoresIB` separate), then a new type isn't necessary. Does that make sense?

src/requirements/requirement-exam-utils.ts
```diff
@@ -133,6 +133,7 @@ function toSubjects(data: ExamRequirements[]) {
 export const examSubjects: ExamSubjects = {
   AP: toSubjects(examData.AP),
   IB: toSubjects(examData.IB),
+  CASE: ['Computer Science', 'Chemistry', 'Physics', 'Foreign Language'], // stub in for now
```
hahnbeelee, line 136: Can you make this comment in the format of `// TODO ...`

src/user-data.d.ts
```diff
@@ -45,6 +45,18 @@ type FirestoreAPIBExam = {
   readonly optIn?: FirestoreAPIBOverriddenFulfillments;
   readonly optOut?: FirestoreAPIBOverriddenFulfillments;
 };
+
+/** Represents the name of an exam a student can take for transfer credit */
+type TransferExam = 'AP' | 'IB' | 'CASE';
```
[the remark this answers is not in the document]
hahnbeelee: i like the name `examType`

hahnbeelee (approved):
- Checked frontend functionality and everything seems to be working just as it did on prod. (Onboarding modal needs to be improved 😭) Both web and mobile look good.
- Code looks good to me too. Although I agree with Ben that some of the things you did to reduce code duplication makes the code a bit more confusing. As someone who has never worked with this specific part of the codebase before I am a bit confused. Maybe adding comments / changing names of things make clarify some of my confusion?

Overall everything looks pretty steady to me. I think all you have left to do is address Ben's comments!

src/components/Modals/Onboarding/OnboardingTransfer.vue
zachary0kent: Yeah that makes sense. How does it look now?

benjamin-shen (approved):
LGTM!

PR #581, opened 2021-11-09
Enforce package-lock.json does not change on CI

SamChou19815:
### Summary 

Check on CI that it doesn't change after an install.

### Test Plan 

CI


benjamin-shen (approved):
lgtm! ty

PR #582, opened 2021-11-09
Bump linters and remove `@vue/cli-service`

SamChou19815:
### Summary 

A while ago, I switched the build system to Vite. `@vue/cli-service` should be removed back then, but it's not since the linter config is still depending as on it.

After digging deep, the linter config that depends on `@vue/cli-service` is `@vue/eslint-config-airbnb`, and it has the following content:

Screen Shot 2021-11-09 at 14 58 42

As you can see, it's really just some simple wrapper around `airbnb-base` config, so it makes sense to just inline it into our eslint config. The dependency on `@vue/cli-service/webpack-config` is only used to get some module resolution logic (in our case, the `@` to `src` alias). Therefore, I replaced that with our own alias config through `eslint-import-resolver-alias`.

### Test Plan 

CI. Things can build, and devserver can start as usual.


.eslintrc.js
```diff
@@ -6,13 +6,13 @@ module.exports = {
...
     'plugin:vue/recommended',
-    '@vue/airbnb',
+    'airbnb-base',
     '@vue/typescript',
     'plugin:import/typescript',
     'prettier',
-    'prettier/vue',
     'plugin:cypress/recommended',
   ],
+  reportUnusedDisableDirectives: true,
```
SamChou19815, line 15: eslint 8 can detect unused suppressions now

.eslintrc.js
```diff
@@ -67,10 +73,19 @@ module.exports = {
       { vars: 'all', args: 'after-used', ignoreRestSiblings: true },
     ],
     'no-use-before-define': ['off'],
-    'vue/no-mutating-props': ['warn'],
```
SamChou19815, line 70: no more warnings, so let's make this an error

benjamin-shen (approved):

Issue #576, opened 2021-11-06
AP/IB Infrastructure Update


benjamin-shen:
# Reference
Some of the AP exams used as examples in this issue are presented below, for quick reference.

| Subject | Credits | Placement |
| - | - | - |
| Chemistry | 4 | Placement out of CHEM 2070 or CHEM 2090. <br /><br />A student taking CHEM 1560, CHEM 2070, or 2090 will forfeit AP credit.  Students taking CHEM 2150 will retain AP credit. A student may also receive credit for CHEM 2070 or CHEM 2090 by passing an exam given during Fall orientation. A distinct exam is given before the start of the Spring semester that, if passed, will allow a student to receive credit for CHEM 2080. See chemistry.cornell.edu/course-information for further information. |
| Computer Science A | 4 | Placement out of CS 1110. Department also offers placement exam during fall orientation. |
| Statistics (excluding engineering students) | 4 | Placement out of AEM 2100, BTRY 3010, BTRY 6010, ILRST 2100, ILRST 6100, MATH 1710, PAM 2100, PAM 2101, PSYCH 3500, SOC 3010, STSCI 2100, STSCI 2150, STSCI 2200. Students in the College of Agriculture and Life Sciences will find credit and placement information at cals.cornell.edu/academics/registrar. Students in Hotel Administration will find credit and placement information in the SHA Student Handbook. |
| Mathematics BC | 8 | Placement out of Calculus I (MATH 1106 or MATH 1110) and Calculus II (MATH 1120, MATH 1220, or MATH 1910). |
| Microeconomics | 3 | Placement out of ECON 1110 and HADM 1410. |
| Spanish Literature | 3 | Department of Romance Studies determines placement. Students should take the LPS and afterwards the CASE to obtain appropriate placement. |

# Current Infra
Exams can be represented by a set of equivalent courses (aka course equivalents), which are represented as `CourseTaken`s outside of a user's semester courses. They are computed based on the user's college/major, taken exams, and those exams' scores. The equivalent courses are then added to the user's taken courses to be used in the requirements algorithm.

[...]

PR #584, opened 2021-11-12
Update exam data schema, add course ID to each fulfillment option

benjamin-shen:
### Summary 

This pull request is task 1 in the AP/IB Infra Migration Update #576.

- [x] updated exam data schema and exams taken type
- [x] deprecated old types and functions
- [x] implemented function to generate exam courses
- [x] updated tests
- [x] removed `NO_EQUIVALENT_COURSES_COURSE_ID` from requirements json

Remaining TODOs:

- [ ] update variable name `NO_EQUIVALENT_COURSES_COURSE_ID` => `NO_FULFILLMENT_COURSE_ID`
- [ ] refactor `'AP' | 'IB'` everywhere as `ExamType`

### Test Plan 

`npm run test exam-credit` (or check CI)
Also check that AP/IB exam functionality is unchanged.

### Notes

The initial plan was to give each exam a single course ID.

After the updated schema, the new plan implemented in this PR is to give each fulfillment option a course ID. (Each exam so far has 1-2 fulfillment options that are chosen based on the user's score.) This is so we don't need to encode the fulfillment option logic into the requirements algorithm past the exam course generation.

To handle more nuanced criteria (such as majors excluded, college-specific equivalent courses within each fulfillment option, etc), we can encode the conditions/criteria into the requirements json as suggested by @SamChou19815. For example, each requirement may have an additional field:
```ts
{
  ...otherRequirementJSONData,
  conditions?: {
    [course_id]: { // theoretically key can also be course code, but not for ap/ib conditions
      colleges?: ['EN'], // if the user is not in the EN college, course_id cannot fulfill the requirement
      excludedMajors?: ['CS'] // if the user is a CS major, course_id cannot fulfill the requirement
    }
  }
}
```

Open to comments and other suggestions.


src/requirements/data/exams/ExamCredit.ts
```diff
@@ -1,235 +1,204 @@
...
-    readonly credits: number;
-    readonly majorsExcluded?: string[];
-  };
+export type ExamFulfillment = {
+  readonly courseId: number;
+  readonly courseEquivalents?: Record<string, number[]>;
+  readonly minimumScore: number;
+  readonly credits: number;
+  readonly majorsExcluded?: string[];
 };
```
hahnbeelee, line 9: Can you write a comment for what `ExamFulfillment` is? It's another one of those vague terms that feel very intuitive for us, but might be very confusing to a newbie

src/requirements/data/constants.ts
```diff
@@ -1,10 +1,10 @@
...
-export const NO_EQUIVALENT_COURSES_COURSE_ID = 10;
+export const NO_EQUIVALENT_COURSES_COURSE_ID = 10; // TODO @bshen rename to NO_FULFILLMENTS_COURSE_ID
 
-/** Course ID for special course that satisfies total academic credits requirement. */
+/** @deprecated Course ID for special course that satisfies total academic credits requirement. */
 export const CREDITS_COURSE_ID = 11;
 
-/** Course ID for special course that satisfies FWS requirement. */
+/** @deprecated Course ID for special course that satisfies FWS requirement. */
 export const FWS_COURSE_ID = 12;
```
hahnbeelee, line 8: If this is deprecated what does this mean for this data? This is what I got after running `npm run req-graph-dump hl985@cornell.edu`
image

src/requirements/requirement-exam-utils.ts
```diff
@@ -123,25 +135,75 @@ export default function getCourseEquivalentsFromUserExams(
...
+      if (!prev || prev.minimumScore < curr.minimumScore) {
+        return curr;
+      }
+    }
+    return prev;
+  }, undefined);
+  return fulfillment;
+};
+
+export const examsTakenToExamCourses = (exams: ExamsTaken): CourseTaken[] => {
```
hahnbeelee, line 153: So does this mean that an exam can only have one course equivalent?

hahnbeelee (approved):
LGTM! Other than adding a comment to make things more clear, despite this being my first time really looking at this code everything made a lot of sense to me. Although I also left some clarifying questions just so I'm more aware of what's going on. :)

src/requirements/data/exams/ExamCredit.ts
```diff
@@ -1,235 +1,204 @@
...
+      { courseId: 100, minimumScore: 4, credits: 4 },
+      { courseId: 101, minimumScore: 5, credits: 8 },
+    ],
+    Chemistry: [
+      {
+        courseId: 102,
         courseEquivalents: {
           DEFAULT: [351265], // CHEM 2070
           EN: [359187], // CHEM 2090
         },
```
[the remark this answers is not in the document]
benjamin-shen: Yes, thanks for bringing this up! They should look similar because the conditions (in the requirements json) will be derived from the exam fulfillments (in the exam data).

The exam fulfillments data is currently being used for generating `courseTaken`s from `courseEquivalent`s, which was part of the old infra. For the new infra, it will be used to determine the conditions.

There are a few subtle differences, though:
- the exam fulfillments are properties of the exams, while the conditions are properties of the requirements
- `colleges` is a list of colleges, while the exam data uses the college as the key
- DEFAULT becomes a list of colleges

Here's a more concrete example if it helps. With the current data, AP CHEM has one equivalent course: CHEM 2090 for EN, and CHEM 2070 for DEFAULT (non-EN). This will cause the script to draw edges from an AP CHEM exam course with course id 102 to the requirements that CHEM 2070 *and* CHEM 2090 fulfill. This is the same thing as adding the course ids to the requirements json `courses` lists.
Say CHEM 2070 fulfills req1 and req2, and CHEM 2090 fulfills req1 and req3. Then, the exam course fulfills req1, req2, and req3.
- For req1, `conditions` will be undefined (or at least wouldn't have a key of 102) because CHEM 2070 and CHEM 2090 (all colleges) fulfill req1.
- For req2, `conditions` for course id 102 will be something like `colleges: [all colleges except 'EN']` because CHEM 2070 (non-EN colleges) fulfills req2.
- For req3, `conditions` for course id 102 will be something like `colleges: ['EN']` because CHEM 2090 (EN college only) fulfills req3.

src/requirements/data/constants.ts
benjamin-shen: Whoops, this was a mistake! `FWS_COURSE_ID` shouldn't be deprecated.

src/requirements/requirement-exam-utils.ts
benjamin-shen: One exam can have multiple course equivalents, but it will only be converted to a single exam course.

I think you might be conflating course equivalents and `courseTaken`s -- this does mean that in the new infra, an exam is represented by only one `courseTaken`.

hahnbeelee: Gotcha, so this is only infra for getting up to the `CourseTaken` part, and later when you refactor the graph it'll do the calculations for course equivalents?

hahnbeelee: And just to clarify - `courseTaken` is just the "middle ground"/connecting point between AP/IB and courses so that they're all represented by the same thing?

benjamin-shen: Yeah, later it'll calculate the exam course's edges based on course equivalents.

`courseTaken` is the "courses" part. Ultimately, we want `courseTaken`s to be used in the requirements computation. Semester courses are converted to `courseTaken`s for the computation.

The middle ground would be course equivalents. In the old infra, we generated one `courseTaken` per course equivalent (that basically represented the same course as the course equivalent). In the new infra, we'll generate one `courseTaken` per exam (but still use the course equivalents to determine the edges / connected requirements).

src/requirements/data/exams/ExamCredit.ts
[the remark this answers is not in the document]
hahnbeelee: O i see so the checkers for defining which courses an exam fulfills are more complex than just defining the school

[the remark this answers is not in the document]
benjamin-shen: Do you mean the checkers for defining which requirements an exam fulfills are more complex than just defining the course ids? If so, yes. We need the user data (college and major), which we can get at runtime but not when the requirements json is being generated.

hahnbeelee (approved):
Am I allowed to approve twice? LOL LGTM!!! ship it

PR #585, opened 2021-11-14
November Release

SamChou19815:
### Summary 

- Potential Breaking Changes
  - New colors
  - AP/IB infra cleanup & refactor
  - Desktop toggle
- Migration
  - Migration script for liberal studies
  - Migration script for new requirement choice data format

### Test Plan 

👀👀👀


willespencer (dismissed):
LGTM - specifically was sure to test AP/IB (including the no major bug) and the new colors implementation!

My one concern is that the semester view is now 3 columns on the non-compact view for larger Windows machines like mine, which makes it a lot harder to visualize the plan on a year-by-year basis. Would it be possible to strike a balance between having 2 columns on this view for most users, while still supporting more columns on ultra-wide screens or when the reqs bar is hidden @Enochen ?

hahnbeelee (approved):
LGTM!
I tested basically all frontend functionality. Bug that I pointed out during the meeting is no longer happening.

benjamin-shen (approved):
Colors work really well! I'm super happy with them.

zachary0kent (approved):
Looks great! Also can't recreate the bug that we fixed today.

PR #589, opened 2021-11-16
Add AS Major

zachary0kent:
### Summary 

This PR adds the Atmospheric Sciences Major. It was fairly simple to implement, as the requirements for AS were fairly straightforward. The only caveat is that Cornell lists "suggested" electives, and it's unclear which others (if any) could be substituted. For now, you just have to select from those Cornell suggests.

### Test Plan 
Screen Shot 2021-11-15 at 11 43 55 PM


SamChou19815 (commented):
This looks good to me. Maybe we can just merge it as it is. Hopefully this problem can be solved later by arbitrary opt-in. What's your opinion, @einc

hahnbeelee (approved):
agree with Sam - lgtm!

PR #590, opened 2021-11-16
Add EAS Major

zachary0kent:
### Summary 

This PR adds the EAS major to. The requirements for the EAS major were very explicit, so I believe they should be accurately reflected here. The only issue I ran into is that one requirement needs to be fulfilled through both courses and requirements, and I'm unsure of the best way to express that using our current infrastructure.

### Test Plan 

Screen Shot 2021-11-16 at 2 21 12 AM


hahnbeelee:
> The only issue I ran into is that one requirement needs to be fulfilled through both courses and requirements, and I'm unsure of the best way to express that using our current infrastructure.

Which one are you talking about here?

hahnbeelee:
> The only issue I ran into is that one requirement needs to be fulfilled through both courses and requirements, and I'm unsure of the best way to express that using our current infrastructure.
Refer the engineering liberal arts requirement: https://github.com/cornell-dti/course-plan/blob/master/src/requirements/data/colleges/en.ts#L105-#L136

image

hahnbeelee (approved):
Thanks Zak! I think this looks solid to me - although I'm curious about what you decided to do for the requirement that has course + credit requirements?

PR #591, opened 2021-11-16
[easy] Limit cypress test concurrency

SamChou19815:
### Summary 

Our cypress test is not concurrency-safe, since the test will change data of the same account in the database. So tests of two PRs opened simultaneously will be interfering with each other, causing unnecessary failures.

This PR limits the number of concurrent running frontend tests to be 1.

### Test Plan 

See https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

### DIFF
--- a/.github/workflows/ci-check.yml
+++ b/.github/workflows/ci-check.yml
@@ -36,6 +36,7 @@ jobs:
         run: exit $(git status --porcelain | wc -l)
   frontend-tests:
     runs-on: ubuntu-latest
+    concurrency: cypress-tests
     steps:
       - uses: actions/checkout@master
       - name: Set up Node


.github/workflows/ci-check.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         run: exit $(git status --porcelain | wc -l)
   frontend-tests:
     runs-on: ubuntu-latest
+    concurrency: cypress-tests
```
benjamin-shen, line 39: Is `cypress-tests` just an arbitrary environment name?

SamChou19815: Yep

benjamin-shen (approved):

PR #592, opened 2021-11-16
Placeholder courses

willespencer:
### Summary 

This pull request adds placeholder courses to CoursePlan to eventually support templates.

- [x] Add placeholder courses that display when Firestore data added (see below)
- [x] Style appropriately (normal/compact view, hover states, etc.)
- [x] Add `FirestoreSemesterPlaceholder` type and make the semester course array have objects of either that type of `FireStoreSemesterCourse`
- [x] Ensure placeholders can be dragged and save changes when refreshed 
- [x] Add cautions to placeholders when starting semester does not match current semester, and handle both semester order cases

image

Remaining TODOs:
- [ ] Click placeholder to open up a new course adding modal (future PR)

Remaining design questions
- [x] Does the compact view placeholder look ok?
- [x] Should hover state be on the entire card, or only the part right of the 6 dots (like it is now)?
- [x] What cursor should be used where? Pointer to look clickable, or hand to look draggable?
- [ ] How can the warning system be improved? Right now, it is hardcoded to check if the semester index matches where we think the placeholder should be, but this does not handle if users ever have differing numbers of semesters than what we expect (like winter and summer semesters, first year transfer, do not add all their old semesters, etc.).

### Firestore Schema 
Placeholders are stored in the list of courses in a given semester instead of being their own field, to ensure that they can be dragged and dropped anywhere on a user's plan (like between 2 normal courses. The schema of a placeholder generally matches that of the template data files (as seen in #564). The only fields that do anything right now are uniqueID, name (shows up on the placeholder), and startingSemester (determines if a warning shows up).
image

### Test Plan 

1. Add a placeholder to a courses array in a semester following the above schema.
2. Confirm it can be dragged and dropped to other semesters while saving
3. Confirm a warning displays if the startingSemester prop does not match the semester the placeholder is in, on both "Newest" and "Oldest" toggles.
3. Confirm the design of the placeholder and warnings look good on both views, including hover states
4. Confirm that code changed to handle different types still works as expected, specifically...
- Credit counts on semesters are still accurate
- Placeholders are minimized when semesters are, and cleared when semesters are
- Selectable requirements work (and do not show placeholders in the dropdown)
- Colors for academic subjects can be changed
- Other warnings still work

### Notes 

Placeholders will only show up if data is manually added to the semesters Firestore collection, so this PR can be safely merged without a gatekeep


hahnbeelee:
first off, 😩 this PR summary is so good🤌✊

hahnbeelee:
hmm weird thing happened where I dragged into a semester and it didn't get associated with a requirement.
image

einc:
very exciting!!! looks pretty good overall!!! 
i think whatever queen designer kehui decides for the design questions should be good! 

- super nitpicky, but i did notice when i drag using the the 6 dots, the blue outline doesn't show. maybe we can have it for consistency if possible! 

Screen Shot 2021-12-01 at 1 32 02 AM

- I think the warnings are fine for now, once we start using real data we can prob give more context like it's recommended to take this course in sem __ bc of ___

einc:
> hmm weird thing happened where I dragged into a semester and it didn't get associated with a requirement. image

i think we have a bug on prod rn so this isn't specific to this PR!

kkkehui:
Hey Will, here are the answers to the design questions.
*  Does the compact view placeholder look ok?
- Yes! Sorry, I forgot to design the compact placeholder… But it looks really good! 

*  Should hover state be on the entire card, or only the part right of the 6 dots (like it is now)?
- I think the current hover state looks good! It is consistent with the hover state design for existing course cards. 

*   What cursor should be used where? Pointer to look clickable, or hand to look draggable?
- I think using hand would make more sense since the hovering state already suggests the placeholder is clickable. And we are using hand for the existing course card. So, hand!

  *   How can the warning system be improved? 
- What I was thinking for the warning system is that only tell the user “This requirement is suggested to be fulfilled in this semester.“ Or “this requirement is suggested to be fulfilled in nth semester.” Just like what it is now. I think it is better to keep this simple since users can always use the requirement bar to check their progress and evaluate their course choices. 

Hope I answered your questions! Lmk if you have any other concerns! Thanks!

src/components/Semester/Semester.vue
```diff
@@ -226,10 +235,11 @@ export default defineComponent({
 
   computed: {
     coursesForDraggable: {
-      get(): readonly FirestoreSemesterCourse[] {
+      get(): readonly (FirestoreSemesterCourse | FirestoreSemesterPlaceholder)[] {
         return this.courses;
       },
       set(newCourses: readonly AppFirestoreSemesterCourseWithRequirementID[]) {
+        // TODO: does this need to be updated? probably
         const courses = newCourses.map(({ requirementID: _, ...rest }) => rest);
```
hahnbeelee, line 243: @benjamin-shen Looks like you wrote this code. What's the reasoning for leaving out `requirementID`?

src/utilities.ts
```diff
@@ -151,3 +151,9 @@ export function getReqColor(groupName: string, onboardingData: AppOnboardingData
   }
   return reqGroupColorList[0];
 }
+
+// Determines whether the given element in a FireStoreSemester list is a Placeholder or not
+export const isPlaceholderCourse = (
+  element: FirestoreSemesterPlaceholder | FirestoreSemesterCourse
+): element is FirestoreSemesterPlaceholder =>
+  !!(element as FirestoreSemesterPlaceholder).startingSemester;
```
hahnbeelee, line 159: Can you explain this?

hahnbeelee (approved):
* Test plan works
* Code looks good! - I left some clarifying comments

Everything looks fine, but I do have an overarching concern about the introduction to templates in hopes that it won't affect the requirements algorithm. The edits you had to put in were quite minimal here, so there isn't much to judge. I hope nothing comes up though along the way.

src/components/Course/CourseCaution.vue
```diff
@@ -27,56 +39,101 @@
...
+      if (isPlaceholderCourse(this.course)) {
+        switch (this.course.startingSemester) {
+          case 1:
+            return '1st';
+          case 2:
+            return '2nd';
+          case 3:
+            return '3rd';
+          default:
+            return `${this.course.startingSemester}th`;
```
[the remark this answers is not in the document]
SamChou19815: Unfortunately this is still wrong as it does not get things like 11th and 21st wrong. The right way to do it is by `Intl` API supported by modern browsers (I don't think we care about IE): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules/PluralRules#using_options

hahnbeelee:
As for your design concerns, I know Kehui already addressed them, but I also wanted to address :)
* I agree with Kehui that the grabbing hand would make more sense
* For the semester logic - right now it's dependent on the number of semesters you have. So I could have FA2019, SP20, FA20 and it'll think that it belongs in FA20 if it's supposed to be in the 3rd semester. I think storing it as being required to be in the 3rd semester is still good, but I think that "3rd semester" should be defined by the user's entrance year. Vaish made logic for pre-populating the semesters in accordance to the user's entrance year. So for example if I say I entered 2018 (you can assume it's fall) the 3rd semester will be FA20.
If FA20 isn't in the schedule they'll have to add it. Maybe the warning could also say "This course should be taken in FA20 (your 3rd semester)" So that they know. I know this increases logic but I feel like it helps the user more/ is more accurate. Thoughts?

src/components/Semester/Semester.vue
SamChou19815: I think I wrote it. The courses with `requirementID` is dragged from requirements bar. It's used to indicate which requirement the user wants the course to match to.

src/components/Course/CourseCaution.vue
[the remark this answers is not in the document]
hahnbeelee: LOL

src/components/Semester/Semester.vue
willespencer: Sorry I didn't mean to leave this comment in there! I think this should be fine for now.

src/utilities.ts
willespencer: I got this idea from Stackoverflow, but basically it takes in an element of either type of object that can be in a semester courses list (placeholder or course), interprets it as a placeholder, and checks to see if a field unique to placeholders exists on it (like `startingSemester`). If it does, it returns true, otherwise false. 

More documentation here: https://www.typescriptlang.org/docs/handbook/advanced-types.html

willespencer:
Resolved all comments, including fixing the dragging and hover state issues mentioned above!

> * For the semester logic - right now it's dependent on the number of semesters you have. So I could have FA2019, SP20, FA20 and it'll think that it belongs in FA20 if it's supposed to be in the 3rd semester. I think storing it as being required to be in the 3rd semester is still good, but I think that "3rd semester" should be defined by the user's entrance year. Vaish made logic for pre-populating the semesters in accordance to the user's entrance year. So for example if I say I entered 2018 (you can assume it's fall) the 3rd semester will be FA20.
>   If FA20 isn't in the schedule they'll have to add it. Maybe the warning could also say "This course should be taken in FA20 (your 3rd semester)" So that they know. I know this increases logic but I feel like it helps the user more/ is more accurate. Thoughts?

I think this is way better than what I have!! Going to put this in a follow-up item for when we better define the warnings, thanks Hahnbee!

hahnbeelee (approved):
Changes LGTM!

PR #593, opened 2021-11-16
Add semester season property from semester type

benjamin-shen:
### Summary 

This pull request is a script to migrate all old instances of semester type to semester season. After this, the schema will include identical values for `type` and `season` (instead of just including `type`). I ran it on dev, it just needs to be run on prod.

Depends on #546 

### Test Plan 

`npm run ts-node -- src/requirements/admin/semester-season-migration <user email>` to dry-run on a single user.  
`npm run ts-node -- src/requirements/admin/semester-season-migration --run-on-db` to run on the entire database.  

Pasted Graphic

### DIFF
--- a/src/requirements/admin/semester-season-migration.ts
+++ b/src/requirements/admin/semester-season-migration.ts
@@ -1,0 +1,48 @@
+/* eslint-disable no-console */
+
+import { usernameCollection, semestersCollection } from '../../firebase-admin-config';
+
+/**
+ * Perform migration of semester type to semester season
+ */
+async function runOnUser(userEmail: string, runOnDB: boolean) {
+  const semesters = await semestersCollection
+    .doc(userEmail)
+    .get()
+    .then(it => {
+      const data = it.data();
+      return (data && data.semesters) || [];
+    });
+
+  console.group(`Running on ${userEmail}...`);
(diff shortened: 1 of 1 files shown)


src/requirements/admin/semester-season-migration.ts
```diff
@@ -0,0 +1,48 @@
...
+    .get()
+    .then(it => {
+      const data = it.data();
+      return (data && data.semesters) || [];
+    });
+
+  console.group(`Running on ${userEmail}...`);
+  const newSemesters = semesters.map(semester =>
+    // each semester should have type but not necessarily season
+    ({ ...semester, season: semester.season || semester.type })
```
SamChou19815, line 20: maybe also do some cleanup to remove the old type field?

benjamin-shen: There's currently still a dependency on `type` in the code. There was probably a better way to do it 😭 but I plan on doing the cleanup in a future script.

benjamin-shen: Oh, I remember why I did it this way. It was because I didn't want to update the database schema while prod and other devs were still using `type`, so I split up the migration into multiple parts

SamChou19815 (approved):

PR #594, opened 2021-11-16
Fix dragging courses to semester due to race condition

SamChou19815:
### Summary 

Race condition since each course is trying to run some async code

### Test Plan 

https://user-images.githubusercontent.com/4290500/142082585-0add63b5-3b57-40aa-bad8-afa471818700.mov


willespencer (approved):
Looks fixed! Impressed by the speed 🚤

PR #596, opened 2021-11-17
Release containing only #575

SamChou19815:
### Summary 

A small release that only contains #575 that changes the engineering liberal studies requirement. This PR's merge is delayed since we need to run migration scripts first. Now that the migration script is finished. It's time to release.

Edit: also contains the PR to delete migration scripts, which shouldn't impact end users.

### Test Plan 

👀


hahnbeelee (approved):
Compared my engineering requirements between the deployment link and master. Seemed in sync.

benjamin-shen (approved):
😄

willespencer (approved):
LGTM! My liberal studies data works for each part of the requirement, courses cannot be double-counted with other requirements, and not fulfilling 1 part of the computed req leaves it unfulfilled.

PR #600, opened 2021-11-18
Added the template file for the ECE major

Aryan77:
### Summary 

This pull request adds the template data for the ECE major, closely following the implementation of the CS major template.

- [x] fixed a very small and (so far) insignificant typo in the ECE major file - an extra space in the description that would have resulted in a bug when running the script that constructs the actual templates on the front end, as said script would involve exactly matching between descriptions in a major file and names in the corresponding template file.

Remaining TODOs:

- [ ] add further templates for other majors.

### Test Plan 

Compare the data file with the ECE FA2021 template to ensure all requirements are correctly matched to where placeholders should be.

Screen Shot 2021-11-18 at 13 31 40

### DIFF
--- a/src/requirements/data/templates/ece.ts
+++ b/src/requirements/data/templates/ece.ts
@@ -28,16 +28,28 @@ const eceTemplate: Template = [
   },
   {
     reqGroup: 'EN',
-    name: 'Advanced Programming/Computer Engineering',
-    placeholderSemesters: [3],
+    name: 'Intro to Engineering',
+    placeholderSemesters: [1],
   },
   {
     reqGroup: 'EN',
+    name: 'Computing from engineering',
+    placeholderSemesters: [2],
+  },
+  /*
+  {
+    reqGroup: 'ECE',
+    name: 'Advanced Programming/Computer Engineering',
+    placeholderSemesters: [3],
+  },
+  */
+  {
+    reqGroup: 'ECE',
     name: 'ENGRD 2300',
     placeholderSemesters: [3],
   },
   {
-    reqGroup: 'EN',
+    reqGroup: 'ECE',
     name: 'Core Courses',
     placeholderSemesters: [3, 4],
   },
@@ -52,34 +64,34 @@ const eceTemplate: Template = [
     placeholderSemesters: [3, 4, 5, 6, 7, 8],
   },
   {
-    reqGroup: 'EN',
+    reqGroup: 'ECE',
     name: 'Outside ECE Technical Electives',
     placeholderSemesters: [5, 6, 7],
   },
   {
-    reqGroup: 'EN',
+    reqGroup: 'ECE',
     name: 'Culminating Design Experience (CDE)',
     placeholderSemesters: [7],
   },
   {
     reqGroup: 'EN',
     name: 'Advisor Approved Electives',
-    placeholderSemesters: [3, 4],
+    placeholderSemesters: [5, 6],
   },
   {
-    reqGroup: 'EN',
+    reqGroup: 'ECE',
     name: 'Foundation Courses',
     placeholderSemesters: [5, 6],
   },
   {
-    reqGroup: 'EN',
+    reqGroup: 'ECE',
     name: 'Upper-Level ECE Electives',
-    placeholderSemesters: [6],
+    placeholderSemesters: [6, 7, 8],
   },
   {
-    reqGroup: 'EN',
+    reqGroup: 'ECE',
     name: 'Upper-Level ECE Electives: 4000 level',
-    placeholderSemesters: [7, 8],
+    placeholderSemesters: [8],
   },
 ];
 
--- a/src/requirements/data/templates/ece.ts
+++ b/src/requirements/data/templates/ece.ts
@@ -14,7 +14,12 @@ const eceTemplate: Template = [
   {
     reqGroup: 'EN',
     name: 'Physics',
-    placeholderSemesters: [2, 3, 4],
+    placeholderSemesters: [2, 3],
+  },
+  {
+    reqGroup: 'ECE',
+    name: 'Physics 2214/2218',
+    placeholderSemesters: [4],
   },
   {
     reqGroup: 'EN',
@@ -33,7 +38,7 @@ const eceTemplate: Template = [
   },
   {
     reqGroup: 'EN',
-    name: 'Computing from engineering',
+    name: 'Computing',
     placeholderSemesters: [2],
   },
   /*
--- a/src/requirements/data/templates/ece.ts
+++ b/src/requirements/data/templates/ece.ts
@@ -1,0 +1,86 @@
+import { Template } from '@/requirements/types';
+
+const eceTemplate: Template = [
+  {
+    reqGroup: 'UNI',
+    name: 'Physical Education',
+    placeholderSemesters: [1, 2],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Mathematics',
+    placeholderSemesters: [1, 2, 3, 4],
+  },
+  {
+    reqGroup: 'EN',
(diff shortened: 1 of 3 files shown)


src/requirements/data/templates/ece.ts
```diff
@@ -0,0 +1,86 @@
...
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Mathematics',
+    placeholderSemesters: [1, 2, 3, 4],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Physics',
+    placeholderSemesters: [2, 3, 4],
```
willespencer, line 17: I think this should just be [2, 3] as the engineering requirement only checks PHYS 1112 and PHYS 2213

src/requirements/data/templates/ece.ts
```diff
@@ -0,0 +1,86 @@
...
+    name: 'Chemistry',
+    placeholderSemesters: [1],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'First-Year Writing Seminars',
+    placeholderSemesters: [1, 2],
+  },
+  {
+    reqGroup: 'EN',
```
willespencer, line 30: I think the reqGroup for this requirement and most of the ones below should be `ECE` since they are from that data file and not the engineering file

src/requirements/data/templates/ece.ts
```diff
@@ -0,0 +1,86 @@
...
+    name: 'Advanced Programming/Computer Engineering',
```
willespencer, line 31: @einc I believe this requirement is a check-off requirement as it does not show up in the engineering handbook flowchart and says this on the linked website in our ece file. If that is correct, @Aryan77 could you set the requirement in the ECE file to be double countable, and comment out this object in this templates file?

image

src/requirements/data/templates/ece.ts
```diff
@@ -0,0 +1,86 @@
...
+    placeholderSemesters: [3],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'ENGRD 2300',
+    placeholderSemesters: [3],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Core Courses',
```
willespencer, line 41: This template does not match the reqs file, but I think you have the most up to date information. @einc it looks like the current ECE requirement just has core courses ECE 2300 and ECE 2720, should we update the reqs file?

image

src/requirements/data/templates/ece.ts
```diff
@@ -0,0 +1,86 @@
...
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Culminating Design Experience (CDE)',
+    placeholderSemesters: [7],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Advisor Approved Electives',
+    placeholderSemesters: [3, 4],
```
willespencer, line 67: I think these should be semester 5 and 6 based on the chart?

src/requirements/data/templates/ece.ts
```diff
@@ -0,0 +1,86 @@
...
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Upper-Level ECE Electives',
+    placeholderSemesters: [6],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Upper-Level ECE Electives: 4000 level',
+    placeholderSemesters: [7, 8],
```
willespencer, line 82: I think both the template and reqs data are wrong for ECE electives. I think there should be 3 upper-level ece electives (semesters 6, 7, 8), 2 upper-level ECE electives (both in semester 8), and 1 CDE course (accurate). If @einc agrees, could you update the major and the template?

image

src/requirements/data/templates/ece.ts
```diff
@@ -0,0 +1,86 @@
...
+  },
+];
+
+export default eceTemplate;
```
willespencer, line 86: I think some requirements are missing from just templates
- Computing from engineering in Semester 2
- Intro to engineering in Semester 1

Additionally, I think some data is missing from the reqs file and templates. @einc do you agree with all of these, and if so, could you update them Aryan?
- Phys 1110 for students starting in Fall 2021 (not sure how we only would require for some students though).
- PHYS 2214 or 2218 in semester 4

willespencer (changes requested):
Looks really good! Just left some comments where some requirements are missing from templates, or places where the requirements file does not match templates/the requirements on the ECE website. If those all get in sync, this should be good to go 😄

src/requirements/data/templates/ece.ts
Aryan77: Yes, there were a number of inconsistencies between what the handbook says and what the requirements in the ECE file says. I wrote the template to match exactly with the data we currently had - I will update both the ECE major file and the template files accordingly. Thanks!

src/requirements/data/templates/ece.ts
einc: Yea, would you guys mind updating the ECE file as well? They haven't been updated in a while, thank you!

src/requirements/data/templates/ece.ts
Aryan77: Yes, the data in the major file seems to suggest 1 elective and 3 4000+ electives, which I tried to match. It does seem to be out of date however so I will update the major file too.

src/requirements/data/templates/ece.ts
Aryan77: The current placeholderSemesters for Physics is [2, 3, 4], because I thought I would include PHYS 2214 along with the other 2 physics classes. Would this not work?

Aryan77:
I updated the template file to address most of Will's concerns - focusing exclusively on being true to the handbook since I realize the actual major file was outdated. 
I will make a separate PR to update the ECE major file, and if I come across some inconsistency with the template file will update that as well.

src/requirements/data/templates/ece.ts
willespencer: I don't think so, because Physics is an Engineering-wide requirement, but the PHYS 2214/2218 requirement is specific to ECE. I think we would need a new PHYS 2214/2218 checker in the ECE file.

src/requirements/data/templates/ece.ts
```diff
@@ -28,16 +28,28 @@ const eceTemplate: Template = [
   },
   {
     reqGroup: 'EN',
+    name: 'Intro to Engineering',
+    placeholderSemesters: [1],
+  },
+  {
+    reqGroup: 'EN',
+    name: 'Computing from engineering',
```
willespencer, line 36: Sorry, should have been more specific - this requirement should just be called "Computing" !

willespencer (changes requested):
Left 1 small change to fix before merging, otherwise this should be good to go for now! Just need to address the rest of the data/ECE discrepancies commented on in a separate PR

src/requirements/data/templates/ece.ts
willespencer: Note: Not going to handle PHYS 1110 ourselves as this is a broader eng change that will be fixed by other devs

src/requirements/data/templates/ece.ts
willespencer: After talking to @einc - it looks like for pre-2021 students we are going to want a total of 5 electives, but 6 electives for post 2021 (per the handbooks). We can talk about how to do this (toggleable requirements probably) after break!!

src/requirements/data/templates/ece.ts
willespencer: Looks like the ECE core courses changed in 2021 - similarly to electives, will probably want toggleable requirement depending on class year!

willespencer (approved):
Looks good! Will talk about updating the format and updating the ECE file offline

PR #601, opened 2021-11-18
Convert TrackUsers to script to store data in Firestore daily

willespencer:
### Summary 

This pull updates TrackUsers to be easier to work with and read data per Sam's suggestion here #565. 

- [x] Convert TrackUsers to a regular script (similar to other ones we run). Now we do not need to worry about issues deploying to Firebase or testing it on Google Cloud.
- [x] Store TrackUsers data in a new Firestore collection named `track-users`, where each document has the output data of the function at a given timestamp
- [x] Write a Github action to run the script daily at 4am. Now we do not need to manually run it, we can just reference the newest document.

image

Remaining TODOs:

- [ ] Test the Github action to make sure the job writes to Firestore correctly
- [x] Add the `track-users` collection to prod (not just dev) and add security rules if necessary before merging

### Test Plan 

Run `npm run track-users` and confirm a new document is created on the dev Firestore with the current timestamp and correct user data summary.

### Notes 

Two potential future PRs...
1. I plan on making a route on courseplan.io to display the newest `track-users` document so that non-devs can access the information easier!
2. I could make a PR to remove Firebase functions from CoursePlan as we will not be using it anymore. Not sure if I think I should do this though, because if we ever want to use it in the future, it might be hard to re-set up.


package.json
```diff
@@ -20,7 +20,8 @@
     "test": "jest",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "cypress:e2e": "npm run build:dev && concurrently \"npm run serve:build\" \"wait-on http-get://localhost:8080 && npm run cypress:run\" --kill-others --success first"
+    "cypress:e2e": "npm run build:dev && concurrently \"npm run serve:build\" \"wait-on http-get://localhost:8080 && npm run cypress:run\" --kill-others --success first",
+    "track-users": "npm run ts-node -- src/requirements/admin/track-users.ts"
```
SamChou19815, line 24: Maybe move it up to keep all the scripts that are run by `ts-node` together

SamChou19815 (approved):
LTGM. Do we want to also remove the code for the old script?

PR #602, opened 2021-11-26
[easy] Fix enforce release workflow job

benjamin-shen:
### Summary 

This pull request fixes the `enforce-release-workflow` job in the `ci-policies` workflow. Before, it was failing on PRs from `master` to any non-`beta-release` branch. Rather, it should fail on PRs from non-`master` branches to `beta-release`.

### Test Plan 

before: [ci passed](https://github.com/cornell-dti/course-plan/runs/4337584044?check_suite_focus=true) (should fail)
image
image

after: [ci failed](https://github.com/cornell-dti/course-plan/runs/4337552051?check_suite_focus=true) (should fail)
image
image

### Notes
(not related to this PR)

I discovered a bug with Github where if you make a PR `A=>B` and then a new PR `A=>C`, the workflow instance for the second PR overwrites the workflow instance for the first PR. Then, when you delete the second PR and rerun the workflow instance for the first PR, the PR-specific env variables (like base ref and head ref) don't update properly. Editing the PR didn't fix the issue, but closing and reopening the PR (which created new workflow instances) fixed it.

### DIFF
--- a/.github/workflows/ci-policies.yml
+++ b/.github/workflows/ci-policies.yml
@@ -11,7 +11,7 @@ jobs:
           echo "Your head ref is ${{ github.head_ref }}."
           echo "Your base ref is ${{ github.base_ref }}."
       - name: Fail if try to push release from non-master branch
-        if: github.base_ref == 'beta-release' && github.head_ref != 'master'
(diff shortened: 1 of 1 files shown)


SamChou19815 (approved):
LGTM

PR #595, opened 2021-11-17
Remove semester type from database

benjamin-shen:
### Summary 

This pull request cleans up the semesters collection by removing the `type` property. Should be merged and run on prod after #604 is deployed to prod.

### Test Plan 

`npm run ts-node -- scripts/migration/semester-season-migration <user email>` to dry-run on a single user.  
`npm run ts-node -- scripts/migration/semester-season-migration --run-on-db` to run on the entire database.

Pasted Graphic 1

PR #604, opened 2021-11-29
Remove semester type from codebase

benjamin-shen:
### Summary 

This pull request removes the `type` property from the codebase. This should only be deployed to prod after #593 is run on prod.

- [x] removed semester type

Remaining TODOs:

- [ ] test, merge, and run #595

### Test Plan 

Make sure that add-semester functionality still works.
- previously added semester (still with `type` property)
- newly added semester (without `type` property)

image
image


SamChou19815 (approved):
LGTM

PR #607, opened 2021-11-30
Add analytics data to CoursePlan route

willespencer:
### Summary 

This pull request adds the most recent TrackUsers analytics data directly to a new route on CoursePlan (courseplan.io/analytics) designed similarly to the 404 page. This should allow anyone on CP to see the daily prod runs (#601) without needing permissions to the actual prod Firestore.

- [x] Created an analytics route (`/analytics`)
- [x] Pretty print most recent TrackUsers document
- [x] Add a timestamp so people know when it was last run
- [x] Hide data (and style route differently) for those not on CP
- [x] Add timestamp field to documents, optimize code to only retrieve doc with newest timestamp 

image

Remaining TODOs:

- [x] Add the `CPWhitelist` collection to prod and update it so all current team members are on it
- [x] Add the `CPWhitelist` rule to prod (see below) so data can only be read by those on the whitelist

image

### Test Plan 

1. Ensure that the most recent data run is shown with a timestamp
2. Confirm the /analytics route cannot be accessed unless logged in.
3. Confirm route does not show Firestore information when logged in to an email of someone not on `CPWhitelist` (and is nicely styled).

image

### Notes 

I unsuccessfully tried to block users not on the CPWhitelist from accessing the new `/analytics` route all together. If that would be significantly more desirable, let me know and I can further investigate this option! But the Firestore rules should prevent anyone who should not have access from having it.


src/containers/Analytics.vue
```diff
@@ -0,0 +1,108 @@
...
+            newestDocDate = date;
+            newestDocData = doc.data();
+          }
+        });
+
+        const output = JSON.stringify(newestDocData, null, 2);
+        this.analyticsData = output;
+        this.analyticsTimestamp = newestDocDate.toLocaleString();
+      });
+    },
```
benjamin-shen, line 51: Can we refactor this function, eg. to `global-firestore-data`?

Also, you can potentially make this more efficient: https://stackoverflow.com/a/54178993

willespencer: Refactored and made more efficient, thanks Ben!!

src/requirements/admin/track-users.ts
```diff
@@ -32,7 +32,7 @@ function isOld(semester: FirestoreSemester) {
   if (semester.year < year) {
     return true;
   }
-  if (seasonToMonth(semester.season) <= month) {
+  if (semester.season && seasonToMonth(semester.season) <= month) {
```
benjamin-shen, line 35: @willespencer can we investigate why/where `season` is a missing property? This check shouldn't be necessary.

src/containers/Analytics.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+  <div>
+    <div class="body-container" :class="{ 'no-data': !hasData() }">
```
hahnbeelee, line 3: if there's no data can you say "rip no data or there was an error" something of the sorts

src/requirements/admin/track-users.ts
hahnbeelee: Can you actually add a comment for what this function is? Idk what old means and what it's referring to

hahnbeelee (changes requested):
Wait I was able to access it without logging in :o

hahnbeelee:
It doesn't show up for when I'm logged in as someone else though!

image

src/requirements/admin/track-users.ts
willespencer: @benjamin-shen investigated and talked offline - at least one dev user still does not have a season property, but all prod users do. Solution is to re-run the script on dev.

src/containers/Analytics.vue
willespencer: The reason why I don't do this is to avoid showing that message to users that are not on CoursePlan, who we might want to avoid even knowing this sort of data exists. Let me know if there is an appropriate message that should show here though!

willespencer:
> Wait I was able to access it without logging in :o https://cornelldti-courseplan-dev--pr607-analytics-route-htahvfos.web.app/analytics

Are you sure you weren't already logged in / the log in wasn't cached? If from `courseplan.io/analytics` you are able to switch your url to `courseplan.io` without being redirected to the login page, then you should be able to have access to the analytics route.

src/router/index.ts
```diff
@@ -27,6 +28,14 @@ const router: Router = createRouter({
       name: 'Policy',
       component: Policy,
     },
+    {
+      path: '/analytics',
+      name: 'Analytics',
+      component: Analytics,
+      meta: {
+        requiresAuth: true,
```
benjamin-shen, line 36: If you could modify this to be something like `requiresWhitelist`, I think that'd be more ideal than throwing an error due to firestore rules.

benjamin-shen (approved):
LGTM! I tried logging in as someone not on the whitelist and the data didn't show.

I have one comment about the auth handling, but it can be addressed in a future PR.

I think we should add a Firestore rules file in the codebase to make it easier to deploy those rules to dev/prod. This can also be done later.

hahnbeelee (approved):
ok it isnt happeneing anymore - maybe i was tripping haha

PR #609, opened 2021-11-30
Use service account secret when provided

willespencer:
### Summary 

This pull request ensures the service account is used when provided as an environmental variable (attempted fix at TrackUsers failing in #608). Furthermore, makes track-users as prod so the prod firestore url is used.

### Test Plan 

1. Ensure running `npm run track-users` locally still uploads a Firestore doc to the dev instance (see [here](https://console.firebase.google.com/u/0/project/cornelldti-courseplan-dev/firestore/data/~2Ftrack-users~2F2021-11-30T18:18:11.184Z))
2. Re-run the TrackUsers action when merged to see if the error is resolved and a Firestore doc is added to the prod instance.

### DIFF
--- a/src/firebase-admin-config.ts
+++ b/src/firebase-admin-config.ts
@@ -4,9 +4,10 @@ import * as admin from 'firebase-admin';
 import { getTypedFirestoreDataConverter } from './firebase-config-common';
 
 const serviceAccountFilename = process.env.PROD ? 'serviceAccountProd.json' : 'serviceAccount.json';
-const serviceAccount =
+const serviceAccountUnparsed =
   process.env.SERVICE_ACCOUNT ??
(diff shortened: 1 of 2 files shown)


src/firebase-admin-config.ts
```diff
@@ -4,12 +4,17 @@ import * as admin from 'firebase-admin';
 import { getTypedFirestoreDataConverter } from './firebase-config-common';
 
 const serviceAccountFilename = process.env.PROD ? 'serviceAccountProd.json' : 'serviceAccount.json';
-const serviceAccount = JSON.parse(
-  fs.readFileSync(path.join(__dirname, '..', serviceAccountFilename)).toString()
-);
+const serviceAccount =
+  process.env.SERVICE_ACCOUNT ??
```
SamChou19815, line 8: This also needs to be JSON parsed

SamChou19815 (approved):

PR #610, opened 2021-11-30
Avoid tracking users that have not finished onboarding

willespencer:
### Summary 

Fixes bug in #604 that breaks TrackUsers - some users still do not have a `season` property in the database, so not using the `type` parameter is season is null breaks some of the code.

This is because, as found by @benjamin-shen, users that do not finish onboarding are not in the usernames collection, and thus the script that added `season` param for every existing `type` param on a semester for users in the username collection (#593) does not add `season` to those who have not finished onboarding. 

To resolve this, there is now a check to ensure a user has finished onboarding (and is in the username collection) before tracking their other dats. This is a positive change regardless of the bug, as we only want to track semester and onboarding data for users that have actually gotten past onboarding.

### Test Plan 

Run TrackUsers locally to ensure it now works on dev without error, and that the number of users in the username collection matches the number of users counted in other collections (instead of being smaller like before)

image

### Notes 

I believe there needs to be a script run to add the `semester` field to users that have not finished onboarding, regardless of this PR fix. Otherwise,, after `type` is deleted from the codebase, if these users ever come back and finish onboarding, their semester view will be broken.

### DIFF
--- a/src/requirements/admin/track-users.ts
+++ b/src/requirements/admin/track-users.ts
@@ -77,6 +77,7 @@ function addYearToFrequencyDictionary(year: string, freqDict: Record<string, num
 /**
  * TrackUsers outputs user metrics to a Firestore document with timestap based on
  * data from the user-name and user-semesters Firestore collections.
+ * Users are only included if they have finished onboarding (and have a document in the username collection)
  *
  * It returns the total number of users (total-users),
  * the total number of semesters across all users (total-semesters),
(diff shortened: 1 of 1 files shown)


hahnbeelee (approved):
LGTM! Test plan checks out and the code looks good too!

Maybe in the future we can track the number of people who didn't complete onboarding to see how many people we're losing at onboarding - whatever numbers we get from that might be an incentive to improve the onboarding process 👀

PR #611, opened 2021-11-30
Specialize requirements

zachary0kent:
### Summary 

This pull request implements a framework for "specializing" college requirements for a specific major. For example, COE requires that MATH 1910, 1920, and one of 2930 or 2940 must be taken, whereas the CS major requires MATH 2940. This was accomplished by adding an optional `specializations` field to every major in `index.ts` which indicates how college requirements should be specialized for that major.

Remaining TODOs:

- [ ] handle conflicting specializations for double majors. Right now, all specializations are added to the College requirements, which could result in a requirement appearing twice but specialized differently. It seems somewhat difficult to automatically resolve any conflicting specializations in the general case. However, hopefully this should be fairly infrequent, and also doesn't seem to be a significant issue anyways--both specializations would would have to be completed as desired.

### Test Plan 
Below, we show specializing the COE Math requirement for CS:
Screen Shot 2021-11-30 at 5 29 46 PM

### DIFF
--- a/src/requirements/specialize.ts
+++ b/src/requirements/specialize.ts
@@ -14,10 +14,10 @@ export function specialized<R extends RequirementCommon>(
   collegeReqs: readonly R[],
   majors: readonly Major<R>[]
 ) {
-  const specializations = majors.flatMap(({ specializations }) => specializations ?? []);
-  const excluded = new Set(specializations.map(({ name }) => name));
+  const allSpecializations = majors.flatMap(({ specializations }) => specializations ?? []);
+  const excluded = new Set(allSpecializations.map(({ name }) => name));
   const filteredReqs = collegeReqs.filter(({ name }) => !excluded.has(name));
-  return [...filteredReqs, ...specializations];
+  return [...filteredReqs, ...allSpecializations];
 }
 
 /**
--- a/src/requirements/specialize.ts
+++ b/src/requirements/specialize.ts
@@ -35,6 +35,6 @@ export default {
       fulfilledBy: 'courses',
       perSlotMinCount: [1, 1, 1],
       slotNames: ['MATH 1910', 'MATH 1920', 'MATH 2940'],
-    } as CollegeOrMajorRequirement
-  }
+    } as CollegeOrMajorRequirement,
+  },
 };
--- a/src/requirements/specialize.ts
+++ b/src/requirements/specialize.ts
@@ -1,0 +1,20 @@
+import { Major } from './types';
+
+/**
+ * Many majors often "specialize" the requirements of their respective college(s)
+ * `specialized(collegeReqs, majors)` is an array of `collegeReqs` having been
+ * specialized according to `majors`
+ *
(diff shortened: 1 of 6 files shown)


hahnbeelee:
Can you outline what future work needs to be done for this?

src/requirements/specialize.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function specialized<R extends RequirementCommon>(
+  collegeReqs: readonly R[],
+  majors: readonly Major<R>[]
+) {
+  const allSpecializations = majors.flatMap(({ specializations }) => specializations ?? []);
+  const excluded = new Set(allSpecializations.map(({ name }) => name));
+  const filteredReqs = collegeReqs.filter(({ name }) => !excluded.has(name));
+  return [...filteredReqs, ...allSpecializations];
+}
```
SamChou19815, line 21: Ideally, we would like to preserve the original order of requirements in the college, so maybe:

```typescript
export function specialized<R extends RequirementCommon>(
  collegeReqs: readonly R[],
  majors: readonly Major<R>[]
) {
  const allSpecializations = majors.flatMap(({ specializations }) => specializations ?? []);
  const specializationMap = new Map(allSpecializations.map((it) => [it.name, it]));
  return collegeReqs.map((it) => specializationMap.get(it.name) ?? it);
}
```

SamChou19815 (commented):
Looks good with one nit

src/requirements/specialize.ts
```diff
@@ -0,0 +1,40 @@
...
+      description: 'MATH 1910, 1920, 2940, and a mathematics course chosen by the Major.',
+      source:
+        'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements',
+      checker: includesWithSubRequirements(['MATH 1910'], ['MATH 1920'], ['MATH 2940']),
+      fulfilledBy: 'courses',
+      perSlotMinCount: [1, 1, 1],
+      slotNames: ['MATH 1910', 'MATH 1920', 'MATH 2940'],
+    } as CollegeOrMajorRequirement,
+  },
+};
```
hahnbeelee, line 40: were you thinking that all specializations are defined here in this file?

hahnbeelee (changes requested):
Overall, looks good, but I was wondering if you could create a better infrastructure where the data for specializers are in a separate file within the data folder. I think your TO-DO is good and we can address that later, but I just want to make sure the current state of this PR is clean before merging.

zachary0kent:
> Overall, looks good, but I was wondering if you could create a better infrastructure where the data for specializers are in a separate file within the data folder. I think your TO-DO is good and we can address that later, but I just want to make sure the current state of this PR is clean before merging.

Yeah I agree. I factored out the specializations into a file in `data` and think it's much cleaner now.

hahnbeelee (approved):
Thanks Zak for the clean up!
@einc curious if you have other cases that you have on the top of your mind for this?

PR #615, opened 2021-12-05
Setup infrastructure for two-graph-based progress computation

SamChou19815:
#### Background

There are concerns over the current too liberal approach of computing requirement as proposed in the original plan above. In the previous proposal, we keep all edges by default even if there is double-counting and let user opt-out of those edges later.

As a response to that, we have explored ideas that the edges are not added by default. Instead, the user has to explicitly do opt-in for courses that might introduce double counting. This proposal indeed ensures that we are conservative about progress computation, but it also makes certain things hard to explain to users:

- why user has to make choices
- how to find potential constraint violation because an edge is added
- how to inform user about making choices without exposing all the graph internals

#### Amendament Solution

It turns out that we can try to find the best of both worlds by computing two graphs. The first graph is the graph we will compute according to the original proposal above. It will keep all edges, and user choices to remove double counting will be represented as opt-out. (Phase 3 computation). Then based on the graph, we remove all the edges that are associated with constraint violation, which results in a graph that only contain safe edges (Phase 4 computation).

Both graphs will be useful. Phase 3 graph helps us to compute constraint violations that can be used as source of suggestions and warnings for users, and phase 4 graph provides a source for progress computation that is conservative and safe.

#### Example

Consider the following example before we account for user choices. By default, all edges are kept, even if there are constraint violations (i.e. double counting).

Screen Shot 2021-11-02 at 21 47 33

Then we start to take account of user choices. In this case, the user opts-out the edge `CS5414 --- MEng_CS_15_Credits`. Now we get a graph that on the left of the following figure.

Screen Shot 2021-11-02 at 21 47 42

This graph still has constraint violations (edges marked in red), so we remove all the edges that are associated with constraint violations and obtain the graph on the right. The graph on the right will then by the source of truth for computing fulfillment progress.

### Implementation

This PR cherry picks the changes from #599 that involves updating the core requirement algorithm to use two graphs. The first graph applies user choices as before, and the second removes all edges that involve illegal double counting. 

Intuitively, the first graph is optimistic but dangerous: if we use the first graph to compute progress, we will overcount. However, it is still useful since it gives user guidance on what their current courses can fulfill in the most optimistic settings. The second graph is conservative and safe. It allows us to compute the real progress. The diff of the progress will be considered as uncertain progress, as shown in the proposed design in #599:

uncertain-progress

The core update of the algorithm is rather small, just a single added function `removeIllegalEdgesFromRequirementFulfillmentGraph` that generates the second graph from the first graph. However, it forces the following downstream changes:

- Since we now have two graphs, we must also compute progress for both graph. Changes in `src/requirements/requirement-frontend-computation.ts` are handling that. It's actually the largest diff in this PR.
- In a lot of Vue components that query the graph or render some progress info, they now need to decide whether to use the safe/dangerous graph/progress. In this PR, most of them use the safe graph, but some uses the dangerous graph. I will explain the dangerous graph usage in inline comments. In this PR, it matters less, since it doesn't introduce mechanism to let user introduce possible double counting yet. However, it will matter more in the future.

Apart from the requirement debugger, there shouldn't be any UI, functionality, and behavior changes. Requirement debugger should always show that safe graph is the same as dangerous graph.

## Test Plan

Added unit tests on `removeIllegalEdgesFromRequirementFulfillmentGraph`.

Updated requirement graph debugger to include both safe and dangerous graph:

Screen Shot 2021-12-05 at 11 49 04 AM

### DIFF
--- a/src/components/Requirements/RequirementDebugger.vue
+++ b/src/components/Requirements/RequirementDebugger.vue
@@ -1,27 +1,38 @@
 <template>
-  <svg class="debugger-svg" :width="900" :height="itemSize * 48 + 36">
-    <g
-      v-for="(requirement, index) in requirementGraphForDisplay.requirements"
-      :key="index"
-      @click="setOnlyNodeToKeepEdge(requirement)"
-    >
-      <rect :x="0" :y="index * 48 + 16" :width="300" :height="32" class="requirement-box"></rect>
-      <text :x="8" :y="index * 48 + 36" :font-size="16">{{ requirement }}</text>
-    </g>
-    <g
-      v-for="(course, index) in requirementGraphForDisplay.courses"
-      :key="index"
-      @click="setOnlyNodeToKeepEdge(course.uniqueId)"
-    >
-      <rect :x="600" :y="index * 48 + 16" :width="300" :height="32" class="course-box"></rect>
-      <text :x="608" :y="index * 48 + 36" :font-size="16" fill="white">{{ course.code }}</text>
-    </g>
-    <path v-for="(line, index) in lines" :key="index" :d="line" stroke="black" />
-  </svg>
+  <div>
+    <div class="graph-toggler">
+      <label class="graph-toggler-choice">
+        <input type="radio" value="dangerous" v-model="graphType" /> Dangerous Graph
+      </label>
+      <label class="graph-toggler-choice">
+        <input type="radio" value="safe" v-model="graphType" /> Safe Graph
+      </label>
+    </div>
+    <svg class="debugger-svg" :width="900" :height="itemSize * 48 + 36">
+      <g
+        v-for="(requirement, index) in requirementGraphForDisplay.requirements"
+        :key="index"
+        @click="setOnlyNodeToKeepEdge(requirement)"
+      >
+        <rect :x="0" :y="index * 48 + 16" :width="300" :height="32" class="requirement-box"></rect>
+        <text :x="8" :y="index * 48 + 36" :font-size="16">{{ requirement }}</text>
+      </g>
+      <g
+        v-for="(course, index) in requirementGraphForDisplay.courses"
+        :key="index"
+        @click="setOnlyNodeToKeepEdge(course.uniqueId)"
+      >
+        <rect :x="600" :y="index * 48 + 16" :width="300" :height="32" class="course-box"></rect>
+        <text :x="608" :y="index * 48 + 36" :font-size="16" fill="white">{{ course.code }}</text>
+      </g>
+      <path v-for="(line, index) in lines" :key="index" :d="line" stroke="black" />
+    </svg>
+  </div>
 </template>
 <script lang="ts">
 import { defineComponent } from 'vue';
 import store from '@/store';
+import RequirementFulfillmentGraph from '@/requirements/requirement-graph';
 
 type RequirementGraphForDisplay = {
   requirements: readonly string[];
@@ -30,9 +41,9 @@ type RequirementGraphForDisplay = {
 };
 
 function computeRequirementGraphForDisplay(
+  graph: RequirementFulfillmentGraph<string, CourseTaken>,
   onlyNodeToKeepEdge: string | number | null
 ): RequirementGraphForDisplay {
-  const graph = store.state.requirementFulfillmentGraph;
   const requirements = graph.getAllRequirements();
   const courses = graph.getAllCourses();
   const requirementToIndexMap = new Map(requirements.map((it, index) => [it, index]));
@@ -54,11 +65,11 @@ function computeRequirementGraphForDisplay(
   return { requirements, courses, edges };
 }
 
-type Data = { onlyNodeToKeepEdge: string | number | null };
+type Data = { onlyNodeToKeepEdge: string | number | null; graphType: 'safe' | 'dangerous' };
 
 export default defineComponent({
   data(): Data {
-    return { onlyNodeToKeepEdge: null };
+    return { onlyNodeToKeepEdge: null, graphType: 'dangerous' };
   },
   methods: {
     setOnlyNodeToKeepEdge(clickedNode: string | number) {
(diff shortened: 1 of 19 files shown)


src/components/Requirements/RequirementFulfillmentSlots.vue
```diff
@@ -118,10 +118,10 @@ export default defineComponent({
        * the nested requirement inside compound requirement the user chooses to display.
        */
       const matchedCourses = (
-        (this.requirementFulfillment.additionalRequirements || {})[
+        (this.requirementFulfillment.fulfillment.additionalRequirements || {})[
           this.compoundRequirementChoice
-        ] || this.requirementFulfillment
-      ).courses;
+        ] || this.requirementFulfillment.fulfillment
+      ).dangerousCourses;
```
SamChou19815, line 124: Here we are showing all the matching courses. We should use dangerous ones, since we want to show warning around those course cards. Using safe courses will eliminate the possibility of that.

src/components/Requirements/RequirementSelfCheckSlots.vue
```diff
@@ -39,7 +39,7 @@ export default defineComponent({
     fulfilledSelfCheckCourses(): readonly CourseTaken[] {
       // selectedCourses are courses that fulfill the requirement based on user-choice
       // they are taken from requirement graph
-      const selectedCourses = store.state.requirementFulfillmentGraph.getConnectedCoursesFromRequirement(
+      const selectedCourses = store.state.dangerousRequirementFulfillmentGraph.getConnectedCoursesFromRequirement(
         this.requirementFulfillment.requirement.id
       );
```
SamChou19815, line 44: Same as `RequirementFulfillmentSlots.vue` for using dangerous graph info.

src/components/Requirements/RequirementDebugger.vue
```diff
@@ -54,11 +65,11 @@ function computeRequirementGraphForDisplay(
   return { requirements, courses, edges };
 }
 
-type Data = { onlyNodeToKeepEdge: string | number | null };
+type Data = { onlyNodeToKeepEdge: string | number | null; graphType: 'safe' | 'dangerous' };
```
zachary0kent, line 68: Is there a reason we use a union type here as opposed to using an enumerated type / boolean?

src/components/Requirements/RequirementGroup.vue
```diff
@@ -148,12 +148,12 @@ export default defineComponent({
...
       const completed: RequirementFulfillment[] = [];
       this.req.reqs.forEach(req => {
-        if (req.minCountFulfilled < req.minCountRequired) {
+        if (req.fulfillment.safeMinCountFulfilled < req.fulfillment.minCountRequired) {
           ongoing.push(req);
         } else if (
-          req.additionalRequirements != null &&
-          Object.values(req.additionalRequirements).some(
-            it => it.minCountFulfilled < it.minCountRequired
+          req.fulfillment.additionalRequirements != null &&
```
zachary0kent, line 154: Could probably break up this expression a bit / factor it into a function for better readability

src/components/Requirements/RequirementHeader.vue
```diff
@@ -217,29 +217,37 @@ export default defineComponent({
     requirementFulfilled(): number {
       let fulfilled = 0;
       this.req.reqs.forEach(req => {
-        [req, ...Object.values(req.additionalRequirements || {})].forEach(reqOrNestedReq => {
```
zachary0kent, line 220: Can probably replace `||` with `??`

src/components/Requirements/RequirementHeader.vue
```diff
@@ -217,29 +217,37 @@ export default defineComponent({
...
+            if (reqOrNestedReq.safeMinCountFulfilled >= reqOrNestedReq.minCountRequired)
+              fulfilled += 1;
+          }
+        );
       });
       return fulfilled;
     },
     // number of requirements that can be fulfilled (so no pure self-checks)
     requirementTotalRequired(): number {
       let totalRequired = 0;
```
zachary0kent, line 231: Can probably simplify this accumulation with `fold`

src/requirement-types.d.ts
```diff
@@ -124,11 +124,27 @@ type RequirementFulfillmentStatisticsWithCoursesWithAdditionalRequirements = Req
     readonly [name: string]: RequirementFulfillmentStatisticsWithCourses;
   };
 };
+type MixedRequirementFulfillmentStatistics = {
```
zachary0kent, line 127: can probably use the `Readonly` type constructor here as opposed to explicitly marking every field

src/requirements/requirement-frontend-utils.ts
```diff
@@ -501,3 +503,49 @@ export function getRelatedUnfulfilledRequirements(
   }
   return { directlyRelatedRequirements, selfCheckRequirements };
 }
+
+export function getAllEligibleRequirements(
```
zachary0kent, line 507: This function contains conditionals & loops that are fairly deeply nested--I think it might be more readable to factor some of this code into separate functions

zachary0kent (approved):
Looks great overall! The idea of using 2 graphs is quite clever, and I think it allows us to reach a good compromise. Just left a few nits.

src/requirements/requirement-frontend-utils.ts
SamChou19815: Don't think refactors can help a lot. These logic requires to be colocated together...

src/components/Requirements/RequirementDebugger.vue
SamChou19815: Mostly to make the intention clear, and be consistent in naming with the rest of the system

src/requirement-types.d.ts
SamChou19815: This is trying to keep consistent with rest of the code

PR #616, opened 2021-12-06
Dashboard entrypoint

bungaepyo:
### Summary

This pull request implements the new navbar design and adds the entry point for the tools page. Tools page is essentially a dashboard-type page with features such as progress tracker, advisors, calendar, links, and exports. Specific features within the tools page will be implemented in the future by another dev (good luck). The navigation between the plan page (originally the home page) and the tools page is done through conditional displaying rather than routing because there are no plans for adding additional pages in the navigation bar. Profile page is kept as a modal.

- [x] add new icons
- [x] add entrty point for tools
- [x] mobile responsiveness

Remaining TODOs:

- [ ] swap mobile plan icon

### Test Plan 

Screen Shot 2021-12-05 at 6 04 09 PM
Screen Shot 2021-12-05 at 6 04 24 PM


einc:
😍 🥳 🤯 🤠
THIS IS EXCITING !!!!!! LOOKS GREAT TOBY !!!!!

one small thing, when i test mobile if I click on "Tools" then "View Schedule", the page doesn't change. it seems to change when i click on it in a different order tho 

https://user-images.githubusercontent.com/14082828/144842323-65fa4b24-19d9-486a-bf49-faa3cf2a7faf.mov

hahnbeelee:
> 😍 🥳 🤯 🤠 THIS IS EXCITING !!!!!! LOOKS GREAT TOBY !!!!!
> 
> one small thing, when i test mobile if I click on "Tools" then "View Schedule", the page doesn't change. it seems to change when i click on it in a different order tho
> 
>  view.schedule.mov

Oh man mobile is a hot mess

src/components/NavBar.vue
```diff
@@ -187,6 +252,15 @@ $mobile-navbar-height: 4.5rem;
     }
   }
 
+  // Need to add plan-mobile-icon
```
hahnbeelee, line 255: looks like you can remove this comment

src/components/ToolsPage/ToolsContainer.vue
```diff
@@ -0,0 +1,45 @@
+<template>
```
hahnbeelee, line 1: Instead of this being in `src/components/ToolsPage/ToolsContainer.vue` I think you should just call it `Tools.vue` and put it under the `/containers` folder

hahnbeelee (changes requested):
Ayo overall LGTM! One request is to just hide the entrypoint under a feature flag!

hahnbeelee (approved):
Added the feature flag lol 😁

benjamin-shen (approved):
lgtm to merge!

A few notes for the future
UI: can we have an indication for which page we're on?
Implementation: it feels a bit hacky to conditionally render the different components in `Dashboard.vue`, I think we should refactor everything using routing.

hahnbeelee:
> can we have an indication for which page we're on?
Implementation: it feels a bit hacky to conditionally render the different components in Dashboard.vue, I think we should refactor everything using routing.

I completely agree and was very confused when no routing existed

PR #612, opened 2021-11-30
Update ECE Major Requirements

Aryan77:
### Summary 

This pull request updates the ECE major file to reflect updates to the requirements in Fall 2021. Source: https://www.ece.cornell.edu/ece/programs/undergraduate-programs/majors/program-requirements

### Test Plan 

Refer to the latest engineering handbook.

PR #618, opened 2021-12-07
Remove as2

zachary0kent:
### Summary 
This pull request removes the AS1 and AS2 options from the onboarding college dropdown, replacing these with one "Arts and Sciences" option. Upon onboarding submission, whether the user is actually enrolled in AS1 or AS2 is determined by checking to entrance year and season the user recorded.

### Test Plan 
The new "Arts and Sciences" option:
Screen Shot 2021-12-07 at 5 33 50 PM
Whether you're in AS1 or AS2 is inferred on onboarding submission:
Screen Shot 2021-12-07 at 5 34 27 PM

**NEW** Confirm that an entrance year of Spring 2020 is AS before Fall 2020 (now that season is taken into account).
image

Future STODOs
- [ ] Handle the ISST major in the same way 
- [ ] Add a test to confirm this works (maybe Fall 2019, Spring 2020, Fall 2020, Fall 2021) on Cypress


src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -188,7 +191,7 @@ export default defineComponent({
       placeholderText,
       gradYear: this.onboardingData.gradYear,
       entranceYear: this.onboardingData.entranceYear,
-      collegeAcronym: this.onboardingData.college ? this.onboardingData.college : '',
+      collegeAcronym: this.onboardingData.college ?? '',
```
hahnbeelee, line 194: Whoa I didn't know `??` existed. This is cool!

src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -199,16 +202,20 @@ export default defineComponent({
...
+        .filter(college => !college[0].startsWith('AS'))
+        .map(([key, { name }]) => [key, name]);
+      base.push(['AS', getCollegeFullName('AS')]);
+      base.sort((c1, c2) => c1[1].localeCompare(c2[1]));
+      return Object.fromEntries(base);
     },
     majors(): Readonly<Record<string, string>> {
       const majors: Record<string, string> = {};
       const majorJSON = reqsData.major;
+      const acr = this.collegeAcronym.replace('AS', 'AS2');
```
hahnbeelee, line 215: is this arbitrarily choosing `AS2` to replace `AS` so that it passes the if statement?

hahnbeelee (dismissed):
Basic functionality works, but
There's a weird bug where the chosen college + major disappears once I open the modal up again after choosing Arts & Science.
image

src/components/Modals/Onboarding/OnboardingBasic.vue
willespencer: I think so! The major lists for `AS1` and `AS2` should be the same

willespencer (commented):
Reviewed the code and found one small change to make. After fixing that, resolving merge conflicts, and testing, this should be good for someone else to review!

src/components/Modals/Onboarding/OnboardingBasic.vue
willespencer: Confirmed this was the case in the `index.ts` in the data folder.

jerrry1123 (commented):
everything makes sense for me except the part where we're doing a special case for A&S at the end. Shouldn't we be directly changing the JSON instead?

jessfeng (commented):
Tested on preview url with entrance year before Fall 2020, Fall 2020, and after Fall 2020 and they all look good! 😄 The bug I noticed during dev meeting is when adding another major and clicking finish on the onboarding, navigating to the new added major's header doesn't display anything below it (steps and screenshots are below). It's fine after refreshing the page (progress bar + requirements show up), so I'm not sure how big of a problem this is. I tried the same steps on prod, and it seems to have expected behavior (progress bar + requirements show up after finishing edit profile w/o having to refresh).
Screen Shot 2022-03-10 at 12 19 58 AM

src/utilities.ts
```diff
@@ -50,6 +50,9 @@ export const yearRange = 6;
 
 export function getCollegeFullName(acronym: string | undefined): string {
   // Return empty string if college is not in requirementJSON
+  if (acronym === 'AS') {
```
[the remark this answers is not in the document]
benjamin-shen: This question is totally fair, and it's due to an inconsistency with the frontend and "backend" (how we represent requirements data).

A&S before and after 2020 have different requirements. They are represented as `AS1` and `AS2` in our requirements data. That is, they are considered separate colleges (because we assume one college = one set of requirements). On the frontend, we want the user to see it as a merged college, i.e. `AS`. So, Zak decided to implement this as a frontend-only change by writing conversions between these two representations. An upside of this is that the backend format can stay the same. A downside of this is that A&S needs to be special-cased in few different places.

An alternative implementation would have been something along the lines of adding variants to the requirements data format. So, the backend would look something like `AS:Version1` and `AS:Version2`, where `AS` refers to a single college and `Version1` and `Version2` refer to different sets of requirements. An upside would be that it would be more extensible -- that is, other colleges can have multiple versions as well (which is great if requirements are changed relatively frequently). A downside would be that every other college would have to support this added complexity (for example, `EN` would need to turn into `EN:Version1` or `EN:Default`).

I personally think a single special case like this is fine. But if there are more special cases to be added in the future, or if we somehow manage to reach the backlog, I think there should be a refactoring effort.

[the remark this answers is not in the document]
noschiff: Thanks for explaining that Ben! If we ever have a chance to refactor this to use the alternate implementation you mentioned, that could allow us to handle more specific cases of requirements changing over time (such as the [Probability and Stats requirement for ECE](https://github.com/cornell-dti/course-plan/pull/612#discussion_r767367196)). It's definitely down the line, but it could be useful for our users and help to reduce some confusion.

willespencer:
> Tested on preview url with entrance year before Fall 2020, Fall 2020, and after Fall 2020 and they all look good! 😄 The bug I noticed during dev meeting is when adding another major and clicking finish on the onboarding, navigating to the new added major's header doesn't display anything below it (steps and screenshots are below). It's fine after refreshing the page (progress bar + requirements show up), so I'm not sure how big of a problem this is. I tried the same steps on prod, and it seems to have expected behavior (progress bar + requirements show up after finishing edit profile w/o having to refresh). Screen Shot 2022-03-10 at 12 19 58 AM

Thank you for the detailed testing plan!! However, I still am not able to reproduce this :(
- @einc could you try to see if you can replicate it?
- @jessfeng did you try this on the [dev webapp](https://cornelldti-courseplan-dev.web.app/)? If not could you see if it fails there, because if so, it is probably unrelated to my changes.

Additionally, I fixed the bug @einc pointed out during the meeting that A&S was still labeled pre-post 2020 in the reqs bar! This PR should be good to go now besides this potential bug.

image

einc (approved):

PR #621, opened 2021-12-18
Two-graph individual requirement progress and warnings

willespencer:
### Summary 

This pull request implements two-graph progress tracking for individual requirements based off both @SamChou19815 's experimental code #599, and warnings on individual requirement slots and courses where conflicts occur based on Yuxuan's designs [here](https://www.figma.com/file/vz9T397rHdewbB2McmUDEr/Yuxuan-Working-File?node-id=2035%3A1560)

- [x] Add warnings to requirement bar and schedule view on courses that have conflicts (by updating `CourseCaution`)
- [x] Color courses in req bar and plan when they have these conflicts
- [x] Refactor code to handle two-graphs and display number of combined safe/dangerous courses fulfilling individual requirement slots in their progress counters (including compound requirements)
- [x] **NEW** Use  `RequirementConflictsEnabled` GK so only those with GK enabled see changes.

Remaining TODOs:

- [x] Add dummy data or a gatekeeper in some way so this can be better tested
- [x] In the next PR, color progress bar yellow and add new warning when requirements within it have dangerous fulfillments (refer to Sam's PR)
- [ ] In a PR afterwards, add the modal to make requirements choices, and link to the `FIX NOW` hyperlink in the warnings.

### Test Plan 

To test the below, I updated the function `buildRequirementFulfillmentGraphFromUserData` in `requirement-graph-builder-from-user-data.ts` locally with a harcoded `doubleCountedCourseUniqueIDSet` of `uniqueIds` corresponding to a few courses in my plan (which can be found on Firestore) as there is no way to add conflicts as of now.

image

1. Confirm requirements with conflicts show the total number of fulfillments, even if dangerous, and courses that conflict have warnings and yellow highlight underneath them.
image
image

2. Confirm courses with conflicts have warnings and yellow highlight in semester plan (and conflict warning comes first if there are multiple).

image
image

3. Confirm fulfillment counting works correctly for compounded requirements and credit requirements. Specifically, that they count dangerous fulfillments and can be completed.
image
image

4. Confirm other warning functionality is unchanged.

5. **NEW** Confirm that disabling the GK turns off all the above changes. 
image
image

### Notes 

This can be safely merged with no changes to the user - ~~as there is no way for users to have any conflicts yet to my knowledge.~~ as there is now a GK

Design questions remaining...
1. If a requirement has conflicts that complete it, should the req show up in "completed requirements" ?
2. If a course has a yellow highlight and is clicked on in the semester plan, should the blue or yellow border show? Right now I give the blue prioritization.


einc:
nitpick, but i think we can prob get rid of the "This class is not matched to any requirement..." warning as we have the FIX NOW cta now!

Screen Shot 2021-12-20 at 11 06 48 PM

For design questions,
1. Let's have it not show up in "completed requirements" for now (next sem we can have design look at it), but i think bc completed reqs are usually uncollapsed it might be hard for the user to tell.
2. blue prioritization sounds good as the course is focused! the warning symbol is still there so should be fine! 

idt I can test this (or if i can lmk and i can try!) but the screenshots look good & very thorough!!!! so excited for this <3

willespencer:
> nitpick, but i think we can prob get rid of the "This class is not matched to any requirement..." warning as we have the FIX NOW cta now!

I can pretty easily fix this by changing `CourseCaution` to combine `noMatchedRequirement` and `hasConflictRequirement` - but doing so and merging this branch would get rid of the warning before we fully implement the new requirements flow if we push to master. Making an item, but think we should merge these warnings and ensure they can be fixed with the new modal in a later PR!

> 1. Let's have it not show up in "completed requirements" for now (next sem we can have design look at it), but i think bc completed reqs are usually uncollapsed it might be hard for the user to tell.

I think this is implemented as you want it (completion is determined by the number of safe reqs fulfilled)! Testing it following my plan does not seem to work, but I think that is just because the modal is not yet implemented.

> 2. blue prioritization sounds good as the course is focused! the warning symbol is still there so should be fine!

Agreed!

src/components/Course/CourseCaution.vue
```diff
@@ -39,27 +45,45 @@
...
-  course: FirestoreSemesterCourse | FirestoreSemesterPlaceholder,
+  course: FirestoreSemesterCourse | FirestoreSemesterPlaceholder | CourseTaken,
   semesterIndex: number
 ): CourseCautions => {
   const {
     safeRequirementFulfillmentGraph,
     derivedCoursesData: { duplicatedCourseCodeSet, courseToSemesterMap },
   } = store.state;
+
+  const uniqueID = isCourseTaken(course) ? course.uniqueId : course.uniqueID;
```
benjamin-shen, line 68: This is pretty confusing, we should standardize it to `uniqueId` or something in a future PR (so all 3 kinds of "courses" have the same key for unique id)

src/global-firestore-data/utils.ts
```diff
@@ -16,3 +16,6 @@ export const incrementUniqueID = (amount = 1): number => {
     .set({ uniqueIncrementer: updatedID });
   return updatedID;
 };
+
+export const isCourseConflict = (uniqueId: string | number): boolean =>
+  store.state.doubleCountedCourseUniqueIDSet.has(uniqueId);
```
benjamin-shen, line 21: I don't think this is specific to `global-firestore-data` (which should be for functions directly accessing or manipulating the firestore data), maybe it should go in `requirement-frontend-utils`?

src/components/Course/CourseCaution.vue
willespencer: Agreed! It took me a long time to figure out why `course.uniqueId` was not working only to discover the capitalization difference. Made an item: https://www.notion.so/courseplan/82bf355c4b7b49f99691eb9f3afb203c?v=5da9d646f3ef424bbceb1ca0e8e00cf8&p=f0868e5ead134bdf8acdd11405d7eb3f

src/global-firestore-data/utils.ts
willespencer: Moving this to `requirement-frontend-utils` leads to a dependency cycle between that file and `store`. The `utils` file it is in now already has the `store` import so leaving it there might work, even though I agree it is not specific to `global-firestore-data`. Do you have any other suggestions?

benjamin-shen: you can put it in `store.ts` (maybe a new file `store-utils.ts` or straight up not have it refactored. I think it's definitely more confusing where it is now

src/components/Course/CourseCaution.vue
```diff
@@ -38,28 +44,45 @@
...
+      hasConflictRequirement,
+    };
+  }
+
   const noMatchedRequirement =
     !isPlaceholderCourse(course) &&
     safeRequirementFulfillmentGraph.getConnectedRequirementsFromCourse({
-      uniqueId: course.uniqueID,
+      uniqueId: uniqueID,
     }).length === 0;
```
benjamin-shen, line 86: I know this is not part of this diff but I just happened to notice this during my AP/IB and constraint violation tasks. I think we should use the dangerous requirement graph here. If the safe requirement graph has no connected requirements from the course, that might mean that all the edges are just in a constraint violation. If the dangerous requirement graph has no connected requirements from the course, that means there is no matched requirement.

willespencer: Pushing this off to a later PR based on you and @einc 's above comments about warnings.

https://www.notion.so/courseplan/Hook-up-FIX-NOW-with-conflict-modal-fix-warning-discrepancy-4922837323ce4476b8a3c4f9caba4b17

benjamin-shen (approved):
Let's merge this before it deviates too much :)) thanks for all the changes @willespencer. This is super exciting!!!

PR #625, opened 2021-12-28
Add exam course ids and conditions to requirements json

benjamin-shen:
### Summary 

This pull request is step 3 in the AP/IB Infra Update #576. It builds off the pseudocode described in #606.

- [x] added exam course ids and conditions to requirements json
- [x] generalized decorator logic in `requirement-json-generator`
- [x] conditionally add the `specializations` property (to remove `"specializations": []` in the json) @zachary0kent 

### Test Plan 

No frontend changes 😿. Check `decorated-requirements.json` to see that the added exam course ids and conditions look reasonable.

### Notes 

We might want to consider using [json-stable-stringify](https://www.npmjs.com/package/json-stable-stringify) in the future if the json is not always 100% stable. We can also specify a particular key order with the package.

### DIFF
--- a/src/requirement-types.d.ts
+++ b/src/requirement-types.d.ts
@@ -9,16 +9,6 @@ type RequirementCommon = {
   readonly allowCourseDoubleCounting?: true;
   /** If this is set to true, then AP/IB credits cannot be applied towards this requirement. */
   readonly disallowTransferCredit?: true;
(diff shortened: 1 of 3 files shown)


SamChou19815 (approved):
LGTM. 
Don't think we need to worry about json stability. Although there is no spec on json key order and it might behave different across browsers, the generator is run in node so we know it's always v8 and it respects the order you add the keys.

PR #626, opened 2021-12-31
Update CS major

einc:
### Summary 

This pull request updates the CS major, specifically

- [x] adds the CHEM 2080 Requirement that was previously missing
- [x] also adds CS 4414 for CS 4410 Requirement

Remaining TODOs:
- [ ] Bug where CHEM 2080 substitute course doesn't fulfill Probability Requirement
I followed the example in #611 to specialize the Chem Requirement for CS majors as we are currently missing a req (CHEM 2080), but I might have of missed something? 

The Probability req for CS majors already has "allowCourseDoubleCounting: true" which I thought should fix this problem? any ideas?

Screen Shot 2021-12-30 at 8 28 36 PM

Screen Shot 2021-12-30 at 8 28 44 PM

### Test Plan 

Test Chem 2080 req and CS 4414 for CS major.


noschiff:
adding a course that fulfills CHEM 2080 and Probability works fine when adding it manually from the "+ Course" button, so I would assume that the requirements are done correctly… it's only when the course is dragged from the sidebar that the bug occurs (also please lmk if I'm commenting on this PR correctly lol)

src/requirements/data/specializations/en.ts
```diff
@@ -12,4 +12,27 @@ const MATH2940: CollegeOrMajorRequirement = {
   slotNames: ['MATH 1910', 'MATH 1920', 'MATH 2940'],
 };
 
-export default MATH2940;
+const CHEM2080: CollegeOrMajorRequirement = {
+  name: 'Chemistry',
+  description:
+    'CHEM 2090 and CHEM 2080 (can also be substituted by CHEM 2150, PHYS 2214, PHYS 2218, BTRY 3080, ECON 3130, MATH 2930, or MATH 4710)',
```
willespencer, line 18: Nit but can there be a period at the end of this description per convention

src/requirements/data/specializations/en.ts
```diff
@@ -12,4 +12,27 @@ const MATH2940: CollegeOrMajorRequirement = {
...
+      'PHYS 2218',
+      'BTRY 3080',
+      'ECON 3130',
+      'MATH 2930',
+      'MATH 4710',
+    ]
+  ),
+  fulfilledBy: 'courses',
+  perSlotMinCount: [1, 1],
+  slotNames: ['CHEM 2090', 'CHEM 2080'],
```
willespencer, line 35: Another nit but could the `CHEM 2080` name be something like `CHEM 2080 (or alternative)`

willespencer (approved):
LGTM! The CS 4414 option works, the CHEM specialization works, and neither change negatively affects older CS students.
(For CHEM, checked that the CHEM 2080 alternatives match the CS website, and tested that they can be used to substitute this req in the CS major, but this second slot is not present in other engineering majors)

Can we make an item to specialize the CHEM requirement across the board? Noticing other majors like CE and EE have alternatives as well.

willespencer:
> adding a course that fulfills CHEM 2080 and Probability works fine when adding it manually from the "+ Course" button, so I would assume that the requirements are done correctly… it's only when the course is dragged from the sidebar that the bug occurs (also please lmk if I'm commenting on this PR correctly lol)

Noticing this same thing! (and yes this is commented correctly haha 😆)

Furthermore, I'm noticing both on this branch and on dev that adding `MATH 4710` from the selectable add modal for `Advisor Approved Electives` leads to the same issue of it not counting for Probability. This is true for all courses overlapping with a college and major (for example, adding `CS 1110` by dragging from the eng `Computing` req or from the selectable req add course dropdown do not let it fulfill the CS `Introductory Programming` req). This does not happen on prod though - is this a result of the requirements re-work and if so, will the new front-end resolve this issue?

benjamin-shen (approved):
The bug is not related to this PR, it seems to be caused by an opt-out choice when dragging from the sidebar. `Semester.vue` where `getAllEligibleRelatedRequirementIds()` is called -- looping in @SamChou19815

src/requirements/data/specializations/en.ts
einc: added these, thanks for catching! 

and yes i'll make an item to specialize CHEM reqs and do it in my next pr :)

PR #627, opened 2021-12-31
Switch to new AP/IB infra

benjamin-shen:
### Summary 

This pull request is step 4 in the AP/IB Infra Update #576.

- [x] switched from old infra to new infra 😄 
- [x] deprecated the `derivedAPIBEquivalentCourseData` implemented in #506 (it is no longer being used)

#### TODO
Now is probably a good time to write some e2e cypress tests -- new years resolution?? 🍾

### Test Plan 

AP/IB credit should not be broken 🙈 
image

Some flows to test:
- add exam that should satisfy some requirements
- add exam that shouldn't satisfy any requirements
- add multiple exams simultaneously
- remove exams
- check edges cases (like total credits requirement, FWS)


willespencer (approved):
Code looks good! Also ran through your test plan / added some of my real test credit and everything works as expected. Additionally noticed that AP credit does not automatically fulfill things like `Advisor-Approved Electives` 😄 

Also love the new years resolution, let me know if you want any help!

benjamin-shen:
> Additionally noticed that AP credit does not automatically fulfill things like Advisor-Approved Electives 😄

I think this issue was actually fixed a while ago (since at least #532 - see staging)
image

But you're right that the double counting issues caused by a one-to-many mapping (eg. if an exam has multiple equivalent courses) will also be fixed with the new infra.

hahnbeelee:
Test plan worked for me:
- [x] add exam that should satisfy some requirements
- [x] add exam that shouldn't satisfy any requirements - Made AP CS = 2 and CS 1110 was no longer satisfied
- [x] add multiple exams simultaneously
- [x] remove exams
- [x] check edges cases - total requirements + FWS worked

src/requirements/requirement-exam-utils.ts
```diff
@@ -1,5 +1,5 @@
 import examData, { ExamFulfillment, ExamFulfillments } from './data/exams/ExamCredit';
-import { NO_EQUIVALENT_COURSES_COURSE_ID, CREDITS_COURSE_ID } from './data/constants';
+import { NO_FULFILLMENTS_COURSE_ID, CREDITS_COURSE_ID } from './data/constants';
 
 /** @deprecated old infra */
```
hahnbeelee, line 4: What's the reasoning for keeping the deprecated code?

hahnbeelee (approved):
Both code + test plan LGTM!

src/requirements/requirement-exam-utils.ts
benjamin-shen: So I can remove it all at once! That's step 5 😈 it will be glorious

PR #629, opened 2022-01-03
Remove hardcoded dropdown start year to fix new year Cypress bug

willespencer:
### Summary 

This PR makes the dropdown year range (+- 6 years from current year) common, and imports that within Cypress tests, to avoid hardcoding a start year when selected in a test file. The tests broke because of the new year - which changed our dropdown start year from 2015 to 2016.

image

image

### Test Plan 

1. Confirm that all cypress tests pass
2. Confirm that the new semester dropdown and onboarding entrance/grad year dropdowns have the new expected range (2016-2028)

### DIFF
--- a/cypress/integration/test.spec.ts
+++ b/cypress/integration/test.spec.ts
@@ -3,9 +3,9 @@
  * Can and should be expanded on in more files to test more specific functionality and ensure future bugs are caught earlier
  */
 
-import { getCurrentYear, getYearRange } from '../../src/utilities';
+import { getCurrentYear, yearRange } from '../../src/utilities';
 
-const startYear = getCurrentYear() - getYearRange();
+const startYear = getCurrentYear() - yearRange;
 
 // Before running tests, starts on landing page, logs in to firebase, then visits the dashboard
 // Log in occurs with TEST_UID of the courseplan testing account using a function from the cypress-firebase package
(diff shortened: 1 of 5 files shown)


src/utilities.ts
```diff
@@ -44,6 +44,11 @@ export function getCurrentYear(): number {
   return new Date().getFullYear();
 }
 
+// the number of years +- the current year to show in new semester and onboarding dropdowns
+export function getYearRange(): number {
+  return 6;
+}
```
benjamin-shen, line 50: What does "the number of years +- the current year" mean?

Also maybe it should be a constant instead of a function for now?

cypress/integration/test.spec.ts
```diff
@@ -3,6 +3,10 @@
  * Can and should be expanded on in more files to test more specific functionality and ensure future bugs are caught earlier
  */
 
+import { getCurrentYear, getYearRange } from '../../src/utilities';
+
+const startYear = getCurrentYear() - getYearRange();
```
benjamin-shen, line 8: If we modify `getYearRange` to be based on the user's onboarding data (i.e. by accessing the vuex store), would that work when called in this cypress file?

src/utilities.ts
willespencer: Will clarify the comment and make it a constant, thank you!

cypress/integration/test.spec.ts
willespencer: I'm not sure - but if it doesn't and we do make that change, I could instead grab the user's entrance/grad year from the onboarding modal.

benjamin-shen (approved):

PR #630, opened 2022-01-06
[easy] Fix HE credits double counting and DEA statistics requirement

benjamin-shen:
### Summary 

This pull request fixes some requirements data.

- [x] HE credits requirement should allow course double counting
- [x] DEA statistics requirement should not accept AP credit

### Test Plan 

See summary.
- ~nothing can be tested until #631 is merged~
- ~DEA statistics can not be tested with the graph until new ap/ib infra is merged, but it can be tested by looking at the requirements sidebar~*

*Note: This is because in the old infra, `disallowTransferCredit` is taken into consideration after the graph builder and in the frontend computation -- specifically, `computeFulfillmentStatistics()`. However, in the new infra, it is taken into consideration before the graph builder during the precomputation (requirements json generation) -- specifically, `decorateRequirementWithExams()`.

### DIFF
--- a/src/requirements/data/colleges/he.ts
+++ b/src/requirements/data/colleges/he.ts
@@ -14,6 +14,7 @@ const humanEcologyRequirements: readonly CollegeOrMajorRequirement[] = [
     ],
     fulfilledBy: 'credits',
     perSlotMinCount: [43],
+    allowCourseDoubleCounting: true,
   },
   {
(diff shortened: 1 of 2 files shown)


einc (approved):
Test plan works! HE credits req allows course double counting (except when dragging from Req Bar but that's beyond this PR) and DEA stats didn't accept / show AP credit@

PR #631, opened 2022-01-06
[easy] Fix AP/IB old infra bug for multiple equivalent courses

benjamin-shen:
### Summary 

This pull request fixes a bug in the old AP/IB infra that breaks exams with more than one equivalent course.

### Test Plan 

Choose college/major: HE/DEA. Say you got a 5 on AP Statistics. On staging there are no fulfillments / graph edges; that should be fixed on this branch.

#### Before (staging):
image

#### After (branch):
image

### Notes 

It's not super important to fix this bug since we are moving to the new AP/IB infra, but it might make comparison testing easier.


hahnbeelee (approved):
It's lit!
image

PR #635, opened 2022-01-11
Add AP/IB exam id tests

benjamin-shen:
### Summary 

This pull request tests that the exam unique IDs are unchanged and unique. Modeled from `requirement-data-id.test.ts`.
- The first test is necessary because we are using them in the database as identifiers representing exams taken, and there could be bugs if we change them without writing a migration script
- The second test is not strictly necessary, since `examData` uses the exam subjects as property names, but I kept it for redundancy.

### Test Plan 

👀 ci

I also tested locally that it caught changes to exam subjects.

### DIFF
--- a/src/requirements/__test__/exam-data-id.test.ts
+++ b/src/requirements/__test__/exam-data-id.test.ts
@@ -1,0 +1,27 @@
+import { examSubjects } from '../requirement-exam-utils';
+
+const examIds = [
+  ...examSubjects.AP.map(subject => `AP ${subject}`),
+  ...examSubjects.IB.map(subject => `IB ${subject}`),
+].sort((a, b) => a.localeCompare(b));
+
+it('ID of the exams are unchanged.', () => {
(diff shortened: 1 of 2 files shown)


src/requirements/__test__/exam-data-id.test.ts
```diff
@@ -0,0 +1,27 @@
+import { examSubjects } from '../requirement-exam-utils';
+
+const examIds = [
+  ...examSubjects.AP.map(subject => `AP ${subject}`),
+  ...examSubjects.IB.map(subject => `IB ${subject}`),
+].sort((a, b) => a.localeCompare(b));
+
+it('ID of the exams are unchanged.', () => {
+  /**
+   * What should you do when this test fails:
```
willespencer, line 10: Thanks for the helpful comment!

willespencer (approved):
Tests look good and pass the CI!

image

PR #638, opened 2022-01-16
Fix requirements double counting

benjamin-shen:
### Summary 

This pull request fixes the bug in #626 and removes most instances of improper `.allowDoubleCounting` usage. It also cleans up some other requirements bugs.

Depends on #633 

### Test Plan 
In comments.

### DIFF
--- a/src/components/Modals/NewCourse/NewCourseModal.vue
+++ b/src/components/Modals/NewCourse/NewCourseModal.vue
@@ -32,9 +32,7 @@
         :key="courseSelectorKey"
         :editMode="editMode"
         :selectedRequirementID="selectedRequirementID"
-        :requirementsThatAllowDoubleCounting="
-          requirementsThatAllowDoubleCounting
-        "
+        :requirementsThatAllowDoubleCounting="requirementsThatAllowDoubleCounting"
         :relatedRequirements="relatedRequirements"
         :potentialRequirements="selfCheckRequirements"
         @on-selected-change="onSelectedChange"
@@ -61,10 +59,8 @@ export default defineComponent({
   components: { CourseSelector, TeleportModal, SelectedRequirementEditor },
   emits: {
     'close-course-modal': () => true,
-    'add-course': (
-      course: CornellCourseRosterCourse,
-      choice: FirestoreCourseOptInOptOutChoices
-    ) => typeof course === 'object' && typeof choice === 'object',
+    'add-course': (course: CornellCourseRosterCourse, choice: FirestoreCourseOptInOptOutChoices) =>
+      typeof course === 'object' && typeof choice === 'object',
   },
   data() {
     return {
@@ -97,18 +93,17 @@ export default defineComponent({
       this.$emit('close-course-modal');
     },
     getReqsRelatedToCourse(selectedCourse: CornellCourseRosterCourse) {
-      const { directlyRelatedRequirements, selfCheckRequirements } =
-        getRelatedUnfulfilledRequirements(
-          selectedCourse,
-          store.state.groupedRequirementFulfillmentReport,
-          store.state.toggleableRequirementChoices,
-          store.state.overriddenFulfillmentChoices
-        );
+      const {
+        directlyRelatedRequirements,
+        selfCheckRequirements,
+      } = getRelatedUnfulfilledRequirements(
+        selectedCourse,
+        store.state.groupedRequirementFulfillmentReport,
+        store.state.toggleableRequirementChoices,
+        store.state.overriddenFulfillmentChoices
+      );
 
-      const allReqs = [
-        ...directlyRelatedRequirements,
-        ...selfCheckRequirements,
-      ];
+      const allReqs = [...directlyRelatedRequirements, ...selfCheckRequirements];
       const {
         requirementsThatDoNotAllowDoubleCounting,
         courseToRequirementsInConstraintViolations,
--- a/src/components/Modals/NewCourse/NewCourseModal.vue
+++ b/src/components/Modals/NewCourse/NewCourseModal.vue
@@ -51,6 +51,7 @@ import CourseSelector from '@/components/Modals/NewCourse/CourseSelector.vue';
 import store from '@/store';
 import {
   allowCourseDoubleCountingBetweenRequirements,
+  getRelatedRequirementIdsForCourseOptOut,
   getRelatedUnfulfilledRequirements,
 } from '@/requirements/requirement-frontend-utils';
 import { getConstraintViolationsForSingleCourse } from '@/requirements/requirement-constraints-utils';
@@ -104,10 +105,7 @@ export default defineComponent({
       );
 
       const allReqs = [...directlyRelatedRequirements, ...selfCheckRequirements];
-      const {
-        requirementsThatDoNotAllowDoubleCounting,
-        courseToRequirementsInConstraintViolations,
-      } = getConstraintViolationsForSingleCourse(
+      const { requirementsThatDoNotAllowDoubleCounting } = getConstraintViolationsForSingleCourse(
         { uniqueId: -1 },
         allReqs.map(({ id }) => id),
         (reqA, reqB) =>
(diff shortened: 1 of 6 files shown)


src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -136,9 +133,13 @@ export default defineComponent({
...
-        optOut: this.relatedRequirements
-          .filter(it => it.id !== this.selectedRequirementID)
-          .map(it => it.id),
+        optOut: getRelatedRequirementIdsForCourseOptOut(
+          this.selectedCourse.crseId,
+          this.selectedRequirementID,
+          store.state.groupedRequirementFulfillmentReport,
+          store.state.toggleableRequirementChoices,
+          store.state.userRequirementsMap
+        ),
```
benjamin-shen, line 141: This fixes the following issue:
College: ENG
Major: CS
Minor: Human Development

Add HD 1150 as a course through the add course modal / semester. Due to the limitation of the add modal, the user cannot choose more than one non-double countable requirement. However, after adding the course, the dangerous graph should have all edges not in the constraint violation (eg. if you choose Liberal Studies, both minor requirements should be connected in the dangerous graph).

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -170,10 +175,12 @@ export default defineComponent({
...
           // the course to this specific requirement.
-          optOut: getAllEligibleRelatedRequirementIds(
+          optOut: getRelatedRequirementIdsForCourseOptOut(
             newCourse.crseId,
+            this.subReqId,
             store.state.groupedRequirementFulfillmentReport,
-            store.state.toggleableRequirementChoices
+            store.state.toggleableRequirementChoices,
+            store.state.userRequirementsMap
           ),
```
benjamin-shen, line 184: This fixes the bug in #626.
College: ENG
Major: CS

Drag MATH 4710 from the CHEM 2080 requirement to your schedule. The CS probability requirement should also be connected.

src/components/Semester/Semester.vue
```diff
@@ -280,11 +280,13 @@ export default defineComponent({
...
-            const optOut = getAllEligibleRelatedRequirementIds(
+            const optOut = getRelatedRequirementIdsForCourseOptOut(
               crseId,
+              requirementID,
               store.state.groupedRequirementFulfillmentReport,
-              store.state.toggleableRequirementChoices
-            ).filter(it => it !== requirementID);
+              store.state.toggleableRequirementChoices,
+              store.state.userRequirementsMap
+            );
```
benjamin-shen, line 289: This refers to dragging a course card from the requirements bar to the user's semesters. This should still work properly (edge cases need to be tested but I'm not sure what they are).

src/requirements/requirement-frontend-utils.ts
```diff
@@ -524,14 +547,30 @@ export function getRelatedUnfulfilledRequirements(
...
+    allRequirements.map(({ id }) => id),
+    (reqA, reqB) =>
+      allowCourseDoubleCountingBetweenRequirements(
+        userRequirementsMap[reqA],
+        userRequirementsMap[reqB]
+      )
+  );
+  const automaticallyFulfilledRequirements = relatedRequirements.filter(
+    ({ id }) => !requirementsThatDoNotAllowDoubleCounting.has(id)
+  );
```
benjamin-shen, line 573: Everything related to `automaticallyFulfilledRequirements` fixes the following bug on staging:
College: ENG
Major: CS
Grad: None

Adding INFO 1200 to the schedule causes Engineering Communications to show up as automatically fulfilled. However, after adding the course, the requirement is not actually fulfilled. This is because the requirement is self-check, and should be excluded from that list. This bug will be irrelevant after the new add modal is implemented.

src/components/Modals/NewCourse/NewCourseModal.vue
willespencer: Tested this and confirmed what you are saying is true! 

This isn't ideal on the front-end, as it means that when users choose one of the minor options in the constraint instead of liberal studies, they are still getting liberal studies set to the course. However the new modal will make this flow more clear right?

src/components/Modals/NewCourse/NewCourseModal.vue
```diff
@@ -89,38 +93,31 @@ export default defineComponent({
...
-          requirementsThatAllowDoubleCounting.push(it.name);
-        } else {
-          selfCheckRequirementsThatDoesNotAllowDoubleCounting.push(it);
-        }
-      });
-
-      this.requirementsThatAllowDoubleCounting = requirementsThatAllowDoubleCounting;
-      this.relatedRequirements = relatedRequirements;
-      this.selfCheckRequirements = selfCheckRequirementsThatDoesNotAllowDoubleCounting;
-      if (relatedRequirements.length > 0) {
```
willespencer, line 122: Happy to see all of this confusing code deprecated 😄

src/components/Modals/NewCourse/NewCourseModal.vue
willespencer: Same issue when adding `CS 1110` where adding it to Computing also adds it to Introductory Programming

src/requirements/requirement-frontend-utils.ts
```diff
@@ -444,10 +445,12 @@ export function computeFulfillmentCoursesAndStatistics(
   };
 }
 
-export function getAllEligibleRelatedRequirementIds(
+export function getRelatedRequirementIdsForCourseOptOut(
```
willespencer, line 448: Could you add a comment here? Code looks great though

src/requirements/requirement-frontend-utils.ts
```diff
@@ -501,9 +527,6 @@ export function getRelatedUnfulfilledRequirements(
         toggleableRequirementChoices
       );
       // potential self-check requirements
-      if (requirementSpec == null && !subRequirement.allowCourseDoubleCounting) {
```
willespencer, line 504: Why do we not need the second check here anymore?

src/requirements/requirement-frontend-utils.ts
willespencer: Confirmed this is now fixed!

src/components/Semester/Semester.vue
willespencer: Confirmed courses can drag to semesters (and fulfill multiple reqs at once like CS 1110).

src/components/Requirements/IncompleteSelfCheck.vue
```diff
@@ -101,6 +101,8 @@ export default defineComponent({
             return;
           }
 
+          /* TODO @bshen fix .allowCourseDoubleCounting flag
+             we should allow the constraint violation to be broken, i.e. don't return early */
```
willespencer, line 105: What does this refer to?

src/components/Requirements/IncompleteSelfCheck.vue
willespencer: This works!

willespencer (approved):
Changes look good and everything works as expected - outside of just one remaining question!

src/components/Modals/NewCourse/NewCourseModal.vue
benjamin-shen: > This isn't ideal on the front-end, as it means that when users choose one of the minor options in the constraint instead of liberal studies, they are still getting liberal studies set to the course.

I think this is similar to what would happen in the new add flow. There are no conflicts, so we might as well add it. I believe if you choose a non-liberal studies and non-minor requirement, liberal studies would be no longer connected.

src/components/Requirements/IncompleteSelfCheck.vue
benjamin-shen: The usage of `.allowCourseDoubleCounting` in the following lines.

```
// There are 2 ways we can add the course to the requirement without double counting violations:
// 1. This requirement allows double counting.
// 2. All the already matched requirements allow double counting.
// If both don't hold, we cannot add.
```

We should change this to allow adding regardless of whether or not the requirement allows double counting, and resolve the constraint violations later (basically remove the early return). This fix should wait until after most of the frontend is implemented.

src/requirements/requirement-frontend-utils.ts
benjamin-shen: The function is only called once in `NewCourseModal.vue`, to generate the automatically fulfilled requirements and the non-automatically fulfilled requirements dropdown. In the past, the if-expression was used to filter out the requirements that allowed double counting which we assumed would always be automatically fulfilled. However, with our reframing of the double counting flag (and in this new implementation), it is filtered later in a similar way.

src/components/Requirements/IncompleteSelfCheck.vue
willespencer: Can you make an item for this haha

src/components/Modals/NewCourse/NewCourseModal.vue
benjamin-shen: I modified the function so that it opted out of all `requirementsThatDoNotAllowDoubleCounting`. So, there should be no constraint violations resulting from the minor requirements.

It still automatically selects the liberal studies requirement if it doesn't cause a constraint violation, which is a slight disconnect from the UI. I don't think this can be fixed until the new add modal.

PR #639, opened 2022-01-16
Implement course id conditions

benjamin-shen:
### Summary 

This pull request finishes implementing AP/IB conditions (more generally, course id conditions).

- [x] modified `getMatchedRequirementFulfillmentSpecification()` to filter courses that don't satisfy the requirement conditions
- [x] refactored graph builder to use eligible courses from the spec computation

### Test Plan 

Before:
image

After:
image

AP/IB conditions should work now.
Test colleges condition: Human Ecology - Human Development major / AP Mathematics BC (Engineering vs Non-Engineering)
Test excluded major condition: AP Statistics / Biological Sciences major.

### DIFF
--- a/src/requirements/__test__/requirement-frontend-utils.test.ts
+++ b/src/requirements/__test__/requirement-frontend-utils.test.ts
@@ -6,8 +6,17 @@ const mockRequirementCommon = {
   id: 'MOCK_ID',
   source: '',
   sourceType: 'College',
-  sourceSpecificName: 'Engineering',
+  sourceSpecificName: 'EN',
 } as const;
+
+const mockOnboardingData: AppOnboardingData = {
+  gradYear: '',
+  entranceYear: '',
+  major: [],
+  minor: [],
+  exam: [],
+  tookSwim: 'no',
+};
 
 const getMockCourseTaken = (courseId: number, credits = 0): CourseTaken => ({
   courseId,
@@ -30,7 +39,7 @@ it('computeFulfillmentCoursesAndStatistics self-check test', () => {
 
   // A simple test on self-check.
   expect(
-    computeFulfillmentCoursesAndStatistics(mockRequirement, [], {}, {})
+    computeFulfillmentCoursesAndStatistics(mockRequirement, [], mockOnboardingData, {}, {})
   ).toEqual<StatisticsResultType>({
     courses: [],
     fulfilledBy: 'self-check',
@@ -52,6 +61,7 @@ it('computeFulfillmentCoursesAndStatistics credit test', () => {
     computeFulfillmentCoursesAndStatistics(
       mockRequirement,
       [getMockCourseTaken(1234, 4), getMockCourseTaken(5678, 4)],
+      mockOnboardingData,
       {},
       {}
     )
@@ -67,6 +77,7 @@ it('computeFulfillmentCoursesAndStatistics credit test', () => {
     computeFulfillmentCoursesAndStatistics(
       mockRequirement,
       [getMockCourseTaken(1234, 4), getMockCourseTaken(5678, 4), getMockCourseTaken(9101112, 4)],
+      mockOnboardingData,
       {},
       {}
     )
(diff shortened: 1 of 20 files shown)


src/requirements/requirement-frontend-utils.ts
```diff
@@ -247,15 +251,40 @@ type MatchedRequirementFulfillmentSpecification =
...
+  ) =>
+    coursesList.map(courses =>
+      courses.filter(courseId => {
+        // no requirement conditions
+        if (!(conditions && courseId in conditions)) return true;
+        const { colleges, majorsExcluded } = conditions[courseId];
+        // user college is in colleges list
+        if (college && !colleges.includes(college)) return false;
+        // requirement is in colleges list
+        if (sourceType === 'College' && colleges.includes(sourceSpecificName)) return true;
```
willespencer, line 277: Confused about these two different if statements. The second one makes sense to me as it is structured the same as the `majorsExcluded` check - but what purpose does the first one serve in checking `!colleges.includes(college)`? Is the second check not sufficient?

src/requirements/requirement-frontend-utils.ts
```diff
@@ -247,15 +251,40 @@ type MatchedRequirementFulfillmentSpecification =
...
+    conditions: RequirementCourseConditions | undefined
+  ) =>
+    coursesList.map(courses =>
+      courses.filter(courseId => {
+        // no requirement conditions
+        if (!(conditions && courseId in conditions)) return true;
+        const { colleges, majorsExcluded } = conditions[courseId];
+        // user college is in colleges list
+        if (college && !colleges.includes(college)) return false;
+        // requirement is in colleges list
```
willespencer, line 276: Could you give an example to test where a course would not fulfill a requirement because the college was not in the colleges list? Struggling to come up with one.

willespencer (approved):
Looks good! Test plan works for AP Stat/Bio Sciences

Left a couple questions in the code, should be good to merge after you respond!

src/requirements/__test__/requirement-frontend-utils.test.ts
```diff
@@ -6,9 +6,18 @@ const mockRequirementCommon = {
   id: 'MOCK_ID',
   source: '',
   sourceType: 'College',
-  sourceSpecificName: 'Engineering',
+  sourceSpecificName: 'EN',
 } as const;
 
+const mockOnboardingData: AppOnboardingData = {
```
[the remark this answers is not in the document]
benjamin-shen: I think so

src/requirements/requirement-graph-builder-from-user-data.ts
```diff
@@ -68,24 +69,15 @@ export default function buildRequirementFulfillmentGraphFromUserData(
...
-      let eligibleCoursesList: readonly (readonly number[])[];
-      switch (requirement.fulfilledBy) {
-        case 'self-check':
-          return [];
-        case 'courses':
-        case 'credits':
-          eligibleCoursesList = requirement.courses;
-          break;
-        case 'toggleable':
-          eligibleCoursesList = Object.values(requirement.fulfillmentOptions).flatMap(
```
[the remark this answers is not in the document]
benjamin-shen: This switch is baked into the spec function 😄

src/requirements/requirement-frontend-utils.ts
benjamin-shen: This is anywhere there is an empty list for a particular college (see ExamCredit.ts). For example, AP Math BC (Engineering) should not count unless the user's colleges includes Engineering. Otherwise, the user should use the Non-Engineering AP course.

src/requirements/requirement-frontend-utils.ts
benjamin-shen: Hmm, looking at it now I think it may need some reworking. I might need to tweak this a little bit.

benjamin-shen: Does it make more sense now after my tweaks?

You're right, these two if statements are redundant. I think I only included it as an early return, but it should be the same logic without it. Right now, users can only be in one college. So, there's never a case where their college and college requirements are mismatched.

willespencer (approved):
Logic makes more sense, LGTM!

PR #645, opened 2022-01-27
Assign Course to Placeholder Modal (Front-End)

willespencer:
### Summary 

This pull request implements the front-end for the modal that opens to assign courses to placeholders to continue Templates development. Follow-up from #592.

- [x] Open modal when a placeholder is clicked
- [x] Refactor `NewSelfCheckCourseModal` to handle placeholder assigning as well
- [x] Implement removing placeholder button and styling (with confirmation)
- [x] Add temporary description/learn more values with TODOs  

image

Remaining TODOs:

Complete the requirements side of this modal in a later PR. This should be easier as real placeholder data is populated into a user's Firestore collection. Specifically...

- [ ] Filter out the courses that can be added in the new modal
- [ ] Assign selected course to requirement
- [ ] Replace placeholder in front-end with selected course (and show confirmation)
- [ ] Show corresponding description and learn more link from requirements sidebar

### Test Plan 

Similarly to #592 - placeholder data has to be added to Firestore manually. Afterward...

1. Confirm clicking on a placeholder opens up a modal with the name of the placeholder as its title
2. Confirm the styling of the modal matches the designs
3. Confirm pressing remove on the modal deletes the placeholder from the plan, and refreshing it
4. Confirm courses are properly filtered in a self-check add modal and can be assigned to the requirements and semesters as before.

image

### Notes 

@kkkehui could you take a look and let me know if this looks good? I think I closely followed the designs - but I did keep the placeholder text inside the input box the same as our other add modals for consistency and easier coding. Let me know what you think!

Additionally, This PR can be safely merged as long as the self check add modal remains unchanged, as real users cannot yet have placeholders!

### DIFF
--- a/src/components/Course/Placeholder.vue
+++ b/src/components/Course/Placeholder.vue
@@ -47,18 +47,18 @@ export default defineComponent({
       this.isPlaceholderModalOpen = false;
     },
     assignCourse() {
-      // TODO write the code that assigns a course to a placeholder
+      // TODO @willespencer write the code that assigns a course to a placeholder
     },
     getRequirementID() {
-      // TODO get requirement ID from name and slot
+      // TODO @willespencer get requirement ID from name and slot
       return '0';
     },
     getRequirementDescription() {
-      // TODO get requirement description to show on add modal
-      return 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec vestibulum eu ligula in feugiat. Mauris sed cursus enim, eu feugiat nibh. Vivamus nisi lorem, dictum non aliquet non, blandit non quam. Maecenas vestibulum, ligula et cursus porta, elit nisi facilisis nunc, et faucibus risus arcu a tellus.';
+      // TODO @willespencer get requirement description to show on add modal
+      return 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec vestibulum eu ligula in feugiat. Mauris sed cursus enim, eu feugiat nibh.';
     },
     getRequirementLearnMore() {
-      // TODO get requirement learn more link to show on add modal
+      // TODO @willespencer get requirement learn more link to show on add modal
       return 'https://cornelldti.org';
     },
   },
--- a/src/components/Course/Placeholder.vue
+++ b/src/components/Course/Placeholder.vue
@@ -9,6 +9,7 @@
       :isPlaceholderModal="true"
       @add-course="assignCourse"
       @close-course-modal="closeModal"
+      @delete-placeholder="deletePlaceholder"
     />
     <div class="placeholder-color">
       <img src="@/assets/images/dots/sixDots.svg" alt="" />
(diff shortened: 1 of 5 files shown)


src/components/Course/Placeholder.vue
```diff
@@ -15,17 +26,48 @@
...
   methods: {
     openModal() {
-      // TODO: open modal to assign a course to a placeholder
+      this.isPlaceholderModalOpen = true;
+    },
+    closeModal() {
+      this.isPlaceholderModalOpen = false;
+    },
+    assignCourse() {
+      // TODO @willespencer write the code that assigns a course to a placeholder
```
benjamin-shen, line 55: How were we going to implement assigning courses to placeholders? I thought we modeled placeholders after normal courses because we were going to replace them -- I don't know if this added complexity is worth it.

src/components/Course/Placeholder.vue
```diff
@@ -15,17 +26,48 @@
...
+    },
+    getRequirementID() {
+      // TODO @willespencer get requirement ID from name and slot
```
benjamin-shen, line 58: Just double checking, do placeholders and requirements have a 1:1 mapping? Where is the name and slot stored?

src/components/Course/Placeholder.vue
```diff
@@ -15,17 +26,48 @@
...
+      return '0';
+    },
+    getRequirementDescription() {
+      // TODO @willespencer get requirement description to show on add modal
+      return 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec vestibulum eu ligula in feugiat. Mauris sed cursus enim, eu feugiat nibh.';
+    },
+    getRequirementLearnMore() {
+      // TODO @willespencer get requirement learn more link to show on add modal
+      return 'https://cornelldti.org';
+    },
```
benjamin-shen, line 68: We might want to refactor all of this displayed requirements information into its own component (not necessarily in your PR). I think it's pretty hacky as it is now. I also think it should be calculated at the add placeholder modal level, not in the placeholder component.

benjamin-shen (changes requested):
Thanks @willespencer!

I feel like the add placeholder modal should be a separate component from the add self check modal. I also had some questions and suggestions regarding this implementation.

src/components/Course/Placeholder.vue
willespencer: Requirement name and slot is in the firestore list of courses! Every requirement slot will have at most as many placeholders as courses that are needed to fulfill it, but could have fewer in cases of double counting.

src/components/Course/Placeholder.vue
willespencer: Sorry - assign might be the wrong choice here! Yes the idea was to insert and replace a placeholder with a course in the courses list.

kkkehui:
Hi Will, sorry I just saw this! everything looks good so far! 
The only thing I am not sure about is the outline for the input, it seems to be black in this picture when the input bar is clicked. On the current website, I think it is blue (#0E62C9). If you are reusing the elements from the current add modal, then it should be fine! lmk if there's anything confusing. Thanks!

Screen Shot 2022-02-07 at 7 33 56 PM
Screen Shot 2022-02-07 at 7 36 31 PM

PR #647, opened 2022-02-07
Add entrance/grad seasons to onboarding

willespencer:
### Summary 

This pull request adds entrance and grad season inputs to the Onboarding modal. For new users, they default to Fall & Spring, and for existing users, they will not be able to re-submit onboarding until they are filled out.

This solves the issue in #618 where students that enter in Spring 2020 would be given the post Fall 2020 A&S reqs since we don't check semester season, just year. This also now implements @vaishnavi17 's code in #529 to base starter semesters off of entrance and grad seasons instead of just years

- [x] Added season dropdowns in onboarding with emojis. Initial choice is dependent on if user is new, see above.
- [x] Updated the review page to show the dropdown info
- [x] Updated error check code to block users from submitting unless seasons selected
- [x] Update Cypress tests to select semester seasons

### Test Plan 

1. Test that for existing users, opening onboarding has 'Select one' placeholder text in the new dropdowns, and cannot be submitted until both are filled out. 
image

2. Test editing the fields and re-opening modal has season selections saved, and review page is accurate
image

3. Delete `user-onboarding-data` document for myself and confirm the the new flow, and Fall and Spring respectively selected by default
image

4. Confirm #529 works - specifically that a user's starter semesters populate based on their starting and graduating season now (delete `user-onboarding-data` and `user-semesters)
 
image
(when selecting Spring 2018 and Fall 2022 as entrance/grad)

### Notes 

#618 can be modified after this is merged to use a user's entrance semester!

Some future suggestions based on this PR to make items for (are any of these a priority?)
1. Update TrackUsers analytics to change grad/entrance year counters to grad/entrance sem counters to we know how many spring admits etc. there are
2. Update Vaish's code to handle entrance or grad semesters besides Fall and Winter - selecting Spring or Winter does not populate the correct semesters

### DIFF
--- a/cypress/integration/hidden-features.spec.ts
+++ b/cypress/integration/hidden-features.spec.ts
@@ -35,8 +35,8 @@ it('Onboard a new user with all required fields', () => {
   cy.get('[data-cyId=onboarding]').clickOutside();
   cy.get('[data-cyId=onboarding]').should('be.visible');
 
-  // set Graduation year to 2018
-  cy.get('[data-cyId=onboarding-dropdown]').eq(0).click();
+  // set Entrance semester to 2018 (default Fall)
+  cy.get('[data-cyId=onboarding-dropdown]').eq(1).click();
   cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
     cy.wrap($el)
       .invoke('text')
@@ -49,8 +49,21 @@ it('Onboard a new user with all required fields', () => {
   cy.get('[data-cyId=onboarding-nextButton]').should('be.disabled');
   cy.get('[data-cyId=onboarding-error]').scrollIntoView().should('be.visible');
 
-  // set Graduation year to 2022
-  cy.get('[data-cyId=onboarding-dropdown]').eq(1).click();
+  // set Graduation semester to Summer 2022
+  cy.get('[data-cyId=onboarding-dropdown]').eq(2).click();
+  cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
+    cy.wrap($el)
+      .invoke('text')
+      .then(text => {
+        if (text.includes('Summer')) {
+          cy.wrap($el).click();
+        }
+      });
+  });
+  cy.get('[data-cyId=onboarding-nextButton]').should('be.disabled');
+  cy.get('[data-cyId=onboarding-error]').scrollIntoView().should('be.visible');
+
+  cy.get('[data-cyId=onboarding-dropdown]').eq(3).click();
   cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
     cy.wrap($el)
       .invoke('text')
@@ -64,7 +77,7 @@ it('Onboard a new user with all required fields', () => {
   cy.get('[data-cyId=onboarding-error]').scrollIntoView().should('be.visible');
 
   // set to Engineering college
-  cy.get('[data-cyId=onboarding-dropdown]').eq(2).click();
+  cy.get('[data-cyId=onboarding-dropdown]').eq(4).click();
   cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
     cy.wrap($el)
       .invoke('text')
@@ -80,9 +93,11 @@ it('Onboard a new user with all required fields', () => {
   cy.get('[data-cyId=onboarding-error]').should('not.exist');
   cy.get('[data-cyId=onboarding-nextButton]').click();
 
-  // confirm 2018, 2022, and engineering are selected on the review screen
+  // confirm Fall 2018, Summer 2022, and engineering are selected on the review screen
   cy.get('[data-cyId=onboarding-entranceYear]').contains('2018');
+  cy.get('[data-cyId=onboarding-entranceSeason]').contains('Fall');
   cy.get('[data-cyId=onboarding-gradYear]').contains('2022');
+  cy.get('[data-cyId=onboarding-gradSeason]').contains('Summer');
   cy.get('[data-cyId=onboarding-college]').contains('Engineering');
   cy.get('[data-cyId=onboarding-finishButton]').click();
 });
--- a/cypress/integration/test.spec.ts
+++ b/cypress/integration/test.spec.ts
@@ -90,12 +90,12 @@ it('Edit a semester (Fall of oldest year -> Spring of second oldest year)', () =
     .contains(`Spring ${startYear + 1}`);
 });
 
-// Test that you can change entrance year, grad year, colleges and majors. A later requirements test is dependent on these choices
+// Test that you can change entrance semester, grad semester, colleges and majors. A later requirements test is dependent on these choices
 it('Switch to engineering college and cs major in class of 2022', () => {
   cy.get('[data-cyId=editProfile]').click();
 
-  // set Graduation year to 2018
-  cy.get('[data-cyId=onboarding-dropdown]').eq(0).click();
+  // set Entrance semester to 2018
+  cy.get('[data-cyId=onboarding-dropdown]').eq(1).click();
   cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
     cy.wrap($el)
       .invoke('text')
@@ -106,8 +106,19 @@ it('Switch to engineering college and cs major in class of 2022', () => {
       });
   });
 
-  // set Graduation year to 2022
-  cy.get('[data-cyId=onboarding-dropdown]').eq(1).click();
+  // set Graduation semester to Summer 2022
+  cy.get('[data-cyId=onboarding-dropdown]').eq(2).click();
+  cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
+    cy.wrap($el)
+      .invoke('text')
+      .then(text => {
+        if (text.includes('Summer')) {
+          cy.wrap($el).click();
+        }
+      });
+  });
+
+  cy.get('[data-cyId=onboarding-dropdown]').eq(3).click();
   cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
     cy.wrap($el)
       .invoke('text')
@@ -119,7 +130,7 @@ it('Switch to engineering college and cs major in class of 2022', () => {
   });
 
   // set to Engineering college
-  cy.get('[data-cyId=onboarding-dropdown]').eq(2).click();
+  cy.get('[data-cyId=onboarding-dropdown]').eq(4).click();
   cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
     cy.wrap($el)
       .invoke('text')
@@ -131,7 +142,7 @@ it('Switch to engineering college and cs major in class of 2022', () => {
   });
 
   // set to CS major
-  cy.get('[data-cyId=onboarding-dropdown]').eq(3).click();
+  cy.get('[data-cyId=onboarding-dropdown]').eq(5).click();
   cy.get('[data-cyId=onboarding-dropdownItem]').each($el => {
     cy.wrap($el)
       .invoke('text')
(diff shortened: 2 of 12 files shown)


einc (approved):
can we enforce that graduation year >= entrance year? [this can also be done in the next pr]

put 2017 as my entrance year and 2016 as my grad year (which technically shouldnt be allowed), but it still works as you would expect lol  
Screen Shot 2022-02-16 at 7 11 53 PM
Screen Shot 2022-02-16 at 7 11 30 PM

noschiff:
> can we enforce that graduation year >= entrance year? [this can also be done in the next pr]

@einc yes I'm doing that this week

src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -27,6 +27,24 @@
           >
           <input class="onboarding-input" v-model="lastName" @input="updateBasic()" />
         </div>
+        <div
+          class="onboarding-inputWrapper onboarding-inputWrapper--name onboarding-inputWrapper--description"
+        >
+          <label class="onboarding-label"
+            ><span class="onboarding-subHeader--font"
+              >Entrance Season<span class="onboarding-required-star">*</span>
```
noschiff, line 35: small: personally, I think "semester" makes more sense than season, since we usually refer to the "Fall Semester"

src/components/Modals/Onboarding/Onboarding.vue
```diff
@@ -185,11 +188,11 @@ export default defineComponent({
...
       }
-      if (this.onboarding.gradYear === '') {
-        messages.push('a graduation year');
+      if (this.onboarding.gradYear === '' || !this.onboarding.gradSem) {
+        messages.push('a graduation date');
       }
-      if (this.onboarding.entranceYear === '') {
-        messages.push('an entrance year');
+      if (this.onboarding.entranceYear === '' || !this.onboarding.entranceSem) {
+        messages.push('an entrance date');
```
noschiff, line 195: small: can we flip the order of this so the order of items in the error message matches the order of items in the modal?
i.e. display "Please select an entrance date and a graduation date."

src/components/Modals/Onboarding/OnboardingReview.vue
```diff
@@ -36,6 +36,16 @@
               ><span> {{ userName.lastName }}</span></label
             >
           </div>
+          <div
+            class="onboarding-inputWrapper onboarding-inputWrapper--name onboarding-inputWrapper--description"
+          >
+            <label class="onboarding-label"
+              ><span>Entrance Season<span class="onboarding-required-star">*</span></span></label
```
noschiff, line 43: unrelated to this PR: why bother displaying the * in OnboardingReview? since the review page does not allow you to change data directly, it's sort of irrelevant whether the data you put in was required. also, you need to fill out all required fields before you reach this page anyways.

src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -43,6 +61,23 @@
             />
           </div>
         </div>
+        <div class="onboarding-inputWrapper onboarding-inputWrapper--name"></div>
+        <div class="onboarding-inputWrapper onboarding-inputWrapper--name">
```
noschiff, line 65: most people shouldn't run into this but the wrapping for the drop-downs is a bit weird on a smaller screen. can we just have the rightmost items disappear if the screen is too small?
image

src/components/Modals/Onboarding/OnboardingReview.vue
```diff
@@ -213,6 +234,12 @@ export default defineComponent({
         ? this.onboardingData.entranceYear
         : placeholderText;
     },
+    gradSemText(): string {
+      return this.onboardingData.gradSem ?? placeholderText;
```
noschiff, line 238: couldn't test the placeholder text because I'm not sure if it's even possible to reach review without having a season

noschiff (changes requested):
Looks great and works as expected! I left some comments addressing some small code style and design choices as well as a bug with the wrapping. You can decide if the changes are worth implementing, but they hopefully shouldn't take too long (except for maybe the wrapping issue) if you agree with them.

noschiff:
also, what would you think about saving data to Firestore when the user presses "next" rather than waiting for them to get to "finish"?

src/components/Modals/Onboarding/OnboardingBasic.vue
noschiff: there are some inconsistencies with "season" vs "semester" throughout the code… I've been thinking and I actually do like using season in the code to distinguish it from the "year" component, but for the user I think "semester" might be clearer.

noschiff: i'm working with the drop-downs and I keep getting mixed up lol

src/user-data.d.ts
```diff
@@ -175,9 +177,9 @@ interface CornellCourseRosterCourseFullDetail extends CornellCourseRosterCourse
 // college and grad are optional fields: grad can be undefined if the user hasn't selected a grad program, and college can be undefined if the user has only selected a grad program.
```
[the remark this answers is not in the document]
willespencer: Yes! This is referring to user selecting an undergrad college or grad program (like MPS) which is still optional, as users need to select one of the two but not necessarily one of each.

src/components/Modals/Onboarding/Onboarding.vue
```diff
@@ -185,11 +188,11 @@ export default defineComponent({
       if (this.name.lastName === '') {
         messages.push('a last name');
       }
-      if (this.onboarding.gradYear === '') {
-        messages.push('a graduation year');
+      if (this.onboarding.gradYear === '' || !this.onboarding.gradSem) {
+        messages.push('a graduation date');
```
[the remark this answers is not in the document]
willespencer: I wrote it this way intentionally because it will lead to a shorter error message than if we list ("entrance season, entrance year, graduation season, graduation year, and ...") while still remaining relatively clear. I also think the phrase "Entrance Season" or "Graduation Season" aren't especially clear or used by non cp-devs. Let me know if you still think it is unclear and I can change!

src/components/Modals/Onboarding/OnboardingBasic.vue
willespencer: Yeah we used to use "semester" or even "type" in the code but @benjamin-shen switched it to season mostly across the board last semester

willespencer: Not sure which I prefer, but just tested switching it to "Semester" and it makes the text wrap, so it might just be best to leave it regardless
image

src/components/Modals/Onboarding/OnboardingReview.vue
willespencer: Yeah you make a good point! We didn't always force people to fill in required fields before reaching the review page so it might just be leftover from then. Feel free to bring this up to a designer (maybe Yuxuan) and if they agree we can axe it 😄

src/components/Modals/Onboarding/OnboardingBasic.vue
willespencer: I agree it doesn't look great, but it looks pretty bad on my phone already, and I don't think just deleting the right column fixes the issue. Maybe talk to Kehui about any possible design updates to better improve this section? It was an item last sem but we never got around to it.

image

src/components/Modals/Onboarding/OnboardingReview.vue
willespencer: It shouldn't be! This is here just to handle the type as it might be `''`

cypress/integration/hidden-features.spec.ts
```diff
@@ -80,9 +93,11 @@ it('Onboard a new user with all required fields', () => {
   cy.get('[data-cyId=onboarding-error]').should('not.exist');
   cy.get('[data-cyId=onboarding-nextButton]').click();
 
-  // confirm 2018, 2022, and engineering are selected on the review screen
+  // confirm Fall 2018, Summer 2022, and engineering are selected on the review screen
```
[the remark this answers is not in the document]
willespencer: I'm not it's better just because it is a bit complicated to add constants to Cypress, especially across two files.

cypress/integration/test.spec.ts
```diff
@@ -146,9 +157,11 @@ it('Switch to engineering college and cs major in class of 2022', () => {
   cy.get('[data-cyId=onboarding-nextButton]').click();
   cy.get('[data-cyId=onboarding-nextButton]').click();
 
-  // confirm 2018, 2022, engineering, and computer science are selected on the review screen
+  // confirm Fall 2018, Summer 2022, engineering, and computer science are selected on the review screen
```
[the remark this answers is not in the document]
willespencer: Yes it does use the same account, but the idea of having separate files is that we do not assume anything about the data before the file we are in starts running. This test does not necessarily run right after `hidden-features`, and definitely will not if you run it manually without running `hidden-features` first.

willespencer:
Just responded to all of your comments and made a code change!

> also, what would you think about saving data to Firestore when the user presses "next" rather than waiting for them to get to "finish"?

Don't like this for a couple reasons!
1) Seeing the review page and clicking "Finish" makes it much more clear to the user that there data will be saved then pressing "Next" does, especially if they close the modal before finishing.
2) Users might just be messing around with things so we want them to be sure there data is saved because some of it can have more significant ramifications (i.e. removing an undergrad college I think removed AP test data, and after templates are introduced removng a major might get rid of them). This will be less clear if they don't go through onboarding and see updates to those pages before saving.

Also - per Kehui's new design, this will hopefully be irrelevant in the future as existing users will not use the modal anymore.

src/components/Modals/Onboarding/OnboardingBasic.vue
benjamin-shen: I think it should be called season, and semester should mean (season, year)

cypress/integration/hidden-features.spec.ts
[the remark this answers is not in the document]
noschiff: got it, makes sense!

cypress/integration/test.spec.ts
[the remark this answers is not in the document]
noschiff: sounds good!

src/user-data.d.ts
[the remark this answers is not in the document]
noschiff: ah, I thought that only an `?` was an optional, but it seems like it should work the same

src/components/Modals/Onboarding/Onboarding.vue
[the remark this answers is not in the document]
noschiff: that makes sense! in that case, could we just say graduation semester to encompass both the year and season?

src/components/Modals/Onboarding/OnboardingBasic.vue
noschiff: sounds good! thanks guys!

src/components/Modals/Onboarding/OnboardingReview.vue
noschiff: ok!

noschiff:
> Don't like this for a couple reasons!

that makes sense, thanks for addressing that Will!

noschiff (approved):
Thanks for addressing all my comments, Will! I left a small comment about saying "date" vs "semester" in the error message. It's definitely not necessary, so feel free to leave it as is. Otherwise, everything looks great to me!

src/components/Modals/Onboarding/Onboarding.vue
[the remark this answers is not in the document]
willespencer: Let me check with a designer! Will go with what they prefer since I made this up

PR #648, opened 2022-02-09
Remove console error imports

willespencer:
### Summary 

This pull request removes the imports that are failing in the console on CoursePlan.io as a result of Vue being updated to `v3.2.30`. As noted in #496, a similar attempt to remove a broken import, we manage Bootstrap through NPM and thus do not need the import lines that are failing.

image

### Test Plan 

Ensure that the errors do not appear in the console logs of the deploy link.

Additionally, confirm CP works and is styled as expected. Specifically - confirm styling on the landing page, onboarding modal, and mobile views look as expected

### Notes 

Do we want to push to prod after this is merged since the console errors are on the live site?

### Blockers 

This resolves the blocker on all current PRs with the Cypress tests failing because there were console errors.

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,13 +10,6 @@
     <link rel="icon" href="favicon.ico" />
     <title>CoursePlan</title>
 
-    <!-- Load required Bootstrap and BootstrapVue CSS -->
-    <link
-      type="text/css"
-      rel="stylesheet"
-      href="//unpkg.com/bootstrap-vue@latest/dist/bootstrap-vue.min.css"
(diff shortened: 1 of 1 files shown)


einc (approved):
Thanks for getting to this so fast! 

Tested and no errors appear in console log. I don’t see any differences on the landing page, onboarding, or mobile either. 

We haven’t merged any PRs since Jan release, so pushing to prod (if we want) should be easy as it’s just this PR! @jessfeng thoughts?

PR #658, opened 2022-02-22
[easy] Use .gitignore as prettier ignore path

benjamin-shen:
### Summary 

I didn't like that `npm run format` was taking forever (it was formatting the `dist/` folder, despite it being in the `.gitignore`). Since we are not using `.prettierignore` for anything (other than trying to ignore everything in `.gitignore` file, which doesn't work), I updated the format script as per [this StackOverflow post](https://stackoverflow.com/a/66409885). This should not affect the ci check.

### Test Plan 

`npm run format` should ignore all files in `.gitignore`.

### Notes 

Mentioned in the same StackOverflow post: right now, there is no good way to ignore both the contents of `.gitignore` and the contents of `.prettierignore`.

`.prettierignore` must not be empty, or else there are issues with `actions/checkout@master` in the Github checks.


noschiff:
I'm having trouble reproducing the issue. With `"format": "prettier --write \"**/*.{js,ts,css,scss,html,vue}\""` is in the package.json, `npm run format` already ignores the files in .gitignore. When I added src to the .gitignore, it ignored the files in src. I also ran `npm run format` with the files on master and with the updates in the PR and compared the effects, and there was no difference.

benjamin-shen:
@noschiff is right, I was bugging 🐛 🤣

PR #659, opened 2022-02-22
Course Conflict Modal Frontend

willespencer:
### Summary 

This pull request implements the course conflict resolution modal front-end

- [x] implemented X
- [x] fixed Y

Remaining TODOs:
- [ ] style checkboxes to be CP blue
- [ ] fix when the modal pops up with gatekeep
- [ ] count conflict as resolved when only 1 checkmark selected
- [ ] fix all my TODOs
- [ ] implement Z

Depends on #{number of PR}

### Blockers 

### Breaking Changes  

### DIFF
--- a/src/components/Modals/NewCourse/CourseConflictModal.vue
+++ b/src/components/Modals/NewCourse/CourseConflictModal.vue
@@ -1,0 +1,86 @@
+<template>
+  <TeleportModal
+    title="Fix Course Conflict"
+    content-class="content-course"
+    :rightButtonText="rightButtonText"
+    :rightButtonIsDisabled="selectedCourse == null"
+    @modal-closed="closeCurrentModal"
+    @right-button-clicked="addItem"
+  >
+    <!-- TODO @willespencer factor out selected course? -->
+    <div class="courseConflict-text">Selected Course</div>
+    <div class="selected-course" data-cyId="courseConflict-selectedCourse">
+      {{ selectedCourse.code }}:
+      {{ selectedCourse.name }}
+    </div>
+  </TeleportModal>
+</template>
+
+<script lang="ts">
+import { PropType, defineComponent } from 'vue';
+import TeleportModal from '@/components/Modals/TeleportModal.vue';
+
+export default defineComponent({
+  components: { TeleportModal },
+  emits: {
+    'close-course-modal': () => true,
+    'add-course': (course: CornellCourseRosterCourse, choice: FirestoreCourseOptInOptOutChoices) =>
+      typeof course === 'object' && typeof choice === 'object',
+  },
+  props: {
+    selectedCourse: { type: Object as PropType<FirestoreSemesterCourse>, required: true },
+  },
+  computed: {
+    conflictsResolved(): boolean {
+      // TODO @willespencer update method to determine if conflicts are fully resolved
+      return true;
+    },
+    rightButtonText(): string {
+      return this.conflictsResolved ? 'Done' : 'Save';
+    },
+  },
+  methods: {
+    closeCurrentModal() {
+      this.$emit('close-course-modal');
+    },
+    addItem() {
+      this.addCourse();
+    },
+    addCourse() {
+      // TODO @willespencer handle what happens when conflicts are saved or resolved
+      this.closeCurrentModal();
+    },
+  },
+});
+</script>
+
+<style lang="scss">
+@import '@/assets/scss/_variables.scss';
+.courseConflict {
+  &-text {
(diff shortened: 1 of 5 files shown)


einc:
EXCITING EXCITING!!

Testing works well :) 

1. We can use the default checkbox color!
2. If the user unselects all req options for a conflict, I think technically this should fix the conflict as you’re basically not assigning a course to a requirement. However, this probably isn’t the **intent** of most users that click into the Fix Course Conflict modal. Could we perhaps show the message “Warning: No requirements are selected” but still allow the user to save? 

after playing around with it on CP, I think some styling can be changed to make the modal more clear. will follow up in slack dm with you and yux!

einc (commented):
EXCITING EXCITING!!

Testing works well :)

We can use the default checkbox color!
If the user unselects all req options for a conflict, I think technically this should fix the conflict as you’re basically not assigning a course to a requirement. However, this probably isn’t the intent of most users that click into the Fix Course Conflict modal. Could we perhaps show the message “Warning: No requirements are selected” but still allow the user to save?
after playing around with it on CP, I think some styling can be changed to make the modal more clear. will follow up in slack dm with you and yux!

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -0,0 +1,163 @@
...
+    addCourse() {
+      // TODO @willespencer handle what happens when conflicts are saved or resolved
+      this.closeCurrentModal();
+    },
+    handleChangedConflict(selectedReqs: string[], index: number) {
+      this.selectedReqsPerConflict[index - 1] = selectedReqs;
+      this.updateNumUnresolvedConflicts();
+    },
+    // conflicts exist for each list of reqs in selectedReqsPerConflict if number of reqs selected != 1
+    updateNumUnresolvedConflicts() {
```
benjamin-shen, line 99: This function may need to be changed. It's not necessarily the case that a constraint violation means only one requirement can be chosen. For example, if (A,B,C) is in the requirements list because of constraints between A-B and A-C, then unchecking A but keeping B and C would suffice. The steps should be something like

1. generate requirements lists from `getConstraintViolationsForSingleCourse(currentGraph)`
2. user updates edges, we create a new temporary subgraph from those choices
3. check constraint violations with `getConstraintViolationsForSingleCourse(tempSubgraph)`

src/components/Semester/Semester.vue
```diff
@@ -320,6 +331,9 @@ export default defineComponent({
       }
       return `${credits.toString()} credits`;
     },
+    handleRequirementConflicts(): boolean {
+      return featureFlagCheckers.isRequirementConflictsEnabled();
+    },
```
benjamin-shen, line 336: nice!! 😎

src/components/Semester/Semester.vue
```diff
@@ -376,6 +398,18 @@ export default defineComponent({
       const courseCode = `${data.subject} ${data.catalogNbr}`;
       this.openConfirmationModal(`Added ${courseCode} to ${this.season} ${this.year}`);
     },
+    selectCourse(data: CornellCourseRosterCourse) {
+      // TODO @willespencer handle opening conflict modal better: should not happen if no conflicts
```
benjamin-shen, line 402: Here, I would

1. find all the requirements this course would fulfill, using something like `computeGroupedRequirementFulfillmentReports()` (maybe refactoring some of that function)
2. use `getConstraintViolationsForSingleCourse()` to determine if there are any constraint violations for the course

benjamin-shen (commented):
🤩 nice job!

src/components/Modals/NewCourse/CourseConflictModal.vue
benjamin-shen: @einc maybe the wording in the designs should be changed to reflect this, idk

willespencer: Discussed offline with @einc - since we cannot come up with a concrete example right now, we will push this off and thoroughly test after the rest has been implemented!

willespencer:
> 🤩 nice job!
> 
> We may want to keep the requirements state consistent across all the conflicts. For example, unchecking Requirement 1 in the first list should uncheck Requirement 1 in the second list. It would probably be best to switch to maintaining a temporary (sub)graph in the modal.

As discussed offline - we do not need to do this because Requirement 1 in each conflict will always be different unique reqs per the way conflicts are computed. The names just don't reflect this in the dummy data.

benjamin-shen (approved):

PR #660, opened 2022-02-23
Move documentation from Notion/GitHub Issues to codebase

hahnbeelee:
### Summary 
Adds documentation and links it to relevant code files using https://marketplace.visualstudio.com/items?itemName=mintlify.linked
### Test Plan 
Does this provide value?
What level of documentation should be inside the codebase itself? It's true that if it's in the codebase then it's more easily discoverable. But should architecture specs also be in codebases? I feel like maybe it's too much (specifically Will's template PR) especially cuz it's still a WIP.

This was mainly incentivized in an attempt to provide more context to code - the "WHY".

Let me know what you guys think! Feel free to make changes directly!


willespencer (approved):
This is so amazing Hahnbee! Left some comments about a couple suggested more docs & some author fixes haha, but should be good after I confirm I can see the linked docs with the extension.

Will bring up this system at a future dev meeting so people are aware of this new documentation format and see what people think about it! Thanks Mintlify 😄

PR #664, opened 2022-03-08
Cypress: Add accessibility testing suite

willespencer:
### Summary 

This pull request creates a testing suite for accessibility issues on all views of CoursePlan. Specifically, the [cypress-axe](https://www.npmjs.com/package/cypress-axe) package is imported, which can be used to find "57% of WCAG issues automatically" based on this [ruleset](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md)

See [Dev Assignment](https://www.notion.so/courseplan/Dev-Assignments-a40f92912d0248dab11df9a68df4ba5c?p=3569fff1cf2c4fb49d0f76dbd17e4707) for more details.

- [x] Install the `cypress-axe` package, add to Cypress, and inject after visiting a page
- [x] Create a new spec test file `accessibility-spec.ts` 
- [x] Add tests for the main components of CoursePlan (semester view, reqs bar, and navbar)
- [x] Add tests for hidden large components of CoursePlan (bottom bar and onboarding modal pages)
- [x] Add tests for other public-facing pages (landing page, 404 page, and privacy policy)

Remaining TODOs (for a follow-up PR)

- [ ] Resolve the WCAG errors found by the new testing suite
- [ ] Remove the `skipFailures` flag so that errors on these tests fail the CI in the future (after current ones are resolved)

### Test Plan 

1. Confirm that (without the `skipFailures` flag) the new tests correctly identify WCAG errors in the components they are meant to test.

Example: Bottom bar test, only identifies 2 WCAG issues on the bottom bar
image
(Additionally note that 7/8 of the tests fail)

2. Confirm that after the flag is added, the CI check passes (meaning all the tests pass).
image

### Notes 

Instead of using the `skipFailures` flag, I could have instead opted to set an option to only allow certain types of WCAG issues to cause a CI error ("minor", "moderate", "serious", or "critical"). Because setting it to critical still caused 3+ of the new tests to fail, I decided to opt to skip them entirely until a new PR resolves them, at which point the `options` flag could be considered.

### DIFF
--- a/cypress/integration/accessibility-spec.ts
+++ b/cypress/integration/accessibility-spec.ts
@@ -3,29 +3,29 @@
  * Can and should be expanded in the future.
  */
 
-// Before running tests, starts on landing page, logs in to firebase, then visits the dashboard
+// Before running tests, start on landing page, login to firebase, and inject accessibility scripts
 // Log in occurs with TEST_UID of the courseplan testing account using a function from the cypress-firebase package
-before('Visit site logged in', () => {
+before('Visit landing page logged in', () => {
   cy.visit('localhost:8080/login');
   cy.login(Cypress.env('TEST_UID'));
-  cy.visit('localhost:8080');
+
   // eslint-disable-next-line cypress/no-unnecessary-waiting
-  cy.wait(5000); // ensure the page has time to load
+  cy.wait(2000); // ensure the page has time to load
+
+  cy.injectAxe(); // inject the axe-core library to check accessibility
 });
 
-// Confirm that a semester can be added to the plan
-it('Add a semester (Fall of oldest year)', () => {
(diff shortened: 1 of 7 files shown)


benjamin-shen (approved):
lgtm. This is awesome!! 🤩 

Btw, is `cy.wait()` necessary? It seems like it may cause tests to take longer than necessary, and I think there should be more robust methods to achieve what it's trying to do. This [relevant stackoverflow](https://stackoverflow.com/questions/56973019/how-to-wait-until-page-is-fully-loaded) suggests that `.visit()` has a built-in default `timeout` of 60 seconds.

benjamin-shen:
@willespencer hmmm weird. I think it's probably ok for now

PR #665, opened 2022-03-16
Add fashion design

andxu282:
### Summary 

This pull request is to add the Fashion Design Major. 

- [x] Add Fashion Design Major to index.ts and 

### Test Plan 
- attempt to fill out courses in the major
- double check for double counting
- test the bug found below


jerrry1123:
Everything looks pretty good there's just a few bugs with AP courses that I caught. 

image
There's a bug involving AP Statistics where it should not count for either the Social Science or the Statistics distributions but it double counts it right now. 

image
The other thing is that AP courses are counting for the major's additional requirements even though I'm pretty sure they are not supposed to.

src/requirements/data/majors/fsad.ts
```diff
@@ -0,0 +1,233 @@
...
+  'KCM',
+  'MQR',
+  'LA',
+  'CA',
+  'HA',
+];
+
+const fashionDesignNaturalScienceRequirements: readonly string[] = ['PBS', 'BIOLS-AG', 'BIONLS-AG'];
+
+const fashionDesignRequirements: readonly CollegeOrMajorRequirement[] = [
```
jerrry1123, line 18: see the comment i left in the main pr page for more details, but i found a few bugs relating to how AP courses count for these requirements, and there also might be additional bugs of a simliar nature that i did not catch.

jerrry1123 (changes requested):

src/requirements/data/majors/fsad.ts
einc: @benjamin-shen i think you saw this, but the ap classes have bugs 😭 it's not caused by this PR so I'm gonna add this bug to the notion to be tested before we push to prod.

as we talked off-line, maybe we potentially don't let ap classes automatically count for reqs, but limit it to user opt-in until we fix some of the bigger ap bugs??

the opt-in feature is gonna be added by andrew/will soon so at least there's a workaround (soon) for now

src/requirements/data/majors/fsad.ts
```diff
@@ -0,0 +1,219 @@
...
+  {
+    name: 'Social Science',
```
einc, line 20: let's add "disallowTransferCredit: true" here as a temp fix for now bc AP stats is counting for this but it shouldn't

here's of an example of when we use it:
Screen Shot 2022-04-01 at 1 55 07 AM

src/requirements/data/majors/fsad.ts
```diff
@@ -0,0 +1,219 @@
...
+  },
+  {
+    name: 'Statistics',
+    description:
+      'This fulfills the college distribution quantitative and analytical courses requirement. Must be taken at Cornell, AP Statistics is not accepted.',
+    source: 'https://courses.cornell.edu/preview_program.php?catoid=45&poid=23626',
+    checker: includesWithSubRequirements(['PAM 2100', 'AEM 2100', 'STSCI 2100', 'PSYCH 2500']),
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+    slotNames: ['Course'],
```
einc, line 61: also add "disallowTransferCredit: true" here to not let AP stats count

src/requirements/data/majors/fsad.ts
```diff
@@ -0,0 +1,219 @@
...
+            ifCodeMatch(catalogNbr, '6***'))
+        );
+      },
+    ],
+    fulfilledBy: 'courses',
+    perSlotMinCount: [2],
+    slotNames: ['Course'],
+  },
+  {
+    name: 'Fashion Design Additional Requirements',
```
einc, line 204: also add "disallowTransferCredit: true" here to not let AP stats count

src/requirements/data/majors/fsad.ts
```diff
@@ -0,0 +1,219 @@
...
+  },
+  {
+    name: 'Fashion Design Additional Courses',
+    description:
+      'Take any two FSAD courses at the 3000, 4000, or 6000 level. Course work from Fashion Design Core Courses cannot count here. FSAD 4000, 4010, 4020, 4030, 4990 cannot count here.',
+    source: 'https://courses.cornell.edu/preview_program.php?catoid=45&poid=23626',
+    checker: [
+      (course: Course): boolean => {
+        const { subject, catalogNbr } = course;
+        return (
```
einc, line 191: adding to check that "FSAD 4000, 4010, 4020, 4030, 4990 cannot count here

here's a similar example in CS:
Screen Shot 2022-04-01 at 1 54 00 AM
"

einc (changes requested):
overall looks good, just left a few small changes to make and then ready to merge!!

src/requirements/data/majors/fsad.ts
einc: we're temp not automatically applying ap credits (pr: https://github.com/cornell-dti/course-plan/pull/676#pullrequestreview-928913970), so this problem should be temp fixed rn!

einc (approved):

jerrry1123 (approved):

willespencer (approved):

PR #666, opened 2022-03-19
Entrance + Graduation Year Dropdown Revamp and Validation

noschiff:
### Summary 

This PR revamps the entrance year and graduation year dropdowns in Onboarding Basic Information. The dropdowns are based on the current year and the user's selections, ensuring that the user cannot select invalid years. This PR also validates the user's selection for seasons when the entrance and graduation year are the same.

- [x] populate entrance year dropdown based on current year
- [x] populate entrance year dropdown based on selected entrance year
- [x] suggest entrance year based on current year
- [x] suggest graduation year based on entrance year
- [x] clear graduation year selection if a new entrance year is after old grad year
- [x] when no seasons are selected, check that graduation year is at least entrance year
- [x] check that graduation semester is after entrance semester
- [x] introduce constants for dropdown size
- [x] use dynamic default grad year dropdown with no entrance year

### Test Plan 

- Open the preview URL with a new account, or delete your document for `user-onboarding-data` in Firestore.
- Navigate to "Profile".
- Check that the entrance year options are 6 +/- the current year (2022).
- Check that the graduation year options range from the current year (2022) to 12 years in the future (2032).
- Select an entrance year, and verify that the graduation year options now range from entrance year to entrance year + 8.
- Select a graduation year distinct from the entrance year and verify that your selection is considered valid.
- Select a graduation year = entrance year and verify that the seasons are compared correctly (graduation season ≥ entrance season).
- Select an entrance year that is after the current graduation year and ensure that the graduation year dropdown is cleared and that you are told to select a graduation year.

Also:
- Delete the value for `entranceSem` and `gradSem` for your document under `user-onboarding-data` in Firestore to simulate a user who created their account before seasons were added.
- Select entrance year = graduation year and verify that the only error message is "Please select an entrance semester and a graduation semester."

If desired, you can ensure that the entrance years are generated correctly by changing the return value for `getCurrentYear()` in [utilities.ts](src/utilities.ts) to return a different year (not 2022).

Note that there should never be an error message for the graduation semester being before the entrance semester when the seasons are blank because the graduation year should automatically be cleared when an entrance year that is after the graduation year is selected. The only time the user should receive and error for the graduation semester being before the entrance semester is when the years are the same and seasons are both not blank.

### Notes 

- I chose to allow a user to enter and graduate in the same semester (e.g. Fall 2021 and Fall 2021).
- The order of semesters in a year is: Winter, Spring, Summer, Fall.
- Depends on #647


src/components/Modals/Onboarding/Onboarding.vue
```diff
@@ -172,6 +188,24 @@ export default defineComponent({
         (this.onboarding.grad ? getGradFullName(this.onboarding.grad) === '' : false)
       );
     },
+    /**
+     * Display error if the entrance and graduation year are not blank and the graduation semester comes before entrance semester, comparing the season if not blank
+     *
+     * @returns true if graduation semesters comes before entrance semester, false otherwise
+     */
+    isInvalidGraduationSemester(): boolean {
+      const { gradYear } = this.onboarding;
```
andxu282, line 197: Maybe I don't know the syntax for this, but is this just taking out the gradYear and entranceYear fields from onboarding?

andxu282: Otherwise, the function makes sense to me

src/components/Modals/Onboarding/OnboardingBasic.vue
```diff
@@ -254,6 +254,10 @@ export default defineComponent({
     'click-outside': clickOutside,
   },
   computed: {
+    entranceYears: computeEntranceYears,
```
andxu282, line 257: Is there a reason why this line isn't more similar to the gradYears() one? (line 258-260)

andxu282 (commented):
I added a few questions, but overall, the logic of the code makes sense, I'll run through a test of your changes once I get the demo link and approve

src/components/Modals/Onboarding/Onboarding.vue
noschiff: yup! our ESLint is set to prefer object destructuring over `this.onboarding.gradYear`

src/components/Modals/Onboarding/OnboardingBasic.vue
noschiff: `computeEntranceYears ` doesn't take a parameter, so I can just directly reference it in the Vue as `entranceYears`. since `computeGradYears` takes in a parameter, I had difficultly passing in a parameter in the Vue, so I inserted the `this.entranceYear` parameter in the definition of the function. this was the best solution I found, but if you know of a cleaner way to give the parameter for `:availableChoices="gradYears"`, let me know and I'll change it!

andxu282 (approved):

jerrry1123 (approved):

src/utilities.ts
```diff
@@ -44,9 +45,36 @@ export function getCurrentYear(): number {
...
+    const yr = String(entranceYearNum + i);
+    semsDict[yr] = yr;
+  }
+  return semsDict;
+}
+
+export function computeEntranceYears(): Readonly<Record<string, string>> {
+  const semsDict: Record<string, string> = {};
+  const curYear = getCurrentYear();
+  for (let i = -entranceYearRange; i <= entranceYearRange; i += 1) {
```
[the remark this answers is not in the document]
noschiff: Yes, these two functions are primarily used to populate the dropdown with year options, so I need a list of all the options. I put these functions in utilities.ts so I could take advantage of them when checking that the semesters are valid. Sure, I could've just checked that the chosen year is in the dropdown is legal, but I think its safer and easier to do it this way because there's less of a risk that a future change could break the validation by altering the how we fill the dropdown.

PR #668, opened 2022-03-20
Remove Star from Info Fields in Onboarding Review Page

noschiff:
### Summary 

- [x] Removes the green star (*) after required fields on the Onboarding Review page.

Users can only reach the review page if they have filled out all the required information on the previous pages, so the star is not necessary. Furthermore, users cannot edit their information directly from this page, so it's not useful to remind the user that a field is required. I spoke with Kehui about this change, and she agreed that we can remove the star.

### Test Plan 

Navigate to the Onboarding modal by clicking on "Profile" on the left sidebar, and verify that there are no green stars on the Review page. Also, ensure that there are still green stars for the required fields under the Basic Information page and general page structure and formatting has not changed.

#### Screenshots

| | Basic Information | Review Page |
|-| ----------- | ----------- |
| Before | before      | before       |
|After | after | after

### DIFF
--- a/src/components/Modals/Onboarding/OnboardingReview.vue
+++ b/src/components/Modals/Onboarding/OnboardingReview.vue
@@ -15,9 +15,7 @@
         </div>
         <div class="onboarding-selectWrapperRow-review">
           <div class="onboarding-inputWrapper onboarding-inputWrapper--name">
-            <label class="onboarding-label"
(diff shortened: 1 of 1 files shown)


andxu282 (commented):
LGTM

andxu282 (approved):

PR #669, opened 2022-03-20
Populate conflict modal with requirements

willespencer:
### Summary 

This pull request hooks up the conflict modal created in #659 to the requirements side of CoursePlan such that conflicts can now be resolved!

image

- [x] Only open conflicts modal when there are conflicts for that course
- [x] Populate modal with the actual requirements choices (and refactor to use a map)
- [x] Show self checks if a course can apply to them 
- [x] **NEW** Exclude self-checks that are not in conflict with the other non-self-check reqs
- [x] **NEW** Hide self-check warning if there are no self-checks in modal

Remaining TODOs (for the next PR)

- [x] Ensure the choice affects the added course and requirements fulfilled
- [x] Display the confirmation after the conflict modal is closed

### Questions Remaining 

- What should the selectable requirement logic be? Right now self-checks show up if a course can be applied to that self-check at all. This does not handle cases though where it does not cause a conflict - i.e. if you add HD 1150, it can apply to CS self-checks AND one of the minor reqs at the same time). It also means self-check req could show up in multiple conflicts in the same modal, as shown above.
- Should I delete the "major-cs-" or "minor-hd-" before the requirement name? I assume so, but then there will be no indication of major/minor name which could be confusing for broad req names like "Electives."
- Should the warning text "Requirements with icon are..." only show if there are selectable reqs in the conflict? Right now it always shows up.

### Test Plan 
Set major to Human Development (in human ecology), minor to Human Development 
1. Try to add HD 1150 - confirm you get two sets of conflicts, one for the major and one for the minor, as this course can fulfill multiple reqs but only one at a time in each group.
2. **NEW** Confirm the first conflict has all self-check reqs, including college (as only one of the college & major reqs can be selected). Similarly, confirm the minor group has no self-check reqs.

Set major to CS (in engineering), keep Human Development minor.
1. Try to add CS 4152 - confirm there is a conflict between CS 3000+ elective req, practicum req, and 3 selectable reqs.
image
3. Try to add HD 1150 - confirm that there is a conflict between the two minor requirements. **NEW** Confirm there are no selectable reqs and the selectable req warning does not show.
image
4. Try to add CS 1110 - confirm that if a course has no conflicts (but can apply to self-check reqs), the conflict modal is completely skipped and it is just added to your plan.
6. Confirm that if the flag is turned off, the old add modal v2 shows up after a course is selected.

### Notes 

To enable the GK and use this feature, type GK.enableRequirementConflicts() into the console on the branch or hosted link


einc:
SO EXCITING TO SEE THIS FINALLY ON COURSEPLAN 🥳🤑🤯 THANK U WILL!

For questions remaining:

- The “major-cs-“ for req names is fine for now. We’ll remove this before pushing to prod and I think probably update it to use a label? We can ask kehui more about this, but here are some quick explorations (I like the second row)
label explorartion

- Can we only show the warning text “Requirements with icon are” only show when there are selectable reqs if it’s an easy fix?

einc:
Making sure I understand this right. I think it's correct that self-checks show up if a course can be applied to that self-check at all. 

1. For "It also means self-check req could show up in multiple conflicts in the same modal, as shown above.", this case should be pretty rare right? In your example, it showed up because the major and minor were the same (HD) so the requirements have a lot of overlap?  

2. We're failing to address an edge case when a course can apply to 2 self-checks safely, right? 
As you mentioned, HD 1150 being able to safely fulfill CS major Advisor approved AND HD minor reqs would be an example. 

Screen Shot 2022-04-20 at 12 47 18 PM

The reason why we have the "Conflict: Please only choose one requirement" error is because we don't want users to choose 2 non-self check requirements (where the real error is coming from). 

However, I think it makes sense to allow users to choose one non-self check requirement, and as many self-check requirements as they want? So really the warning should be something like "Conflict: Please only choose one [non self check] requirement" I'll work on the wording a bit bc it's confusing, but is this logic right?

einc (approved):
The rest of the test plan works well!

Tried adding CS 4152, HD 1150, CS 1100 as you wrote. Also tried adding courses without conflicts (CS 2800 & CS 6700) and it works as expected.

I think my comments can be addressed in a future PR. We can finalize the plan during dev meeting tn

willespencer:
> * The “major-cs-“ for req names is fine for now. We’ll remove this before pushing to prod and I think probably update it to use a label? We can ask kehui more about this, but here are some quick explorations (I like the second row)
> label explorartion

> * Can we only show the warning text “Requirements with icon are” only show when there are selectable reqs if it’s an easy fix?
Yes! Will do

Want to discuss your other questions offline as I'm getting a bit confused!

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -104,17 +148,55 @@ export default defineComponent({
...
       }
       this.hasUnresolvedConflicts = unresolved;
       this.hasUnselectedConflicts = unselected;
     },
+    // count number of reqs selected for a conflict (i.e. set to true in the map)
+    countNumberReqsSelected(conflict: Map<string, boolean>) {
+      let count = 0;
+      conflict.forEach((value: boolean) => {
+        if (value) count += 1;
+      });
```
benjamin-shen, line 169: nit: `conflict.filter(Boolean).length`

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -52,14 +59,50 @@ export default defineComponent({
...
+    const numSelfChecksPerConflict: number[] = [];
+
+    const selectableReqIds: string[] = [];
+    this.selfCheckRequirements.forEach(singleSelfCheckReq => {
+      selectableReqIds.push(singleSelfCheckReq.id);
+    });
+
+    this.courseConflicts.forEach(singleConflictList => {
+      const conflictReqIds = [];
+      const singleConflictDict = new Map<string, boolean>();
```
benjamin-shen, line 81: nit: Map instead of Dict

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -104,17 +148,55 @@ export default defineComponent({
...
+      const validConflicts = constraintViolations.courseToRequirementsInConstraintViolations.get(
+        this.selectedCourse.uniqueID
+      );
+      let firstConflict: string[] = [];
+      if (validConflicts) {
+        [firstConflict] = [...validConflicts];
+      }
+
+      return firstConflict;
+    },
```
benjamin-shen, line 199: We should check the selectable requirements one by one and see if the conflicts are disjoint (size 1 vs size 2). If so, append the selectable requirement. (instead of relying on the order of conflicts)

benjamin-shen (changes requested):
nice job :))

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -170,28 +173,42 @@ export default defineComponent({
...
+
+        if (!validConflicts || validConflicts.size !== 1) {
+          return;
+        }
+        const [firstConflict] = [...validConflicts];
+
+        // if selectable req is not in conflict with conflictReqIds, it will be missing from the first and only conflict
+        if (validConflicts && firstConflict.includes(selectableReqId)) {
+          selectableReqIdsInConflict.push(selectableReqId);
+        }
```
benjamin-shen, line 208: Honestly I would just refactor this whole thing to a function in `requirements-constraints-utils`. But that's not super important for right now

I still think `const [firstConflict] = [...validConflicts];` is hard to understand. Also, is `if (validConflicts)` necessary if you check for it earlier? Maybe you can do something like `constraintViolations.requirementsThatDoNotAllowDoubleCounting.has(selectableReqId)`

benjamin-shen: Another thought: is it possible for selectable requirements to conflict with each other? So say you have
- requirement A
- requirement B
- selectable requirement C
- selectable requirement D

You would check {A, B, C} and then {A, B, D}, and add C or D individually to the list of requirement ids if there are potential conflicts. But we should add both C _and_ D if there is a conflict like {A, B, C} and {C, D}, right? If so, maybe this function needs a bit more thought

benjamin-shen (approved):
lgtm

PR #674, opened 2022-03-30
Opt in/out Frontend

andxu282:
### Summary 

This PR implements the frontend for the opt in/out modal. This includes two parts:
- [x] Task 1: Add the Settings icon with the Menu with “Replace, Remove” and “Replace” menus in the Req Bar. The Settings icon should be gray, but blue when hovered over (these colors already exist in the code base)
- [x] Task 2: Create the Replace and Remove modals

### Test Plan 

To test the functionality, I will
- Hover over the gear icon to ensure that the gear changes colors
- Click on the gear icon to ensure the modal opens
- Click on the remove button to open the remove modal
- Click on the replace button to op
- - Choose a class not in the user's current schedule and one that is

For a class that is selected, there are three cases:
1. The class is not in the schedule, we need to add it
2. The class is already in the schedule once, so we just close the modal
3. The class is in the schedule more than once, so we open a modal to resolve duplicates

Case with 0 of that class
Screen Shot 2022-09-14 at 9 01 45 PM
Case with 2+ of that class
Screen Shot 2022-09-14 at 9 01 23 PM


jerrry1123 (commented):
Button changes look pretty good, I just don't see the replace/remove modals that you mention in your summary but I'll assume that's the next step.

src/components/Requirements/CompletedSubReqCourse.vue
```diff
@@ -42,19 +49,21 @@ import { PropType, defineComponent } from 'vue';
...
 import { deleteCourseFromSemesters, deleteTransferCredit } from '@/global-firestore-data';
 import { getCurrentSeason, getCurrentYear, clickOutside } from '@/utilities';
 
 const transferCreditColor = 'DA4A4A'; // Arbitrary color for transfer credit
 
 export default defineComponent({
-  components: { ReqCourse, DeleteCourseModal, SlotMenu },
+  components: { ReqCourse, DeleteCourseModal, ReplaceCourseModal, SlotMenu },
   props: {
     slotName: { type: String, required: true },
```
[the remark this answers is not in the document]
willespencer: Thus, I think what you need to do is pass the requirement name from the parent of this component into `CompletedSubReqCourse` , and then pass it from here to `DeleteCourseModal`. If you are not sure which prop in the parent corresponds to the req name, try console logging the props inside of it!

willespencer (commented):
Left some comments based on the question you asked me offline! Hope this helps 😄

zachary0kent (commented):
Could you add some comments/documentation to the code? In smaller PRs, the code is usually self-explanatory from the PR description, but I think some more context is needed here.

src/components/Modals/DeleteCourseModal.vue
```diff
@@ -9,26 +9,31 @@
     @modal-closed="closeCurrentModal"
     :rightButtonIsDisabled="false"
   >
+    <!-- Remove course -->
     <div v-if="isTransferCredit" class="text-width">
-      Are you sure you want to remove "{{ reqName }}" for this requirement? This will delete the
-      selected transfer credit.
+      Are you sure you want to remove <b>reqName</b> from the <b>{{ reqDesc }} Requirement?</b> This
```
zachary0kent, line 14: Did we determine whether these bold tags are having any effect?

src/components/Modals/NewCourse/ReplaceRequirementDuplicateEditor.vue
```diff
@@ -0,0 +1,111 @@
...
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import { PropType, defineComponent } from 'vue';
+import { GTagEvent } from '@/gtag';
+import SelectSemesterDuplicate from '@/components/Modals/SelectSemesterDuplicate.vue';
+
+export type RequirementWithID = { readonly id: string; readonly name: string };
```
zachary0kent, line 20:
```suggestion
export type RequirementWithID = Readonly<{ id: string; name: string }>;
```

src/components/Modals/NewCourse/ReplaceRequirementDuplicateEditor.vue
```diff
@@ -0,0 +1,111 @@
...
+    editMode: { type: Boolean, required: true },
+    selectedRequirementID: { type: String, required: true },
+    semestersTaken: { type: Array as PropType<readonly FirestoreSemester[]>, required: true },
+    automaticallyFulfilledRequirements: {
+      type: Array as PropType<readonly string[]>,
+      required: true,
+    },
+    // self check requirements
+    potentialRequirements: {
+      type: Array as PropType<readonly { readonly id: string; readonly name: string }[]>,
```
zachary0kent, line 42: Why aren't we using the type `RequirementWithID` here?

src/components/Modals/NewCourse/ReplaceRequirementDuplicateEditor.vue
```diff
@@ -0,0 +1,111 @@
...
+    &-edit {
+      width: 83%;
+    }
+  }
+  &-title {
+    font-size: 14px;
+    line-height: 17px;
+    color: $lightPlaceholderGray;
+    margin-bottom: 6px;
+  }
```
zachary0kent, line 109: Let's prefer `px` over `rm`

src/components/Modals/NewCourse/ReplaceRequirementEditor.vue
```diff
@@ -0,0 +1,114 @@
...
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import { PropType, defineComponent } from 'vue';
+import { GTagEvent } from '@/gtag';
+import SelectSemester from '@/components/Modals/SelectSemester.vue';
+
+export type RequirementWithID = { readonly id: string; readonly name: string };
```
zachary0kent, line 23: Duplicated Type? Let's try to minimize duplication.

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,291 @@
...
+      this.selecting = false;
+      this.getSemestersTaken();
+      const count = this.semestersTaken.length;
+      // opens the add modal if the course does not exist
+      if (count === 0) {
+        this.needToAdd = true;
+        // opens the duplicates modal if the course exists 2+ times
+      } else if (count > 1) {
+        this.hasDuplicates = true;
+        // closes the modal if the course exists exactly once
```
zachary0kent, line 155: Can you move these comments into the conditional block they correspond to?

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,291 @@
...
+          .map(it => it.id),
+        arbitraryOptIn: {},
+      });
+      this.closeCurrentModal();
+    },
+    onSelectedChange(selected: string) {
+      this.selectedRequirementID = selected;
+    },
+    backOrCancel() {
+      if (this.leftButtonText === 'Back') {
```
zachary0kent, line 230: I'm not too keen on using the text to display to determine which state we're in; I think it should be the other way around. For example, we could have a `State` enum that could be used to determine the text to display in each button.

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,291 @@
...
+        if (this.editMode) {
+          this.editMode = false;
+        } else {
+          this.selectedCourse = null;
+          this.courseSelectorKey += 1;
```
zachary0kent, line 235: What is this doing here? Can we add a comment to describe this functionality?

src/components/Modals/SelectSemesterDuplicate.vue
```diff
@@ -0,0 +1,305 @@
...
+      },
+    };
+  },
+  computed: {
+    // shows the semester taken or 'Select Semester'
+    semesterPlaceholder(): string {
+      return this.semesterText || 'Select Semester';
+    },
+    semestersTakenList(): string[] {
+      const semestersHash = new Map<string, number>();
```
zachary0kent, line 101: I think the name `semestersHash` is somewhat misleading--it's really a mapping.

src/components/Modals/SelectSemesterDuplicate.vue
```diff
@@ -0,0 +1,305 @@
...
+      const semestersTakenList: string[] = [];
+      for (const semester of this.semestersTaken) {
+        const { year, season } = semester;
+        const semesterString = `${season} ${year}`;
+        const val = semestersHash.get(semesterString);
```
zachary0kent, line 106: What is `val`?

src/components/Modals/SelectSemesterDuplicate.vue
```diff
@@ -0,0 +1,305 @@
...
+  computed: {
+    // shows the semester taken or 'Select Semester'
+    semesterPlaceholder(): string {
+      return this.semesterText || 'Select Semester';
+    },
+    semestersTakenList(): string[] {
+      const semestersHash = new Map<string, number>();
+      const semestersTakenList: string[] = [];
+      for (const semester of this.semestersTaken) {
+        const { year, season } = semester;
```
zachary0kent, line 104:
```suggestion
      for (const { year, season } of this.semestersTaken) {
```

src/components/Modals/SelectSemesterDuplicate.vue
```diff
@@ -0,0 +1,305 @@
...
+        },
+      },
+    };
+  },
+  computed: {
+    // shows the semester taken or 'Select Semester'
+    semesterPlaceholder(): string {
+      return this.semesterText || 'Select Semester';
+    },
+    semestersTakenList(): string[] {
```
zachary0kent, line 100: Can you add some documentation for this function and the motivation for how it's implemented?

src/components/Modals/SelectSemesterDuplicate.vue
```diff
@@ -0,0 +1,305 @@
...
+  methods: {
+    showHideSemesterContent() {
+      const displayOptions = this.displayOptions.semester;
+      const contentShown = displayOptions.shown;
+      displayOptions.shown = !contentShown;
+
+      if (contentShown) {
+        // clicked box when content shown. So then hide content
+        displayOptions.boxBorder = inactiveGray;
+        displayOptions.arrowColor = inactiveGray;
```
zachary0kent, line 135: Is there any case in which these colors are different?

zachary0kent (changes requested):
Wow, great work!! I think a little more documentation is still needed in some areas of the code, and I think we should also make an effort to eliminate duplication between different components. I know trying to factor out common code caused issues before, but that doesn't mean it's a bad thing to pursue.

zachary0kent (commented):
With so many changes, I also think it would be a good idea to add a couple tests for the new user flow.

src/components/Modals/DeleteCourseModal.vue
andxu282: I think they do make it slightly more bold, but not as bold as the designs make it, so I just made the font larger as well.

zachary0kent:
@andxu282 Can you change the base branch to `user-choice`?

src/components/Modals/DeleteCourseModal.vue
benjamin-shen, line 14: Does `reqName` need to be in double braces?

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,294 @@
...
+        @on-escape="closeCurrentModal"
+        @on-select="selectCourse"
+        data-cyId="newCourse-dropdown"
+      />
+      <div class="newCourse-text" v-if="selectedCourse === null">
+        <img
+          class="requirement-checker-warning-icon"
+          src="/src/assets/images/warning.svg"
+          alt="warning icon"
+          data-v-74b2c930=""
```
benjamin-shen, line 36: What is this for?

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,294 @@
...
+    content-class="content-course"
+    :leftButtonText="leftButtonText"
+    :rightButtonText="rightButtonText"
+    :rightButtonIsDisabled="selectedCourse == null"
+    @modal-closed="closeCurrentModal"
+    @left-button-clicked="backOrCancel"
+    @right-button-clicked="addItem"
+  >
+    <div class="newCourse-text">
+      {{ selecting ? 'Search Course Roster' : 'Selected Course' }}
```
benjamin-shen, line 13: Can this be a computed value? Same with the other expressions and magic strings in this file

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,294 @@
...
+        placeholder='"CS 1110", "Multivariable Calculus", etc'
+        :autoFocus="true"
+        @on-escape="closeCurrentModal"
+        @on-select="selectCourse"
+        data-cyId="newCourse-dropdown"
+      />
+      <div class="newCourse-text" v-if="selectedCourse === null">
+        <img
+          class="requirement-checker-warning-icon"
+          src="/src/assets/images/warning.svg"
```
benjamin-shen, line 34: Should this have a `@` prefix instead of `/src?

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,294 @@
...
+          data-v-74b2c930=""
+        />
+        We can't check that this course correctly fulfills the requirement so check carefully before
+        selecting.
+      </div>
+    </div>
+    <!-- show add modal -->
+    <div v-else-if="needToAdd && selectedCourse != null">
+      <replace-requirement-editor
+        :deleteSemYear="0"
```
benjamin-shen, line 45: Why is this 0?

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,294 @@
...
+          store.state.groupedRequirementFulfillmentReport,
+          store.state.toggleableRequirementChoices,
+          store.state.userRequirementsMap
+        ),
+        // Only include the selected requirement from opt-in.
+        acknowledgedCheckerWarningOptIn: this.selfCheckRequirements
+          .filter(it => it.id === this.selectedRequirementID)
+          .map(it => it.id),
+        arbitraryOptIn: {},
+      });
```
benjamin-shen, line 226: Wouldn't this be an arbitrary opt in choice? Can you add a note to update this?

src/components/Modals/ReplaceCourseModal.vue
```diff
@@ -0,0 +1,294 @@
...
+    onSelectedChange(selected: string) {
+      this.selectedRequirementID = selected;
+    },
+    backOrCancel() {
+      if (this.leftButtonText === leftButtonState.Back) {
+        if (this.editMode) {
+          this.editMode = false;
+        } else {
+          this.selectedCourse = null;
+          this.courseSelectorKey += 1; // courseSelectorKey counts the number of times the courseSelector has been open
```
benjamin-shen, line 238: Not sure if this is the best pattern -- why is a key necessary for the component?

src/components/Modals/SelectSemesterDuplicate.vue
```diff
@@ -0,0 +1,307 @@
...
+      }
+    },
+    selectSemester(semester: string) {
+      this.semesterText = semester;
+      const displayOptions = this.displayOptions.semester;
+      displayOptions.shown = false;
+      displayOptions.boxBorder = '#C4C4C4';
+      displayOptions.arrowColor = '#C4C4C4';
+      displayOptions.placeholderColor = '#757575';
```
benjamin-shen, line 159: These color constants should be refactored

src/components/Modals/SelectSemesterReplace.vue
```diff
@@ -0,0 +1,486 @@
...
+        defaultYear = 'Year';
+      }
+      return String(this.yearText || this.year || defaultYear);
+    },
+    // scroll the bottom of the year dropdown to 4 years after the user's entrance year, or the current year if undefined
+    scrollTopToElement(): number {
+      const topYear = store.state.onboardingData.entranceYear
+        ? store.state.onboardingData.entranceYear
+        : '2017';
+      return Number(topYear);
```
benjamin-shen, line 200: Sorry, could you explain this code?

src/feature-flags.ts
```diff
@@ -5,6 +5,7 @@ const GateKeeperTogglers: Record<string, () => void> = {};
 type FeatureFlagName =
   | 'APIBFulfillment'
   | 'Case'
+  | 'OptInOptOut'
```
benjamin-shen, line 8: Can you remove this if we're not using a feature flag?

benjamin-shen (commented):
Wow! This is a massive PR. Thanks for the hard work on this. Just letting you know that big PRs are [discouraged](https://dev.cornelldti.org/docs/guide-code-review) since they're a lot harder to review; I think it would be better to make small PRs and merge them to a non-master branch.

Could you explain the PR a bit more? Maybe in the style of a PR discussion dev assignment? Specifically, I'm confused about how the components will flow.

I had a few other comments from my initial pass.

src/components/Modals/ReplaceCourseModal.vue
andxu282: I think it's to make the warning icon the correct size, otherwise, it's much larger.

src/components/Modals/ReplaceCourseModal.vue
andxu282: Now that I'm looking at it, I don't think it is

src/components/Modals/ReplaceCourseModal.vue
andxu282: Hmm, I understand what arbitrary opt in is, but I'm not exactly sure how to fix this?

benjamin-shen: I would remove the acknowledged opt in and just make a note for the future, we can discuss more in person

src/components/Modals/ReplaceCourseModal.vue
andxu282: Fixed with [c176926](https://github.com/cornell-dti/course-plan/pull/674/commits/c176926f95eef53a23ca4b6f3d5949776b58e0f6)

src/components/Modals/ReplaceCourseModal.vue
andxu282: I think after all the merging, this may be resolved

andxu282:
I added documentation here! Let me know if it's clear enough now
https://docs.google.com/document/d/12boyLpgJP2aizFjNI99Pt6yP5p5IqhAMkC-qGwes1fw/edit?usp=sharing

zachary0kent (approved):
Looks good to me! Thanks for all the hard work!

PR #675, opened 2022-03-31
Added new profile container with the first page of settings

jerrry1123:
### Summary 

Created a new profile container which shows up when Tools are enabled and when you click on the profile button. This new profile container so far only contains the first page to edit the profile, the rest will be added in a future PR. This is the first step towards implementing the profile redesign that Kehui made. 

Remaining TODOs:
- Add the rest of the sections required to edit the profile and the buttons to navigate them
- The templates will also go into this container in the future. 
- Add the confirmation modal when submitting changes

### Test Plan 

- Make sure that when tools are disabled, the page has the default behaviour of opening the onboarding modals
- Make sure that when editing the profile, the changes are only saved when clicking submit and that no changes are saved when exiting prematurely
- Make sure that switching between the navbar pages functions correctly


noschiff:
is this PR ready to be reviewed? right now the TypeScript type check fails, so we won't be able to merge this in. you can run `npm run type-check` to see the error.

jerrry1123:
> is this PR ready to be reviewed? right now the TypeScript type check fails, so we won't be able to merge this in. you can run `npm run type-check` to see the error.

should be ready now. The only test that's failing right now is the cypress test, but afaik we don't have to worry about that yet.

src/components/NavBar.vue
```diff
@@ -35,7 +35,7 @@
             <span>Tools</span>
           </div>
         </div>
-        <div class="navbar-buttonWrapper desktop" @click="editProfile" data-cyId="editProfile">
+        <div class="navbar-buttonWrapper desktop" @click="openProfile" data-cyId="openProfile">
```
willespencer, line 38: `data-cyId` used here and in other places in this file refers to a "cypress ID," which is an attribute that we use to retrieve this element when running our front-end cypress tests. 

I believe the tests are failing because you changed the name here. If you change it back, they should pass, as the new tools page is behind a gatekeep, so your code should not cause a test to fail (until the gatekeep is removed, at which point we can talk about rewriting it 😄)

src/containers/Dashboard.vue
```diff
@@ -231,17 +244,28 @@ export default defineComponent({
...
+      this.isProfileOpen = false;
     },
 
     editProfile() {
       this.isEditingProfile = true;
       this.startOnboarding();
     },
 
+    openProfile() {
+      if (featureFlagCheckers.isToolsEnabled()) {
```
willespencer, line 261: Clean use of the GK! Nice job 😄

src/components/NavBar.vue
```diff
@@ -77,7 +77,7 @@
           <div class="navbar-iconWrapper tools-mobile-icon" />
           <span class="nav-mobile-button-text">Tools</span>
         </button>
-        <button class="nav-mobile-button" data-cyId="navbar-editProfile" @click="editProfile">
+        <button class="nav-mobile-button" data-cyId="navbar-openProfile" @click="openProfile">
```
willespencer, line 80: Is there a reason you changed the @click method to `openProfile` and added the `openProfile` emit? It seems like it has the same functionality, and if so, I think we should either keep the method/emit `editProfile` or refactor its name, instead of having both.

src/requirements/data/minors/ineq.ts
```diff
@@ -0,0 +1,235 @@
+import { CollegeOrMajorRequirement } from '../../types';
```
willespencer, line 1: I think this file got included from your other PR as a mistake. Could you please remove it from this branch?

src/requirements/data/index.ts
```diff
@@ -357,6 +356,11 @@ const json: RequirementsJson = {
       schools: ['HE'],
       requirements: hdMinorRequirements,
     },
+    INEQ: {
```
willespencer, line 359: Similarly - could you please revert the changes in this file? Let me know if you need any help with how to do so.

src/requirements/decorated-requirements.json
```diff
@@ -199661,6 +199661,158 @@
         }
       ]
     },
+    "INEQ": {
```
willespencer, line 199664: Same here

src/requirements/__test__/__snapshots__/requirement-data-id.test.ts.snap
```diff
@@ -362,6 +362,9 @@ Array [
   "Minor-HP-3000/4000+ Level",
   "Minor-HP-Additional Course",
   "Minor-HP-PAM 2350",
+  "Minor-INEQ-Four Elective Courses",
```
willespencer, line 365: And here!

src/containers/Profile.vue
```diff
@@ -0,0 +1,301 @@
+<template>
+  <div class="profileContainer">
+    <div class="profileContainer-header">
+      <div class="profileContainer-title">
+        <span> Edit Profile</span>
+      </div>
+      <div class="profileContainer-subtitle">
+        <span> get your information up-to-date! </span>
```
willespencer, line 8: Is `Get` supposed to be lower case?

src/containers/Profile.vue
```diff
@@ -0,0 +1,301 @@
...
+      </div>
+    </div>
+    <div class="profileContainer-editPage">
+      <onboarding-basic
```
willespencer, line 12: I'm so glad this worked 😄

src/containers/Profile.vue
```diff
@@ -0,0 +1,301 @@
...
+        this.name.firstName === '' ||
+        this.name.lastName === '' ||
+        this.onboarding.gradYear === '' ||
+        !this.onboarding.gradSem ||
+        this.onboarding.entranceYear === '' ||
+        !this.onboarding.entranceSem ||
+        (this.onboarding.college === '' && this.onboarding.grad === '')
+      );
+    },
+    timelineTextImage(): string {
```
willespencer, line 75: I don't think this computed method is being called anywhere in this new container. Could you remove it along with the imported timeline svgs?

willespencer: This comment applies to a lot of things in this file - such as the `FINAL_PAGE` constant, the emits, etc. Could you clean it up a little? You can always add back stuff as needed.

src/containers/Profile.vue
```diff
@@ -0,0 +1,301 @@
...
+  data() {
+    return {
+      currentPage: 1,
+      name: { ...this.userName },
+      onboarding: { ...this.onboardingData },
+    };
+  },
+  computed: {
+    // Display error if a required field is empty
+    isError(): boolean {
```
willespencer, line 64: I'm noticing the error text does not show up if an error occurs. Similarly - users can save their data even if there is an error - resulting in the possibility a user has no college/grad program selected and thus an empty reqs bar. Could you either enforce the error code like onboarding or add it to your todo list for a future PR?

image

src/containers/Profile.vue
```diff
@@ -0,0 +1,301 @@
...
+      if (this.onboarding.grad !== '' && !this.onboarding.college && this.currentPage > 1) {
+        this.currentPage = 1;
+      } else {
+        this.currentPage = this.currentPage - 1 === 0 ? 0 : this.currentPage - 1;
+      }
+    },
+    setPage(page: number) {
+      this.currentPage = page;
+    },
+    canProgress() {
```
willespencer, line 152: This method should be able to be used to determine if data can be saved!

src/containers/Profile.vue
```diff
@@ -0,0 +1,301 @@
...
+        this.cancel();
+      }
+    },
+  },
+});
+</script>
+<style scoped lang="scss">
+@import '@/assets/scss/_variables.scss';
+
+.profileContainer {
```
willespencer, line 225: Make sure to also delete any styling that is not applicable anymore!

src/containers/Profile.vue
```diff
@@ -0,0 +1,301 @@
...
+  background-color: #f6fafc;
+  padding: 3rem 7rem;
+
+  &-title {
+    width: 523px;
+    height: 40px;
+    right: 753px;
+    top: 40px;
+
+    font-family: 'Proxima Nova';
```
willespencer, line 235: This one is not on you - but could you remove this font-family line? We set the font-family globally so it probably doesn't do anything here.

willespencer (changes requested):
This is great start Jerry! The new page already looks good, the old page works when the GK is disabled, and I'm impressed that the data changes save 😄. Could you add your todo list on an item/page on Notion so that the PM/TPM can track your progress and make sure we don't forget anything?

Left some comments for code changes, and a suggestion to get the Cypress tests to pass (which we want to pass before merging). Afterwards, this should be good to go!

I also found an issue specifically with saving a change of college to A&S, but it seems like this is a problem on the old onboarding model on the dev site, so don't worry about this console log for now.
image

src/components/NavBar.vue
jerrry1123: From what i understood changing it would fix the mobile behaviour to be similar to desktop, but i can change i back for now.

src/containers/Profile.vue
jerrry1123: the figma design had it lower case so i kept it as is

src/components/NavBar.vue
```diff
@@ -77,7 +77,7 @@
           <div class="navbar-iconWrapper tools-mobile-icon" />
           <span class="nav-mobile-button-text">Tools</span>
         </button>
-        <button class="nav-mobile-button" data-cyId="navbar-openProfile" @click="openProfile">
+        <button class="nav-mobile-button" data-cyId="navbar-editProfile" @click="editProfile">
```
willespencer, line 80: Sorry, I was a little confused last time I reviewed this but I think I get the jist now. I think we still want `@click="openProfile"` here since we presumably want mobile to also use the new profile page, but calling `editProfile` means it will open the onboarding modal.

When we do that though, it means the `editProfile` method in this file will never be called, the `emit` from this file will never be used, and the emit will not be received in the parent (`Dashboard`). Could all of that be deleted?

src/components/NavBar.vue
```diff
@@ -141,6 +141,10 @@ export default defineComponent({
       this.menuOpen = false;
       this.$emit('editProfile');
```
willespencer, line 142: So this method would not be needed

src/components/NavBar.vue
```diff
@@ -110,7 +110,7 @@ export default defineComponent({
   props: {
     isDisplayingRequirementsMobile: { type: Boolean, required: true },
   },
-  emits: ['openPlan', 'openTools', 'editProfile', 'toggleRequirementsMobile'],
+  emits: ['openPlan', 'openTools', 'editProfile', 'toggleRequirementsMobile', 'openProfile'],
```
willespencer, line 113: Nor the `editProfile` emit here

src/containers/Dashboard.vue
```diff
@@ -18,12 +18,13 @@
           @openPlan="openPlan"
           @openTools="openTools"
           @editProfile="editProfile"
```
willespencer, line 20: And lastly this event reciever that calls `editProfile` can be deleted. Don't remove the `editProfile` method itself though, as your `openProfile` method in this file calls it!

willespencer (changes requested):
Thanks for making all of the changes! Just clarified one more thing to fix, but after that this should be good to go 😄

jessfeng:
Woohoo!! This page is looking pretty cool, thanks for the work Jerry! 😎 Some things I noticed while playing around the preview url and Figma:

- `Edit Profile` title and its subtitle have 19px spacing in between them on the Figma file, so maybe add padding/margins there

Figma:
Screen Shot 2022-05-01 at 4 17 30 PM
Preview:
Screen Shot 2022-05-01 at 4 18 45 PM


- The headers/text for each of the profile sections and their textboxes have a white highlighted background under the text itself, which isn't super visible but since the background color of the page is a super light cool grey, it's still a little noticeable. It was fine for the first-time onboarding popup bc the background for that was white, so we can either remove the highlight or change the profile page's background to be white. 
   - Also peeked 👀 on the Figma and it seems like @yuxuanchenn's tools page in the nav bar has the light cool grey background but @kkkehui's profile page in the nav bar has a white background, so I can ask them if they want to standardize that.
    Screen Shot 2022-05-01 at 4 21 25 PM


- When narrowing the page (I think to 878px or less) to when we see the side hamburger, the courseplan logo header covers the "Edit Profile" title text. 
Screen Shot 2022-05-01 at 5 10 59 PM


- Everything gets kind of squished/overlapping when narrowing the window, so if you could keep it consistent with the current onboarding page, that would be nice but if it's too much of a hassle, we can keep it as it is now and just file a ticket for it later, up to you!

Prod (modal):
Screen Shot 2022-05-01 at 5 15 51 PM

Preview (page):
Screen Shot 2022-05-01 at 5 15 09 PM

jerrry1123:
> Woohoo!! This page is looking pretty cool, thanks for the work Jerry! 😎 Some things I noticed while playing around the preview url and Figma:
> 
>     * `Edit Profile` title and its subtitle have 19px spacing in between them on the Figma file, so maybe add padding/margins there
> 
> 
> Figma: Screen Shot 2022-05-01 at 4 17 30 PM Preview: Screen Shot 2022-05-01 at 4 18 45 PM
> 
>     * The headers/text for each of the profile sections and their textboxes have a white highlighted background under the text itself, which isn't super visible but since the background color of the page is a super light cool grey, it's still a little noticeable. It was fine for the first-time onboarding popup bc the background for that was white, so we can either remove the highlight or change the profile page's background to be white.
>       
>       * Also peeked 👀 on the Figma and it seems like @yuxuanchenn's tools page in the nav bar has the light cool grey background but @kkkehui's profile page in the nav bar has a white background, so I can ask them if they want to standardize that.
>       
>         Screen Shot 2022-05-01 at 4 21 25 PM
> 
>     * When narrowing the page (I think to 878px or less) to when we see the side hamburger, the courseplan logo header covers the "Edit Profile" title text.
> 
> 
> Screen Shot 2022-05-01 at 5 10 59 PM
> 
>     * Everything gets kind of squished/overlapping when narrowing the window, so if you could keep it consistent with the current onboarding page, that would be nice but if it's too much of a hassle, we can keep it as it is now and just file a ticket for it later, up to you!
> 
> 
> Prod (modal): Screen Shot 2022-05-01 at 5 15 51 PM
> 
> Preview (page): Screen Shot 2022-05-01 at 5 15 09 PM

Thanks for the feedback! I'll work on fixing the stuff in the newest PR I have, so you can probably see the fixes in that PR's preview page

jerrry1123:
Thanks for the feedback! I'll work on fixing the stuff in the newest PR I have, so you can probably see the fixes in that PR's preview page

willespencer (approved):
Thanks for making the changes, this looks good to go now!

Please make sure to address all of @jessfeng's comments in the next PR 😄

PR #676, opened 2022-04-01
[easy] Put AP/IB fulfillment behind feature flag

benjamin-shen:
### Summary 

This pull request puts AP/IB fulfillment behind the feature flag `APIBFulfillment`.

### Test Plan 

Check that AP/IB credit does not count until you set the flag via `GK.enableAPIBFulfillment()`.

### Notes 

By default, this stops AP/IB credit from counting. When pushed to prod, all users will have AP/IB fulfillments removed (maybe this motivates a public change log if people end up being confused).

### DIFF
--- a/src/feature-flags.ts
+++ b/src/feature-flags.ts
@@ -3,6 +3,7 @@ const GateKeeperTogglers: Record<string, () => void> = {};
 
 // Register your new feature flags below, both in type and in `registerFeatureFlagChecker`.
 type FeatureFlagName =
+  | 'APIBFulfillment'
   | 'Case'
   | 'RequirementConflicts'
   | 'RequirementDebugger'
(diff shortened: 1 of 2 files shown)


einc (approved):
Thanks for looking at this so fast Ben!
Tested enable/disabling it for current and new time users and the AP courses seem to apply/not apply as expected. 

On my very first time deploying the link, I did see this in my console log: 
Screen Shot 2022-04-01 at 9 08 24 AM
But I haven't been able to recreate it since so idk if it's related to this PR.

Before we push to prod (end of this sem), we'll definitely make a few small front-end changes alerting current users of this change. Kehui is working on having notifications show up for new alerts, so this would be a good one to have. I'll also probably look into having a small line of text about this in the transfer credits page for current users, and perhaps something about it in onboarding tutorial for new users (depending on how opt in-out progress goes)

PR #677, opened 2022-04-07
Update courses for Fall and Summer 2022

noschiff:
### Summary 

This PR updates our list of courses with offerings from Fall 2022 and Summer 2022.

- [x] ran `npm run courses-gen` and `npm run req-gen`

### Test Plan 

Verify that the site works and has updated course data. Check that a new course such as **ILRHR 5645** can be added and that the bottom bar shows up with expected information for it.

### Notes 

I noticed that there are a lot of courses with malformated values for `catalogWhenOffered`, such as `"Fall or Spring."` or `"Fall, spring."`. I was afraid to go through and fix some of these values because it seems that courses already in a user's plan won't show the updated seasons after being changed in `full-courses.json`, and I was worried that this would break something. I think should make our semester checking more versatile to handle different ways of representing the seasons rather than trying to fix the full courses file.

Before fixing the lowercase s in spring:
example course

After fixing it, the old course is unchanged, but a new one is right:
new
The warning for the fixed course is for it being a duplicate.


willespencer (approved):
Thanks for doing this! Code looks good, and confirmed `ILRHR 5645` can be added on the deployment but not the live site.

Note that new courses have a console error show up because CUReviews data has not been updated yet - but that should not stop us because the front-end displays fine (N/A) and the team has been notified! 
image


Let's get a release PR up during dev meeting tomorrow :)

willespencer:
> I think should make our semester checking more versatile to handle different ways of representing the seasons rather than trying to fix the full courses file.

I agree with this! This is the same issue @andxu282 brought up in slack a couple of weeks ago, but might happen more frequently than we thought. I think there is an item on notion if you wanted to add more details to it!

PR #681, opened 2022-04-13
Honors Physics Track for College of Engineering

noschiff:
### Summary 

Add the honors sequence for physics in the College of Engineering.

- [x] Add PHYS 1116 as an alternate option for PHYS 1112 and add PHYS 2217 as an alternate option for PHYS 2213.
- [x] Update description for physics requirement.

### Test Plan 

Verify that PHYS 1116 can be substituted for PHYS 1112 and that PHYS 2217 can be substituted for PHYS 2213 in the physics requirement for Engineering.

### DIFF
--- a/src/requirements/data/colleges/en.ts
+++ b/src/requirements/data/colleges/en.ts
@@ -38,13 +38,13 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
   {
     name: 'Physics',
     description:
-      'PHYS 1112 and 2213, and, depending on the Major, either PHYS 2214 or a designated mathematics or science course.',
+    'The normal program in physics includes PHYS 1112, PHYS 2213, and PHYS 2214 or the corresponding honors courses (PHYS 1116, PHYS 2217, and PHYS 2218). Depending on the major, a designated mathematics or science course can be substituted for PHYS 2214/2218.',
     source:
-      'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements',
-    checker: includesWithSubRequirements(['PHYS 1112'], ['PHYS 2213']),
+      'https://courses.cornell.edu/content.php?catoid=45&navoid=17985#Engineering_Core_Requirements',
+    checker: includesWithSubRequirements(['PHYS 1112', 'PHYS 1116'], ['PHYS 2213', 'PHYS 2217']),
     fulfilledBy: 'courses',
     perSlotMinCount: [1, 1],
-    slotNames: ['PHYS 1112', 'PHYS 2213'],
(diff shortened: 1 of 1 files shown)


src/requirements/data/colleges/en.ts
```diff
@@ -38,13 +38,14 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
   {
     name: 'Physics',
     description:
-      'PHYS 1112 and 2213, and, depending on the Major, either PHYS 2214 or a designated mathematics or science course.',
+      'The normal program in physics includes PHYS 1112, PHYS 2213, and PHYS 2214 or the corresponding honors courses (PHYS 1116, PHYS 2217, and PHYS 2218).' +
```
willespencer, line 41: Can you get rid of the start of this description? Specifically - "The normal program in physics includes ". I think it just adds extra words when we can just say the course info.

jerrry1123 (approved):
checked generated website and the code, everything behaves properly

src/requirements/data/colleges/en.ts
willespencer: Also - we can keep the second half of the wording the same?

willespencer, line 41: 'PHYS 1112 and 2213, or the corresponding honors courses (PHYS 1116 and PHYS 2217). Depending on the Major, also either PHYS 2214 / PHYS 2218 or a designated mathematics or science course.'

src/requirements/data/colleges/en.ts
```diff
@@ -38,13 +38,14 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
...
+      ' Depending on the major, a designated mathematics or science course can be substituted for PHYS 2214/2218.',
     source:
-      'https://www.engineering.cornell.edu/students/undergraduate-students/curriculum/undergraduate-requirements',
-    checker: includesWithSubRequirements(['PHYS 1112'], ['PHYS 2213']),
+      'https://courses.cornell.edu/content.php?catoid=45&navoid=17985#Engineering_Core_Requirements',
```
willespencer, line 44: Can you leave the link the same? I know yours has the honors courses, but it is specific to 2021-2022 (if you scroll to the top), while the original link updates.

willespencer (changes requested):

src/requirements/data/colleges/en.ts
noschiff: sounds good!

src/requirements/data/colleges/en.ts
noschiff: ok!

src/requirements/data/colleges/en.ts
noschiff: ok!

willespencer (approved):
Looks good and deploy link works! 🥳

PR #682, opened 2022-04-19
Conflict Modal Finishing Touches

willespencer:
### Summary 

This pull request puts the finishing touches on the new course conflict modal.

- [x] Toggle requirements fulfilled
- [x] Moved confirmation modal to display when conflict modal submitted (or just closed!)
- [x] Make self-check courses default to be unchecked 
- [x] **NEW** Add dynamically colored pills to represent req group names 
- [x] **NEW** Do not add course if modal closed without saving 

image

Remaining TODOs (for the next PR):

- [ ] Open the conflict editor from the "FIX NOW" link in the warning tooltip when a conflict remains on a course.
- [ ] Change when errors appear, specifically for courses not assigned to reqs.
- [ ] Don't make changes when reqs edited but modal closed from x button from the "FIX NOW" cta
- [ ] Note that there will be problems with `data()` in the current modal, users have made choices, esp with self checks. Look up graph and edges connected to course.

### Remaining Questions 
- Does pressing the x button act exactly like "save" or "done" depending on if conflicts are resolved?
- Right now, a course's assigned reqs change as checkboxes are checked/unchecked (so you can see the req bar changing on the side). Is this functionality ok?

Depends on #669  (this PR should be merged first)

### Test Plan 

Set major to Computer Science and search for CS 4152 from the add modal.

1. Select only one of the 5 requirements and notice that only that requirement shows as fulfilled in the req bar (without pressing done). Also confirm selecting 0 reqs fulfills nothing. Note that the self-check requirements are unselected by default
3. Select 2 of the 5 requirements and press done. Note that the confirmation popup displays now that the new modal is closed, instead of when it is opened
4. See that the course is highlighted in yellow and has the "FIX NOW" warning display.
image

### Notes 

To enable the GK and use this feature, type GK.enableRequirementConflicts() into the console on the branch or hosted link


einc (commented):
I think pressing “x” button shouldn’t act like “save” or “done”, even if the conflicts are resolved. We should keep it consistent with any time the “x” is pressed and not save the actions. So the class shouldn’t be added to the schedule when “x” is pressed. Any thoughts on this?

einc:
We can delete this error message right? As it's being replaced with "Fix Now" in every scenario now?
Screen Shot 2022-04-20 at 12 18 16 PM

willespencer:
> I think pressing “x” button shouldn’t act like “save” or “done”, even if the conflicts are resolved. We should keep it consistent with any time the “x” is pressed and not save the actions. So the class shouldn’t be added to the schedule when “x” is pressed. Any thoughts on this?

This makes the most sense to me! I just need to re-work the code a bit to do this instead.

> We can delete this error message right? As it's being replaced with "Fix Now" in every scenario now?

You mentioned this before, but I actually don't think it is true (in every scenario). A course can still be added that matches no requirements (say a CS class if you're a hotelie or smthn). 

I think the best way to handle this would be to only show the FIX NOW if a course has conflicts, and show the second one only if it is not matched to anything AND has no conflicts. Does that sound good to you?

benjamin-shen (approved):
lgtm

PR #680, opened 2022-04-13
Spring 2022 Pre-enroll Release

willespencer:
### Summary 

This pull request pushes the course data for next semester (Fall 2022) before Spring 2022 pre-enroll, in addition to new features merged this semester. 🥳 

- [x] #677 (@noschiff)
- [x] #639 (@benjamin-shen)
- [x] #661 #663 #667 (@jerrry1123 @andxu282 @noschiff)
- [x] #655 (@noschiff)
- [x] #647 (@willespencer)
- [x] #664 (@willespencer)
- [x] #668 (@noschiff)
- [x] #618 (@zachary0kent @willespencer)
- [x] #673 (@willespencer)
- [x] #676 (@benjamin-shen)
- [x] #666 (@noschiff) 
- [x] #681 (@noschiff)
- [x] #670 (@jerrry1123) 
- [x] #683 (@willespencer)
- [x] #684 (@willespencer)  
- [x] #686 (@noschiff)
- [x] #665 (@andxu282) 

Other features merged that are gatekept and thus will not show for real users
- [x] #659 (@willespencer)
- [x] #621 (@willespencer)

### Test Plan 

Confirm everything works as expected on all pages of CoursePlan site, especially related to new features, with no console log errors. Examples...
- [x] Add a class with new Fall 2022 course data
- [x] Confirm warning icons show on compact courses.
- [x] Confirm onboarding has entrance grad seasons, and the year dropdowns populate correctly.
- [x] Confirm A&S works as expected, and reqs switch if entrance sem goes before/after Fall 2020.
- [x] Confirm AP/IB code and new conflict code is not accessible unless the feature flag is turned on.

PR #683, opened 2022-04-20
Add new warning to transfer page of onboarding

willespencer:
### Summary 

This pull request adds a new warning to the transfer page of onboarding. The purpose of this warning is to notify users about AP/IB bugs before merging the next release (#680).

image

### Test Plan 

👀 

### Notes 

As a follow-up to this PR, the gatekeep to hide AP/IB fulfillment will be removed.

### DIFF
--- a/src/components/Modals/Onboarding/OnboardingTransfer.vue
+++ b/src/components/Modals/Onboarding/OnboardingTransfer.vue
@@ -52,6 +52,10 @@
             >here</a
           >.
         </div>
+        <div class="onboarding-transferCreditDescription">
+          *Some transfer credits may incorrectly apply to requirements. We are working to fix these
+          bugs, so please double check transfer credits!
+        </div>
       </div>
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


src/components/Modals/Onboarding/OnboardingTransfer.vue
```diff
@@ -52,6 +52,10 @@
             >here</a
           >.
         </div>
+        <div class="onboarding-transferCreditDescription">
+          *Some transfer credits may incorrectly apply to requirements. We are working to fix these
```
noschiff, line 56: I'm not a designer, but I'm not a big fan of the * before the warning (and before the warning above). The text is already small and gray, so I already recognize it as "extra information". Having two `*`s is a little intrusive and uneven IMO.

jerrry1123:
I'm also not a designer, but I feel like the message should at least be the first message and maybe bolded since it's a pretty important thing for students to know and definitely more important than the second message.

noschiff (approved):
This design is much better; thanks Will! LGTM!

einc (approved):
looks good! thanks for adding this so fast <3

PR #684, opened 2022-04-20
Make AP/IB feature flag do nothing (at least temporarily)

willespencer:
### Summary 

This pull request comments out the usage of the AP/IB feature flag created here #676. After discussion with @einc, it seems better to just add a new warning message for AP/IB credit (#683) for the new release (#680), instead of making all current user's lose AP/IB data, due to it automatically counting for some requirements.

By commenting it out instead of removing the flag completely, it can be easily re-introduced after this release if desired.

### Test Plan 

Confirm that AP/IB fulfills requirements, even if the flag has been disabled.

image

### DIFF
--- a/src/requirements/requirement-frontend-utils.ts
+++ b/src/requirements/requirement-frontend-utils.ts
@@ -264,9 +264,9 @@ export function getMatchedRequirementFulfillmentSpecification(
   ) =>
     coursesList.map(courses =>
       courses.filter(courseId => {
-        // do not include ap/ib course if feature flag is not toggled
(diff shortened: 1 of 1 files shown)


noschiff:
definitely agree with this change for now!

jerrry1123 (approved):
Tested the preview and made sure that changing the flag does not affect if AP's satisfy requirements

noschiff (approved):
I ran `GK.disableAPIBFulfillment()` and added AP Chem, and it works!
image
image

PR #690, opened 2022-04-30
Tools Page, Advisors Card, and Advisors Backend

noschiff:
### Summary 

Creates customizable tool card component, organizes tools page, and implements advisor tool card.

`Card.vue` is the tool card component, and it takes in a slot another component. `AdvisorCard.vue` is the first type of component implemented for the tools page, and it is used as a child element of a `Card.vue` on the `Tools.vue` container.

- [x] Create tool card component according to design
- [x] Implement collapsing and expanding tool card
- [x] Keep tool cards from changing position on collapse/expansion
- [x] Save collapsed state of individual tool card in browser
- [x] Use tool card props for different desired sizes
- [x] Implement basic organization of tool cards on tools page
- [x] Create advisor card
- [x] Implement front-end components for advisor content table
- [x] Add link icon to email
- [x] Create types for advisors and packages of information and add them to existing requirement architecture
- [x] Write checkers for advisor based on user's last name
- [x] Add advisors for CS major & Engineering college

### Test Plan 

Run `GK.enableTools()`.
Add majors, minors, and colleges that we have advisor data for. Go to the tools page and verify that the advisors are correct based on any checker logic are in the right order (college -> major -> minor). Also make sure the amount of advisors is capped (default is 5 max).

image

### Notes 

- Should we keep the Gatekeep for the tools page?
- Should the checkers be an array like for requirements, and are there any other things we should be able to check aside from first/middle/last name?
- I excluded the advisors from the decorated requirements because I'm doing all the computation by reading the index file directly.
- Do I include the advisor type on the card?
- Is there anything wrong with using pixels?
- What happens when there are no advisors?
- What do I do with source?

Future TODOs for another PR:

From Yuxuan:

- card grows and always has 48px margin bottom
Screen Shot 2022-05-04 at 9 12 20 PM

- let cards reposition when collapsed
Screen Shot 2022-05-04 at 9 16 15 PM

- implement design for no advisors

- Strengthen organization of cards in tools page. Currently, tools will maintain their position when collapsed or expanded, so that is not an issue. However, the currently layout will place cards horizontally then wrap based on space.

https://user-images.githubusercontent.com/47431797/166091905-f7e81744-758a-448f-b87a-d32916a30ab8.mov


src/components/Tools/Card.vue
```diff
@@ -0,0 +1,132 @@
...
+@import '@/assets/scss/_variables.scss';
+
+.card {
+  width: var(--w);
+  height: var(--h);
+  box-sizing: border-box;
+  box-shadow: 0 0 10px 4px $boxShadowGray;
+  position: relative;
+  border-radius: 10px;
+  border-width: 0;
```
noschiff, line 69: Does anyone know why I need to set this to 0 explicitly? This is the highest scope (.card), so I don't know why it has a nonzero value by default.

noschiff:
note from Will: make hidden button not a real button

jessfeng (changes requested):
Woohoo! Our first card!! Thanks for working on this, Noah! ✨ Love how the advisors correspond to the student's last names, college, and majors! Do we maybe want to include the general advising office in case students want to reach them (ie. Engineering Advising)?

A couple things I've noticed so far, just interacting with the preview URL:
- When the page narrows until the hamburger appears, it kind of covers the "Hi Courseplanner" title, whereas this doesn't happen on prod
Preview: 
Screen Shot 2022-05-06 at 3 52 06 AM
Prod:
Screen Shot 2022-05-06 at 3 55 12 AM

- Tested page re-sizing and formatting seems to be a bit wonky, not sure how we want the card to look like if it's this compressed. Maybe @yuxuanchenn has a better answer to this? The subtitle under the header also becomes an ellipse on the preview url compared to the prod version--I feel like we should keep the text and not turn it into an ellipse but let's consult the designers. 
Preview:
Screen Shot 2022-05-06 at 4 12 32 AM
Prod:
Screen Shot 2022-05-06 at 4 12 51 AM


What I tested:
- [x] A single major (Computer Science): college and major advisors look good and correct
- [x] A single major not in College of Engineering: empty card (this seems to be the correct behavior as of now because we have no data on advisors in non-eng, non-cs advisors)
- [x] A single major in CoE but not CS: displays eng advisors, so looks good for now
- [x] Last name `Feng` works
- [x] Last name `Z` works
- [x] **Last name lowercase doesn't**  --> this only displays Alexandria (default engineering advisor) but no last-name specific engineering advisor and no last-name specific cs advisor
  - [ ] same thing happens if there's a space before the last name (ie. ' Aang')

Note: haven't looked too deeply into the code yet though!

noschiff:
Hi @jessfeng, thanks for the feedback! I'm not sure what you mean by including the general advising office in case students want to reach them. I included a required field for the source of our advisor data in the code, so we have a link to at least where we see that the shown advisors are the advisors. Is that what you mean, or would you want something even more general like the Engineering Advising home page? Either way, I think I brought this up in crit, but we don't currently have anywhere to display the source or any other link in the current design. How would you like me to try to include this information?

jessfeng:
> Hi @jessfeng, thanks for the feedback! I'm not sure what you mean by including the general advising office in case students want to reach them. I included a required field for the source of our advisor data in the code, so we have a link to at least where we see that the shown advisors are the advisors. Is that what you mean, or would you want something even more general like the Engineering Advising home page? Either way, I think I brought this up in crit, but we don't currently have anywhere to display the source or any other link in the current design. How would you like me to try to include this information?
> 
> I'm not sure what the intended behavior is when the screen is shrunk so that the card cannot display all the information. Once we decide what should happen, I'll be happy to fix this!

The changes look good--whitespace and lowercasing last names work as expected now! 

As for the general advising office, I meant something like the Cornell Eng Advising office like Yuxuan put on the Figma file, so we minimize the amount of people with the empty state contact advisor card. Not sure if all colleges have a contact like this, but I know I've emailed them a couple times about transfer credits among other things, so it might be useful to have that on there if they need to talk to anyone. It's just a minor thing, so if it's too much work, we can just table it and add it to the backlog.

Screen Shot 2022-05-10 at 5 17 42 PM

As for shrinking the screen, we can do that in the next PR once Yux gets back to us but other than that, I think it looks good to go!

noschiff:
> > Hi @jessfeng, thanks for the feedback! I'm not sure what you mean by including the general advising office in case students want to reach them. I included a required field for the source of our advisor data in the code, so we have a link to at least where we see that the shown advisors are the advisors. Is that what you mean, or would you want something even more general like the Engineering Advising home page? Either way, I think I brought this up in crit, but we don't currently have anywhere to display the source or any other link in the current design. How would you like me to try to include this information?
> > I'm not sure what the intended behavior is when the screen is shrunk so that the card cannot display all the information. Once we decide what should happen, I'll be happy to fix this!
> 
> The changes look good--whitespace and lowercasing last names work as expected now!
> 
> As for the general advising office, I meant something like the Cornell Eng Advising office like Yuxuan put on the Figma file, so we minimize the amount of people with the empty state contact advisor card. Not sure if all colleges have a contact like this, but I know I've emailed them a couple times about transfer credits among other things, so it might be useful to have that on there if they need to talk to anyone. It's just a minor thing, so if it's too much work, we can just table it and add it to the backlog.
> 
> Screen Shot 2022-05-10 at 5 17 42 PM
> 
> As for shrinking the screen, we can do that in the next PR once Yux gets back to us but other than that, I think it looks good to go!

Oh, I see what you're talking about! Yeah, I think we could include that! That would require a little bit of work to do nicely, and I don't have time for that right now with finals. I want to probably implement Yuxuan's final design for the empty state first since we don't have much data yet, but it would definitely be good to include something like that.

jessfeng (approved):
Alright, thanks for addressing my comments + future PR to-dos, this looks good to me!

PR #691, opened 2022-05-01
Add transfer credits and save confirmation modal

jerrry1123:
### Summary 
- Added the second page in the profile which edits transfer credits and the required buttons to navigate between the two pages
- Added Confirmation Modal which will be displayed when a user tries to save their work
- Added logic checks so that users can only save valid profile data and made sure they can only save once they have actually made changes

Remaining TODOs:

- Fix the bug where scrolling is permanently disabled on the profile page if the confirmation modal is closed with the left or right buttons (but not the x) 
- Merge the changes from the profile redesign PR with this PR (@willespencer so please ignore the stuff you already pointed out in ProfileRedesign - I would do this myself but I don't want to mess with Git so I'd prefer doing this in person)

Depends on profile redesign PR which should be merged soon into master

Test:
- Made sure that changes are only saved when confirmed in the confirmation modal and that they are properly discarded if not
- Made sure that the save button is only active when valid changes are available
- Made sure that the proper error message displays when the user inputs invalid data e.g. no college, year, etc.


jerrry1123:
@jessfeng all the comments you left on the last PR should be addressed here. I wasn't sure what to do with narrow pages squishing stuff together since we need space on the page for the left buttons so the actual form is more squished. So I just left it as is for now.

noschiff (changes requested):
Looking good! A few issues:

- The save button is initially inactive, but after making a change and clicking save, the save button remains active. The changes look to be saved when I refresh, so this looks to be a front-end only issue.
- The font for the left side bar doesn't match the design.
- The spacing for the buttons and dropdowns for the Transfer Credit page doesn't match the design and is not consistent between AP Credits and IB Credits. (see screenshot)

image

You're using space-between for the "x" button and dropdowns, and this is causing them to spread out when the screen is wider like on my laptop. Maybe try to group the x button together with the score dropdown?
Screen Shot 2022-05-04 at 11 37 37 AM

jerrry1123:
@noschiff I just fixed the issue with the save button. As for your second point, did you mean the font spacing or the actual font itself? I copied the CSS from Kehui's design so it should be the same. Finally, the buttons inside the transfer credits element are set by what we made in our onboarding modals, since these pages just have copies of the modals we use in onboarding. If I want to change their spacing, I would also need to do the same change to the onboarding modal, which I don't feel comfortable tampering with right now.

jessfeng:
> @noschiff I just fixed the issue with the save button. As for your second point, did you mean the font spacing or the actual font itself? I copied the CSS from Kehui's design so it should be the same. Finally, the buttons inside the transfer credits element are set by what we made in our onboarding modals, since these pages just have copies of the modals we use in onboarding. If I want to change their spacing, I would also need to do the same change to the onboarding modal, which I don't feel comfortable tampering with right now.

michaelfarkouh:
Hi Jerry,
I agree with Noah's points above. I will say this all looks great but I also have some notes below:
1) There is an error in the logic as entrance year and season can be after the graduation year and season if the years are the same but the season for entrance is earlier. (Ex. Entrance: Fall, 2025 and Grad: Spring, 2025 is accepted as a valid entry). Of course the order of the seasons is (winter, spring, summer, fall) and so if the entrance season is after the grad season for the same year this should not be valid.

2) When you make the window smaller the styling of the underline for the name of the page you are on overlaps with the text which may make it hard to read on smaller windows. See photo
Screen Shot 2022-05-09 at 12 15 02 AM


3) I know this is not because of the code for the new pages but from a previous error, however, I notice that we do allow major/minor combinations that are not allowed. It wouldn't make sense to necessarily fix that here but something to note.

4) If I make a change and then change it back to the state that was previously saved the save button still comes up even though the state is the same as what was previously saved. I assume this would require changing the logic for which you are checking whether a change has been made, it is not the end of the world if this has to stay to reduce other edge cases.

5) When I "cancel" an unsaved change it is not discarded as you write it should be in the testing plan, rather it works as if I clicked the x in the corner and went back to the state before clicking save.

6) Error messages look good, though currently we do not have an error message if you put a subject for transfer credit but no score is inputed. I don't think this is a huge concern like something required, but if its fixable than something to note.

7) Finally, as the layout for the sub pages does feel like each is its own page so it may make sense to require a save to be made or notify the user that unsaved change will be discarded as you leave one sub page to the next (ex Background Information to Transfer Credit)

Sorry for a lot of the comments that may not be relevant to your redesign of the profile page but rather the underlying code, but hopefully they can be addressed at some point. Great work though and it looks good!

michaelfarkouh (changes requested):
Hi Jerry,
I agree with Noah's points above. I will say this all looks great but I also have some notes below:
1) There is an error in the logic as entrance year and season can be after the graduation year and season if the years are the same but the season for entrance is earlier. (Ex. Entrance: Fall, 2025 and Grad: Spring, 2025 is accepted as a valid entry). Of course the order of the seasons is (winter, spring, summer, fall) and so if the entrance season is after the grad season for the same year this should not be valid. Note that this error does not occur on Prod meaning I believe it should be an error in this code.

2) When you make the window smaller the styling of the underline for the name of the page you are on overlaps with the text which may make it hard to read on smaller windows. See photo shoes for background information page but same is true for transfer credit page.
Screen Shot 2022-05-09 at 12 15 02 AM


3) I know this is not because of the code for the new pages but from a previous error, however, I notice that we do allow major/minor combinations that are not allowed. It wouldn't make sense to necessarily fix that here but something to note.

4) If I make a change and then change it back to the state that was previously saved the save button still comes up even though the state is the same as what was previously saved. I assume this would require changing the logic for which you are checking whether a change has been made, it is not the end of the world if this has to stay to reduce other edge cases.

5) When I "cancel" an unsaved change it is not discarded as you write it should be in the testing plan, rather it works as if I clicked the x in the corner and went back to the state before clicking save.

6) Error messages look good, though currently we do not have an error message if you put a subject for transfer credit but no score is inputed. I don't think this is a huge concern like something required, but if its fixable than something to note.

7) Finally, as the layout for the sub pages does feel like each is its own page so it may make sense to require a save to be made or notify the user that unsaved change will be discarded as you leave one sub page to the next (ex Background Information to Transfer Credit)

Sorry for a lot of the comments that may not be relevant to your redesign of the profile page but rather the underlying code, but hopefully they can be addressed at some point. Great work though and it looks good!

einc (approved):
Hi Jerry! Thanks for addressing all of those changes. Just one last thing that I wasn't sure about. If a user makes changes, but doesn't save, and tries to leave the page the confirmation modal should show up. I don't think it currently does.

The confirmation modal doesn't show up only 1) they didn't make any changes, 2) they saved it correctly. Otherwise, if users click onto other pages, it should also remind users to save

noschiff (approved):
Thanks for addressing those changes! I still think it's very important that we address the issues I brought up with spacing on the transfer credits page because they are more obvious on a bigger screen, so we just need to make sure we get to that in a future PR.

michaelfarkouh (approved):

jerrry1123:
Thanks for the feedback. I'll make a note to address these things next semester

PR #694, opened 2022-05-05
add psych major

kkkehui:
**Add Psychology Major**

- reference link: https://psychology.cornell.edu/major
- 40-credit requirement, statistics requirement, 1 courses from each of the three areas

Question:
- Psych major only allows a max of 9 credits of independent study/research work counting towards the 40 credits needed to fulfill the major requirement. I am not sure how to check that.
(hopefully other things all work, thanks!)


src/requirements/data/majors/psych.ts
```diff
@@ -0,0 +1,202 @@
...
+} from '../checkers-common';
+
+const psychRequirements: readonly CollegeOrMajorRequirement[] = [
+{
+    name: 'Credit Requirements',
+    description:
+        'A total of 40 credits in psychology (including prerequisites), from which students majoring in psychology are expected to choose, in consultation with their advisors, a range of courses that covers the basic processes in psychology.',
+    source:
+        'https://psychology.cornell.edu/major',
+    checker: includesWithSingleRequirement(
```
noschiff, line 16: don't we have a function to include all PSYCH prefix courses? that'd be cleaner and more future-proof than listing very PSYCH course

benjamin-shen: I agree! We don't have a function (yet) but it should be easy to refactor. For now we can use wildcards: `includesWithSingleRequirement('PSYCH ****')`.

PR #707, opened 2022-08-27
Bump dependencies

benjamin-shen:
### Summary 

This pull request bumps a bunch of dependencies for security reasons & ci/cd fixes.

- [x] firebase web v8 => web v9 (this resulted in the majority of code changes)
- [x] removed definitely-unused dependencies

Remaining TODOs:

- [ ] firestore syntax changes in scripts, if necessary
- [ ] remove vue 2 dependencies (see if we can remove `bootstrap-vue`, [relevant issue](https://github.com/bootstrap-vue/bootstrap-vue/issues/5196))

### Test Plan 

Check out to this PR, `rm -rf node_modules`, `npm install`, `npm run serve`, sanity check local app.
Do the same with the `functions/` directory and check that nothing seemed to break.

### Breaking Changes  

- Everyone will need to update dependencies on their machines, `npm install --force`.


cypress/support/commands.js
```diff
@@ -24,10 +24,10 @@
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
 
-import firebase from 'firebase/app';
-import 'firebase/auth';
-import 'firebase/database';
-import 'firebase/firestore';
+import firebase from 'firebase/compat/app';
```
noschiff, line 27: Thanks so much Ben for updating our firebase code to the modular Web SDK! Is there any reason we still need to use the version 8 SDK through `compat` for cypress?

benjamin-shen: That's what's suggested by the readme for [cypress-firebase](https://github.com/prescottprue/cypress-firebase) (look for "SDK version 9"). I imagine this is so they don't have major discrepancies between their versions, and their implementation can keep using the version 8 SDK.

zachary0kent (approved):

zachary0kent:
I agree that we should remove the `functions` directory though, most of the scripts there fail because there aren't any firebase functions actually implemented.

PR #709, opened 2022-08-31
Add psych major

benjamin-shen:
### Summary 

This pull request adds the [psychology major](https://psychology.cornell.edu/major). Moved the branch in @kkkehui's fork (#694) to an internal PR.
- 40-credit requirement
- statistics requirement
- 1 course from each of the three areas

### Test Plan 

Test the psych major 😃 

Verify that the ["psych ****" commit](https://github.com/cornell-dti/course-plan/pull/709/commits/cd36061edd54bff4aec9752e6ba970d6d6d862b6#diff-38bfbe5fd0f55f8bc0869e1e39f5a52bfc0624e24709f1a69017aa7c448f3b1c) didn't add courses that aren't supposed to be there.

### Notes 

Psych major only allows a max of 9 credits of independent study/research work counting towards the 40 credits needed to fulfill the major requirement. I am not sure how to check that.


src/requirements/data/majors/psych.ts
```diff
@@ -0,0 +1,130 @@
...
+        'PSYCH 3850',
+        'PSYCH 4050',
+        'PSYCH 4300',
+        'PSYCH 4810',
+        'PSYCH 4820',
+        'PSYCH 4840',
+        'PSYCH 4850',
+        'PSYCH 4910',
+      ],
+      ['PSYCH 1101', 'PSYCH 1650', 'PSYCH 1990', 'PSYCH 2500', 'PSYCH 4700', 'PSYCH 4710']
```
noschiff, line 106: I don't understand the purpose of the fourth group since the minimum required is 0. Also, it looks to not even show up in the requirements sidebar at all, which seems like it could be a bug. I think that we would want to still allow users to add courses (potentially unlimited?) from the sidebar for slots that have a minimum of 0. Maybe this could be a special case, because with our current implementation, a 0 minimum seems to be useless.
image

benjamin-shen: I agree, I think they should be strictly opt-in.

noschiff: Is it also worth considering handling a 0 min in general, or should we just always use opt-in for that?

noschiff: I think just using opt-in in fine.

noschiff: For this req data, I think we can just remove those "other"s.

benjamin-shen: I don't think it would ever make sense to intentionally put 0 min, since then it means the slot is always automatically fulfilled. Maybe it makes sense to handle that edge case in the case of bad data on our part, or programmatically block this kind of data.

zachary0kent (approved):
lgtm now

PR #712, opened 2022-09-05
Added a warning to the profile page before leaving and fixed spacing issue on transfer credits

jerrry1123:
### Summary Added 
Created a new modal which warns the user if they try to leave the profile page with unsaved changes
Fixed the spacing issues on the AP/IB exam credits page

This pull request is hopefully the last step we need for the profile page

Remaining TODOs:
- None

### Test Plan 
Test that the behaviour for switching between the Plan, Tools, and Profile Pages behaves as expected
- Try all possibilities of navigating between the pages
Test that the save reminder modal only pops up when changes were made
Test that the new spacing works for screens of all sizes
Test that the onboarding modal is unaffected by the changes to the profile page

New Spacing: 
image
New Modal: 
image

### DIFF
--- a/src/components/Modals/Onboarding/Onboarding.scss
+++ b/src/components/Modals/Onboarding/Onboarding.scss
@@ -432,6 +432,7 @@ input {
       width: 380px;
     }
     &--columnFill {
+      margin: 5px;
       display: flex;
       flex-direction: column;
       width: 100%;
--- a/src/components/Modals/Onboarding/Onboarding.scss
+++ b/src/components/Modals/Onboarding/Onboarding.scss
@@ -302,7 +302,6 @@ input {
     cursor: pointer;
     margin-left: 1rem;
     color: $emGreen;
-    padding-bottom: 0.5rem;
   }
 
   &-bottom {
@@ -424,7 +423,7 @@ input {
     }
     &--column-removeExam {
       display: flex;
-      flex-direction: column-reverse;
+      flex-direction: column;
     }
     &--columnWide {
       display: flex;
@@ -432,7 +431,7 @@ input {
       width: 380px;
     }
     &--columnFill {
-      margin: 5px;
+      margin: 1rem;
       display: flex;
       flex-direction: column;
       width: 100%;
--- a/src/components/Modals/Onboarding/Onboarding.scss
+++ b/src/components/Modals/Onboarding/Onboarding.scss
@@ -431,7 +431,7 @@ input {
       width: 380px;
     }
     &--columnFill {
-      margin: 1rem;
+      margin: 0.313rem;
       display: flex;
       flex-direction: column;
       width: 100%;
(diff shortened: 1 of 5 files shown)


zachary0kent:
Looking a lot better! The score dropdowns look pretty wide though--could we reduce their width?

jerrry1123:
> Looking a lot better! The score dropdowns look pretty wide though--could we reduce their width?

I didn't have any designs to base the margins off of, but we can settle on a width which we think looks good at our meeting tomorrow

zachary0kent:
Also, is there a reason we're not autosaving changes?

src/components/Modals/ProfileSaveReminder.vue
```diff
@@ -0,0 +1,34 @@
...
+    rightButtonText="Continue"
+    content-class="content-clear"
+    @modal-closed="returnToProfilePage"
+    @left-button-clicked="returnToProfilePage"
+    @right-button-clicked="confirmLeaveProfilePage"
+  >
+    <span
+      >There may be unsaved changes in your profile! Please make sure your changes are saved before
+      leaving!</span
+    >
```
zachary0kent, line 14: Can we clean up this formatting a bit?

src/components/Modals/Onboarding/Onboarding.scss
```diff
@@ -432,6 +432,7 @@ input {
       width: 380px;
     }
     &--columnFill {
+      margin: 5px;
```
zachary0kent, line 435: Let's prefer `rem` over `px`

noschiff:
> Also, is there a reason we're not autosaving changes?

Honestly I think it'd be better to auto save changes, but I think this was a design choice a while ago… @zachary0kent

zachary0kent:
@jerrry1123 How much work would it be to add e2e tests for this?

src/components/Modals/ProfileSaveReminder.vue
jerrry1123: this was the formatting suggested by VSCode

noschiff (changes requested):
When the user goes from profile to tools, it takes the user to the plan instead of tools. I believe this has to do with how the new profile page was added (using `v-if`s on boolean flags).

https://user-images.githubusercontent.com/47431797/197672490-0967a37b-6088-41a8-bcc5-2dc915d23a71.mov

zachary0kent (changes requested):
It seems you only get a warning when exiting the profile page with unsaved changes once. If you press "go back" and then navigate to another page, you don't get the warning.

https://user-images.githubusercontent.com/73757337/215859136-8e0ad638-108e-4ead-a8e7-dd05b16c86b4.mov

zachary0kent:
Could you also make the PR title a bit more concise?

zachary0kent:
> I think a bug was reintroduced somewhere from the first semester of the profile page where opening the unsaved changes modal would permanently disable the scroll bar in the profile page. Can someone else check if this is also happening on their computers or if it's just me?

Yes I encounter this as well. Is it a regression?

jerrry1123:
Yes I remember it was an issue with teleport modals the last time, but the modals I added this way were the exact same as the modals that didn't cause issues so I'm not sure what the difference is.

PR #715, opened 2022-09-26
Fix Script Copy Data Between Users

noschiff:
### Summary 

This PR provides a quick fix to the script introduced in #711. `initializeApp()` from `firebase-admin` was missing a specific name in order to initialize multiple apps, which is necessary to copy data between dev and prod. This PR also changes the name of the dev service account to make the script clearer.

### Test Plan 

Run the script from the command line with various arguments in preview mode and actually copy between documents on dev (e.g. copy from your account on dev to another google account that you have). You don't actually have to copy to another user if you want; the "user" only refers to the name of a document, so you can name your documents what you want. However, you'll need an email that corresponds to the document name to actually use the data. I recommend only having the dev service account in your repo at first until you have **thoroughly read through the code** and are confident that it will behave as it is supposed to once a prod service account credentials are in the repo.

Example script: preview a copy from a dummy account to your data on dev:
```
npm run ts-node -- scripts/copy-user-data.ts -f dev/dummyaccount -t dev/your-netid@cornell.edu -o "log.json"
```
Example script: copy from a dummy account to your data on dev:
```
npm run ts-node -- scripts/copy-user-data.ts -f dev/dummyaccount -t dev/your-netid@cornell.edu -o "log.json" --execute
```

### DIFF
--- a/scripts/copy-user-data.ts
+++ b/scripts/copy-user-data.ts
@@ -3,7 +3,7 @@
  * BE CAREFUL WHEN COPYING TO PROD!
  * Script to copy data from one user on production or dev to another user on dev.
  * Requires service accounts for database.
- * serviceAccount.json (if using dev) and serviceAccountProd.json (if using prod) must be at the root.
+ * serviceAccountDev.json (if using dev) and serviceAccountProd.json (if using prod) must be at the root.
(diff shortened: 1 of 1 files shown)


zachary0kent (approved):
lgtm, tested it locally

PR #692, opened 2022-05-03
Resolve existing conflicts from FIX NOW warning

willespencer:
### Summary 

This pull request lets users resolve saved conflicts from the warnings that show on the course or in the requirements sidebar.

- [x] Refactored `CourseConflictModal` to handle `CourseTaken` type from the reqs sidebar.
- [ ] Open the conflict editor from the "FIX NOW" link in the warning tooltip when a conflict remains on a course.
- [ ] Change how self-checks are found.
- [ ] Have the reqs that the course applies to be the ones that are selected.

Remaining TODOs:

- [ ]  Change when errors appear, specifically for courses not assigned to reqs.
- [ ]  Don't make changes when reqs edited but modal closed from x button from the "FIX NOW" cta

### Test Plan 

Add CS 4152 to plan as a CS major, and press Save after checking a self-check req.

1. Confirm that the conflict modal can be opened from a warning on the course - and the same 3 reqs that the course is in conflict with are selected. Remove one and press save.
2. Confirm the conflict modal can be opened from reqs bar for one of the reqs with the conflict still - open it and uncheck one to finish resolving conflict.
3. Finally, confirm pressing x button on conflict modal leads to no changes of checked/unchecked boxes.

### Notes 

After this PR is merged - the gatekeep can be removed in a final PR and everything can be tested thoroughly.

PR #689, opened 2022-04-29
Implement opt in/out functionality with new overridden fulfillments model

benjamin-shen:
### Summary 

This pull request implements opt in/out and introduces functions that write user choice to the database. 

- [x] implemented opt in/out in `global-firestore-data/override-fulfillment-choices.ts`
- [x] refactored adding selectable requirements to overridden fulfillments

#### TODO
- [ ] rename file `override-fulfillment-choices.ts` => `overridden-fulfillment-choices.ts`

### Test Plan 

*Please see note*

Selectable requirements are the main thing that changed. Opt in/out is not finished on the frontend yet, and they probably don't need to be tested right now.

### Notes 

This PR relies on the readiness of the existing course conflict modal.

PR #716, opened 2022-09-27
Fix overridden fulfillments race condition in course conflict modal

benjamin-shen:
### Summary 

This pull request fixes the overridden fulfillments race condition seen in #692. In addition, it
- [x] renames `override-fulfillment-choices.ts` to `overridden-fulfillment-choices.ts`, as mentioned in #689
- [x] removes the comparison to `originalReqsPerConflict` (we shouldn't need to compare to the original state in order to update the user data)

#### TODO  

We might want to change how we represent `selectedReqsPerConflict` in `CourseConflictModal.vue`. For example, instead of mapping reqName => boolean, we could store two sets (one for selected requirements and one for unselected requirements).

We need to add some e2e tests for this modal.

### Test Plan 

1. Enable requirements conflicts in the browser console, i.e. `GK.enableRequirementConflicts()`
2. Add CS 4152 as a CS major (or any other class that would cause a conflict
3. Open the conflict modal by clicking on "FIX NOW" in the warnings
4. Select different combinations of requirements and make sure they save properly and are reflected in the requirements graph/debugger (`GK.enableRequirementDebugger()`).


src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -187,32 +184,32 @@ export default defineComponent({
...
+        );
       }
 
       this.closeCurrentModal();
       this.$emit('resolve-conflicts', this.selectedCourse);
     },
     // req is self check if present in selectableRequirements list
     isReqSelectable(reqName: string) {
-      return this.selectableRequirements.filter(req => req.id === reqName).length > 0;
+      return !!this.selectableRequirements.find(req => req.id === reqName);
```
zachary0kent, line 212:
```suggestion
      return this.selectableRequirements.some(req => req.id === reqName);
```

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -187,32 +184,32 @@ export default defineComponent({
...
-              'optOut'
-            );
-          }
-        });
+        const userChoices = Array.from(this.selectedReqsPerConflict[i].entries());
+
+        const naturallyFulfilledRequirementIds = userChoices
+          .filter(([reqName, selected]) => selected && !this.isReqSelectable(reqName))
+          .map(([reqName]) => store.state.userRequirementsMap[reqName].id);
+        const arbitraryOptInRequirementIds = userChoices
```
zachary0kent, line 192: The "arbitrary" terminology is very confusing IMO but it's already become standard in many places. I think in the future it might be worthwhile to change it.

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -187,32 +184,32 @@ export default defineComponent({
     resolveConflicts() {
       // edit the requirements assigned to the course when editor saved on reqs with changes made
       for (let i = 0; i < this.selectedReqsPerConflict.length; i += 1) {
```
zachary0kent, line 186: Why not for-each loop?

src/global-firestore-data/overridden-fulfillment-choices.ts
```diff
@@ -134,3 +124,39 @@ export const addArbitraryOptIn = (
...
+/**
+ * Precondition: the requirement id lists are disjoint
+ */
+export const resolveConflicts = (
+  courseUniqueID: string | number,
+  naturallyFulfilledRequirementIds: string[],
+  arbitraryOptInRequirementIds: string[],
+  optOutRequirementIds: string[]
+): void =>
+  updateRequirementChoice(
```
zachary0kent, line 137: There is too much code in this single function call IMO

src/components/Modals/NewCourse/CourseConflictModal.vue
```diff
@@ -187,32 +184,32 @@ export default defineComponent({
...
-              this.courseUniqueId,
-              store.state.userRequirementsMap[reqName].id,
-              'optOut'
-            );
-          }
-        });
+        const userChoices = Array.from(this.selectedReqsPerConflict[i].entries());
+
+        const naturallyFulfilledRequirementIds = userChoices
+          .filter(([reqName, selected]) => selected && !this.isReqSelectable(reqName))
```
zachary0kent, line 190: JS really does need a `filterMap`

src/global-firestore-data/overridden-fulfillment-choices.ts
```diff
@@ -134,3 +124,39 @@ export const addArbitraryOptIn = (
...
   );
+
+/**
+ * Precondition: the requirement id lists are disjoint
+ */
+export const resolveConflicts = (
+  courseUniqueID: string | number,
+  naturallyFulfilledRequirementIds: string[],
+  arbitraryOptInRequirementIds: string[],
+  optOutRequirementIds: string[]
```
zachary0kent, line 135: I think it's better to pass these in as an options object

zachary0kent (commented):
looks good overall 👍

src/components/Modals/NewCourse/CourseConflictModal.vue
benjamin-shen: Before it was implemented in a way that depended on `i`. Should be changed!

src/global-firestore-data/overridden-fulfillment-choices.ts
benjamin-shen: Can you be more specific @zachary0kent?

src/global-firestore-data/overridden-fulfillment-choices.ts
benjamin-shen: How do you think I should refactor it?

PR #719, opened 2022-10-14
Improve Tool Cards and Tools Page

noschiff:
### Summary 

This PR adds the four tool cards we intend to implement in the near future to the Tools page and creates the styling for the placement of the cards. In particular, it handles collapsing and resizing for all four toolcards. This PR also improves the Advisors Tool Card introduced in #690, namely by fixing the CSS and removing the cap on the number of advisors.

Please give me feedback on the behavior for when the screen width is changed.

### Test Plan 

Run `GK.enableTools()`.
Add majors, minors, and colleges that we have advisor data for.. 

image


src/components/Tools/AdvisorCard.vue
```diff
@@ -53,7 +53,9 @@ export default defineComponent({
         ...this.onboardingData.minor
           .map(acronym => getAdvisor(acronym, 'minor', this.userInfo))
           .flat(),
```
zachary0kent, line 55:
```suggestion
          .flatMap(acronym => getAdvisor(acronym, 'minor', this.userInfo))
```

src/components/Tools/AdvisorCard.vue
```diff
@@ -2,16 +2,16 @@
   <div class="advisors">
     <div class="advisors-content">
       <div class="advisors-heading-row">
-        <span class="advisors-info, advisors-heading-content">Type</span>
-        <span class="advisors-info, advisors-heading-content">Name</span>
-        <span class="advisors-info, advisors-heading-content">Email</span>
+        <span class="advisors-info advisors-heading-content">Type</span>
+        <span class="advisors-info advisors-heading-content">Name</span>
+        <span class="advisors-info advisors-heading-content">Email</span>
```
zachary0kent, line 7: Can replace this with a `v-for`

src/components/Tools/AdvisorCard.vue
```diff
@@ -2,16 +2,16 @@
...
       </div>
       <div v-for="(advisor, index) in advisors" :key="index" class="advisors-row">
-        <span class="advisors-info,advisors-type"> {{ advisor.type }} </span>
-        <span class="advisors-info, advisors-name">
+        <span class="advisors-info advisors-type"> {{ advisor.type }} </span>
+        <span class="advisors-info advisors-name">
           {{ advisor.name }}
         </span>
-        <a class="advisors-info, advisors-email" :href="'mailto:' + advisor.email">
+        <a class="advisors-info advisors-email" :href="'mailto:' + advisor.email">
```
zachary0kent, line 14:
```suggestion
        <a class="advisors-info advisors-email" :href="`mailto:${advisor.email}`">
```

src/components/Tools/AdvisorCard.vue
```diff
@@ -85,25 +87,29 @@ export default defineComponent({
     color: #3d3d3d;
     align-self: center;
     justify-self: start;
-    margin-left: 0.25rem;
-    margin-right: 0.25rem;
+    //margin-left: 0.25rem;
+    //margin-right: 0.25rem;
```
zachary0kent, line 91: Can we delete this styling if it's unused

src/components/Tools/AdvisorCard.vue
```diff
@@ -53,7 +53,9 @@ export default defineComponent({
         ...this.onboardingData.minor
           .map(acronym => getAdvisor(acronym, 'minor', this.userInfo))
           .flat(),
-      ].slice(0, this.maxItems);
+      ];
+      if (this.maxItems !== undefined) return packages.slice(0, this.maxItems);
+      return packages;
```
zachary0kent, line 58:
```suggestion
      return packages.slice(0, this.maxItems);
```
If `maxItems` is `undefined` it'll select all of the items.

src/containers/Tools.vue
```diff
@@ -9,12 +9,35 @@
...
-      <!--      <card name="Another Card" class="toolsContainer-card" id="x" :desired-width="400">-->
-      <!--        <h1>This one has a smaller width than default!</h1>-->
-      <!--      </card>-->
+
+      <card
+        name="Export Schedule"
+        class="toolsContainer-card"
+        id="export"
+        :desired-width="400"
+        :desired-height="100"
```
zachary0kent, line 28: What's up with the `desired` prefix on the dimensions? Also, can we make these dimensions the default so we don't have to specify them?

src/containers/Tools.vue
```diff
@@ -56,13 +79,44 @@ export default {
...
+
+    > :nth-child(2) {
+      grid-column: 1/2;
+      grid-row: 2/3;
+    }
+
+    > :nth-child(4) {
+      grid-column: 2/3;
+      grid-row: 2/3;
+    }
```
zachary0kent, line 107: What is the purpose of this styling? I think it should be documented here.

src/components/Tools/AdvisorCard.vue
noschiff: lol yup i forgot… that was part of my experimentation

src/containers/Tools.vue
noschiff: the default is wider than 400 (see the bigger ones)

src/components/Tools/AdvisorCard.vue
noschiff: that's the intention

src/containers/Tools.vue
noschiff: grid setup… i could just do individual classes for each tool card but i wanted to keep it more general. i'll prbly use individual classes to be cleaner

src/components/Tools/AdvisorCard.vue
noschiff: this was there before, but sure i'll change it

src/components/Tools/AdvisorCard.vue
zachary0kent: Yeah I mean if you remove the check it'll have the same effect

noschiff: oh really? ok cool

src/containers/Tools.vue
noschiff: desired is sorta the preferred. i did this back in the spring. i wanted the tool card framework to be reusable, so i chose to have this as a prop. i could've just had the size grow from the tool card's content, but it would be hard to enforce a preferred width for the overall card.

noschiff: currently, i allow some cards to grow when the screen is resized (not sure if i like this tho). that behavior is really possible because the size of the tool card is regulated from the top down rather than the inner content growing the tool card to be as big as it wants.

src/containers/Tools.vue
```diff
@@ -56,13 +72,52 @@ export default {
...
+
+    &-export {
+      grid-column: 2/3;
+      grid-row: 1/2;
+    }
+
+    &-links {
+      grid-column: 2/3;
+      grid-row: 2/3;
+    }
```
zachary0kent, line 107: I'm not a huge fan of how the styling is coupled to the number of toolcards.

noschiff: I'm not either. I initially tried to do without it, but it just isn't possible to get the design we want for the page AND nicely handle a smaller screen without it. I tried nested flexes and one big flex and it really didn't turn out nicely.

zachary0kent (approved):

benjamin-shen:
@noschiff I think if the screen width is decreased so that it's a single column, it looks worse to enforce the desired width.
image

As a side note, I know this PR isn't responsible but the tools page doesn't look great on mobile. I think there should be a dedicated effort to make Courseplan mobile-friendly (and maybe remove some functionality). It feels pretty unpolished like this. But happy to discuss this during our next meeting.
image

Issue #688, opened 2022-04-24
Malformed Data in Courses JSON


noschiff:
Our [script](https://github.com/cornell-dti/course-plan/blob/master/src/requirements/courses-json-generator.ts) to populate [full-courses.json](https://github.com/cornell-dti/course-plan/blob/master/src/assets/courses/full-courses.json) takes the data as is from the API, and this can lead to issues when our code tries to use these courses.

First, many courses don't have data in the way we expect it to for `catalogWhenOffered`, such as `"Fall or Spring."` or `"Fall, spring."` Second, some strings have NBSPs (Non-Breaking Space), particularly for the `catalogComments` field. NBSPs are almost never a good idea to use. At best, removing them will avoid formatting issues, and at worst, nothing would change.

I see two primary paths to take with fixing this. We can either implement "cleaning" into our [courses-json-generator.ts](https://github.com/cornell-dti/course-plan/blob/master/src/requirements/courses-json-generator.ts) script to (attempt to) fix the semesters offered and replace NBSPs with spaces, or we can improve our code to verify semesters to accept "weirder" data. I strongly prefer the former, at least for the NBSPs, because this will allow any code that uses the semesters offered to be free to do its job without checking for irregularities. We can also create a script to go through [full-courses.json](https://github.com/cornell-dti/course-plan/blob/master/src/assets/courses/full-courses.json) afterwards to clean the data, but it seems way simplier to just do it as we get it from the API.

One issue with trying to fix the semesters in the JSON is that there are some values that we can't handle yet, such as 7-week courses. This is good info to have for the user (I believe its used in the bottom bar?), so I wouldn't want to turn `Fall (weeks 1-7)` into `Fall`, but the additional information about the weeks messes with our semester validation. I think we could solve this by storing a field of just the season in [full-courses.json](https://github.com/cornell-dti/course-plan/blob/master/src/assets/courses/full-courses.json) while keeping the current `catalogWhenOffered` field for the user to see. That way, we can provide useful information without having to constantly handle bad data throughout the code. We should, of course, clean the `catalogWhenOffered` as best as we can to fix capitalzation and overall make it adhere to a standard.

If we make this change, we need to be careful to deal with courses already in a user's plan. It seems that courses already in a user's plan won't show the updated seasons after being changed in [full-courses.json](https://github.com/cornell-dti/course-plan/blob/master/src/assets/courses/full-courses.json).

Before fixing the lowercase s in spring:
[...]

PR #720, opened 2022-10-16
Clean Data in Course Generator Script

noschiff:
### Summary 

When running the course generator script for SP23, I noticed that we were running into errors using the data because a course was missing a value for `titleLong`, and we just put a null in the JSON. This is a very quick fix (and does not fix the issues in #688), but it should allow us to run the script and update data for SP23.

### Test Plan 

Run the script on SP23 semester and verify there are no nulls and no NBSPs in the JSON. Also verify that the script does not include or exclude extra information.

### DIFF
--- a/src/requirements/courses-json-generator.ts
+++ b/src/requirements/courses-json-generator.ts
@@ -37,9 +37,8 @@ const cleanString = (value: string | null) => {
   return value.replace(/\u00a0/g, ' ');
 };
 
-const cleanMaybeString = (value: string | null | undefined): string | undefined => {
-  return value?.replace(/\u00a0/g, ' ') || undefined;
-};
+const cleanMaybeString = (value: string | null | undefined): string | undefined =>
+  value?.replace(/\u00a0/g, ' ') || undefined;
 
 /** Throws away course object fields we don't need. Used for generate small-enough course json. */
 const courseFieldFilter = ({
--- a/src/requirements/courses-json-generator.ts
+++ b/src/requirements/courses-json-generator.ts
@@ -32,13 +32,7 @@ const getSubjects = async (semester: string): Promise<readonly string[]> => {
   }
 };
 
-const cleanString = (value: string | null) => {
-  if (value === null) return '';
-  return value.replace(/\u00a0/g, ' ');
-};
-
-const cleanMaybeString = (value: string | null | undefined): string | undefined =>
-  value?.replace(/\u00a0/g, ' ') || undefined;
+const cleanString = (value: string | null | undefined) => value?.replace(/\u00a0/g, ' ');
 
 /** Throws away course object fields we don't need. Used for generate small-enough course json. */
 const courseFieldFilter = ({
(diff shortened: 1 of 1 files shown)


src/requirements/courses-json-generator.ts
```diff
@@ -32,6 +32,14 @@ const getSubjects = async (semester: string): Promise<readonly string[]> => {
   }
 };
 
+const cleanString = (value: string | null) => {
+  if (value === null) return '';
+  return value.replace(/\u00a0/g, ' ');
```
benjamin-shen, line 37: Can you refactor `/\u00a0/g` as a constant?

src/requirements/courses-json-generator.ts
```diff
@@ -32,6 +32,14 @@ const getSubjects = async (semester: string): Promise<readonly string[]> => {
...
+};
+
+const cleanMaybeString = (value: string | null | undefined): string | undefined =>
+  value?.replace(/\u00a0/g, ' ') || undefined;
```
benjamin-shen, line 41: This is kind of confusing -- I think the only difference between including and not including `|| undefined` is for when value is `''`, in which case this function returns `undefined` rather than `''`. I don't know if that's the behavior we want. Let me know if you agree with the approach I suggested; then we can remove this weird detail.

benjamin-shen (commented):
I would prefer a single function `cleanField` ("cleanString" is a redundant name) that is implemented like `value?.replace(/\u00a0/g, ' ')`. Then, replace all current invocations of `cleanMaybeString` with `cleanField` and replace all current invocations of `cleanString` with `cleanField(...) || ''`.

src/requirements/courses-json-generator.ts
noschiff: sure

src/requirements/courses-json-generator.ts
noschiff: I only really looked at this briefly, but if we turn all undefineds into '', then we include extra information in the JSON (lots of empty strings). if we return undefined, then the field seems to be excluded from the JSON, which saves space. I'm just trying to preserve the current behavior… there's a lot that could be rethought with this though.

benjamin-shen: Are empty strings returned from the api? In that case something like this kind of makes sense

noschiff: the API just doesn't return anything, I think. I'm only making the longTitle into an empty string because our code expects it to be non null. (also it's very weird that longTitle came in as null but other stuff is just excluded and I guess left as undefined… what's the difference between excluded / not present and undefined?)

benjamin-shen: Probably just a buggy or undefined behavior in the api. Honestly I'm inclined to just keep it as `''` if it comes in as `''`.

noschiff: I don't think it does come in as `''`… I think there just isn't any information for certain stuff (like catalogComments or whatever). We really shouldn't add a '' for those because it makes the JSON bigger.

benjamin-shen: If there is no data (i.e. the field is null or undefined), then it is redundant to include `|| undefined`

benjamin-shen (approved):

noschiff:
@benjamin-shen Now we have empty strings for fields that didn't exist before

image

benjamin-shen:
@noschiff hmm are they coming in as empty strings?

PR #723, opened 2022-10-18
Semesters singleton migration

andxu282:
### Summary 

Begins the migration script to move semesters to a list of semesters by writing to a new field called plans that has a list of a list of semesters for each user.
- [x] Added field called plans and makes it so that it is written to whenever semesters is written to 
- [x] Read from plans whenever reading from semesters

Remaining TODOs:

- [ ] Run migration script in dev/prod
- [ ] Remove semester field
- [ ] Run remove script in dev/prod

PR #724, opened 2022-10-19
Update scripts

benjamin-shen:
### Summary 

This pull request cleans up the script locations and refactors a bit of code that used to live in `src/`, `src/requirements/`, and `src/requirements/admin` when they shouldn't have been. Some other miscellaneous cleanup tasks are listed below.

- [x] removed references to `requirement-generator-dist/` since that was an artifact from a long time ago
- [x] removed `req-graph-dump` script since we have the requirement debugger now
- [x] removed `firebase-admin-config` (essentially moved it to `scripts/firebase-config`), merged `firebase-config-common` and `firebase-frontend-config` to `src/firebase-config`
- [x] moved some Firebase types to `user-data.d.ts` for consistency
- [x] updated `tsconfig.node.json` according to the [tsconfig reference](https://www.typescriptlang.org/tsconfig); it now uses [node14.json base config](https://github.com/tsconfig/bases/blob/main/bases/node14.json)
- [x] added 'active-users' metric to `track-users` script

Remaining TODOs:

- [ ] refactor the migration scripts (but putting this off since it affects open pr #723, and usually migration scripts are one-off so it really only matters for future scripts)

### Test Plan 

Double check your favorite scripts work :) -- I have not tested any scripts that affect prod. Please lmk if the logic looks off somewhere (especially `scripts/firebase-config`)

Hopefully `track-users.ts` will run in prod without any issues 🤞 We will need to test that the 'active-users' metric looks right.

### Notes 

One important detail of moving `track-users.ts` from `src/` to `scripts/` is that we lose the types defined in `user-data.d.ts`. However, imo these scripts should not rely on any assumptions on types since there is no standard schema in Firestore. This should hopefully remove the weird quirk of having to update types with optional fields when doing a schema migration script.

For the 'active-users' metric, I used the condition "has at least one new semester in their schedule". I believe there is currently a bug in figuring out whether or not a semester has passed. This can be fixed in a later PR.

### Breaking Changes  

All current (WIP) and new scripts should go in the `scripts/` directory. FYI @andxu282


package.json
```diff
@@ -10,8 +10,7 @@
     "ts-node": "ts-node -T -P tsconfig.node.json",
     "courses-gen": "npm run ts-node -- src/requirements/courses-json-generator.ts",
     "req-gen": "npm run ts-node -- src/requirements/requirement-json-generator.ts",
-    "req-graph-dump": "npm run ts-node -- src/requirements/admin/requirement-graph-dumper.ts",
-    "track-users": "npm run ts-node -- src/requirements/admin/track-users.ts",
+    "track-users": "npm run ts-node -- scripts/track-user-analytics.ts",
```
noschiff, line 13: as long as this works (we can run it with npm), then I fully expect the track-users workflow to continue working, so we should be good.

scripts/firebase-config.ts
```diff
@@ -0,0 +1,58 @@
...
+
+function getDatabase(serviceAccount: string, databaseURL: string, appName?: string) {
+  return () => {
+    const credential = cert(serviceAccount);
+    const app = initializeApp(
+      {
+        credential,
+        databaseURL,
+      },
+      appName
```
noschiff, line 19: so glad you remembered to include this… I initially didn't have a name when I wrote the copy users script and I ran into issues when we tried having two apps initialized at the same time :)

scripts/firebase-config.ts
```diff
@@ -0,0 +1,58 @@
...
+    );
+    return getFirestore(app);
+  };
+}
+
+export const getProdDb = getDatabase(SERVICE_ACCOUNT_PROD, DATABASE_URL_PROD, 'prod');
+export const getDevDb = getDatabase(SERVICE_ACCOUNT_DEV, DATABASE_URL_DEV, 'dev');
+
+const isProd = process.env.PROD === 'true';
```
noschiff, line 28: it seems like we're sorta mixing firebase-admin stuff for running locally with running this automatically and checking process.env. It might be helpful to document when some of this can be used, because something like my copy user data script (and I assume a migration script) will probably be run locally and most likely own't have anything in process. it looks like the top part of the file with getDatabase deals with the local stuff and the rest is for a process, but I'm worried it could be confusing having this all in one file.

package.json
benjamin-shen: LOL whoops. I just tested it and it broke since I actually renamed it back to `track-users.ts`. Should work now

scripts/track-users.ts
```diff
@@ -24,7 +24,7 @@ function seasonToMonth(season: string) {
 
 // true if a semester is an "old semester," i.e. if a semester on CoursePlan has already passed in real life
 // the idea is that old semesters and new semesters represent whether users are planning in the future or just uploading courses
-function isOld(semester: FirestoreSemester) {
+function isOld(semester) {
```
noschiff, line 27: I don't love the implict `any` type here, we end up getting a lot of **TS7006: Parameter 'xxx' implicitly has an 'any' type**. If we're going to use implicit any, could we add `"noImplicitAny": false` to tsconfig.json? I checked, and doing this only causes minor type errors elsewhere that are easily fixed by replacing `never[]` with a real type in a total of 30 seconds.

benjamin-shen: I'm not seeing that warning, how are you seeing it? Do you have linter settings specific to your IDE?

I think `"noImplicitAny": true` is very useful, and we shouldn't use implicit any. We could add types if there's a warning

zachary0kent (approved):
lgtm after our discussion during dev meeting

src/user-data.d.ts
```diff
@@ -28,18 +28,18 @@ type FirestoreSemesterPlaceholder = {
...
 type FirestoreSemester = {
   readonly year: number;
   readonly season: FirestoreSemesterSeason;
   readonly courses: readonly (FirestoreSemesterCourse | FirestoreSemesterPlaceholder)[];
 };
 
+type FirestoreSemestersData = {
+  readonly semesters: readonly FirestoreSemester[];
+  readonly orderByNewest: boolean;
+};
```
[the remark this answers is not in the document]
benjamin-shen: Is there a difference / benefit? The `readonly` fields is used in most places throughout the codebase so I'm going to keep it. If there's reason to change, we can do that in a separate PR.

scripts/track-users.ts
noschiff: I was getting errors from typescript in Webstorm.

benjamin-shen:
The `track-users` scheduled function ran last night without issues. The active users data on https://courseplan.io/analytics looks like this:
image

In the future, there may be ways to make the value more accurate. For example, we could include a "last seen" timestamp and filter by users that were last seen online in the past 6 months.

PR #726, opened 2022-10-21
Tools Page Formatting & Useful Links

noschiff:
### Summary 

This PR changes how tool card sizes are specified by moving it to the Tools component CSS. This PR also adds the Useful Links tool card and changes the behavior for changing the screen width. (As a bonus, it makes some tweaks to the Advisors Card for when the screen width is too small).

TODO: No advisors case

### Test Plan 

1. Run `GK.enableTools()` in the javascript console.
2. Add the CS major to see some advisors data.
3. Look at Tools Page
4. Verify that the links in Useful Links work and are correct. (What do you think about the links' position?)
6. Resize the screen width
7. Collapse and expand cards

### Screenshots

normal
smaller

### DIFF
--- a/src/components/Tools/UsefulLinks.vue
+++ b/src/components/Tools/UsefulLinks.vue
@@ -25,7 +25,7 @@ export default defineComponent({
           link: 'https://classes.cornell.edu/scheduler/',
         },
         {
-          title: 'Course Roster',
+          title: 'Class Roster',
           link: 'https://classes.cornell.edu/browse/',
         },
         {
--- a/src/components/Tools/UsefulLinks.vue
+++ b/src/components/Tools/UsefulLinks.vue
@@ -1,0 +1,60 @@
+<template>
+  <div class="links">
+    <div v-for="{ title, link } in list" :key="title" class="links-group">
+      <a class="links-link" :href="link">
+        {{ title }}
+        <img class="links-link-icon" src="@/assets/images/link-gray.svg" alt="external link" />
+      </a>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue';
+
+export default defineComponent({
+  computed: {
+    list(): { link: string; title: string }[] {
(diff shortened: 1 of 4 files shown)


src/components/Tools/UsefulLinks.vue
```diff
@@ -0,0 +1,68 @@
...
+        },
+        {
+          title: 'Course Catalog',
+          link: 'https://courses.cornell.edu/',
+        },
+        {
+          title: 'Cornell DTI',
+          link: 'https://www.cornelldti.org/',
+        },
+      ];
```
benjamin-shen, line 31: @michaelfarkouh @zachary0kent any idea if we will add links in the future? Or if the links will be user-specific (eg. we link CS majors to the CS homepage and CS Student Checklist)? If so, after this release it might be good to think about refactoring this data so that it's not hardcoded in a vue file.

Also, would Scheduler be a good addition to this list?

benjamin-shen (commented):
LGTM

src/components/Tools/UsefulLinks.vue
noschiff: Scheduler is definitely good to add. I wasn't really sure what links we wanted so I just picked some. We could definitely make this more dynamic in the future (if someone wants to go link fishing 🎣)

src/containers/Tools.vue
```diff
@@ -78,45 +66,40 @@ export default {
...
+      justify-content: flex-start;
     }
   }
 
   &-card {
     &-progress {
       grid-column: 1/2;
       grid-row: 1/2;
+      width: 600px;
+      height: min-content;
```
noschiff, line 81: do we think the height is uniform behavior for all cards to factor it out?

benjamin-shen: imo it's fine, if it turns out not fine then we can un-factor it later

zachary0kent (approved):

PR #727, opened 2022-10-21
Update Courses for Winter and Spring 2023

noschiff:
### Summary 

This PR updates our list of courses and requirements JSON with offerings from Winter 2023 and Spring 2023.

- [x] ran `npm run courses-gen`
- [x] ran `npm run req-gen`

### Test Plan 

Verify that the site works and has updated course data. Check that a new course such as **AEP 3100** can be added and that the bottom bar shows up with expected information for it.

### Notes 

Total Running Time for `courses-gen`: 682096ms.

PR #728, opened 2022-10-21
Clean up workflows, add update-courses workflow

benjamin-shen:
### Summary 

This pull request cleans up the Github workflows and adds a new update-courses workflow (the implementation is currently incomplete).

- [x] Added `update-courses.yml`, which runs `npm run courses-gen` then `npm run req-gen`. This runs on `workflow_dispatch` (manually triggered). I believe it can only be tested after merging to the main branch.
- [x] Moved the pr snapshot deployment from `ci-build.yml` to `cd-snapshot.yml`
- [x] Renamed `cd.yml` to `cd-staging.yml` and renamed `release.yml` to `cd-production.yml`
- [x] Moved `check-req-json` from `ci-check.yml` to `ci-build.yml` so it runs on PRs instead of pushes
- [x] Added protection logic for `release` branch in workflows
- [x] Replaced all (improper) usages of `actions/checkout@master` with `actions/checkout@v2`
- [x] Updated names all around

Remaining TODOs:

- [ ] test `update-courses` workflow manually after merge

### Test Plan 

CI/CD


.github/workflows/cd-snapshot.yml
```diff
@@ -0,0 +1,28 @@
+name: Deploy PR Snapshot
```
benjamin-shen, line 1: Copied this deploy job from `ci-build.yml`, everything else is the same.

.github/workflows/ci-build.yml
```diff
@@ -2,26 +2,31 @@ name: CI Build
 on: pull_request
 
 jobs:
-  build:
+  check-lockfile:
```
benjamin-shen, line 5: Removed the deploy job from this file, copied check-lockfile and check-req-json jobs from `ci-check.yml`

.github/workflows/ci-check.yml
```diff
@@ -17,43 +17,15 @@ jobs:
         run: npm run format:check
       - name: Run Linter
         run: npm run lint
-      - name: Type Check
+      - name: Run Type Check
         run: npm run type-check
-      - name: Jest Test
+      - name: Run Unit Tests
         run: npm run test
-  check-lockfile:
```
benjamin-shen, line 24: Copied check-lockfile and check-req-json jobs to `ci-build.yml`, so it runs on PRs instead of pulls

.github/workflows/update-courses.yml
```diff
@@ -0,0 +1,35 @@
...
+          cache: 'npm'
+      - name: NPM Clean Install
+        run: npm ci
+      - name: Generate Courses Json
+        run: npm run courses-gen
+      - name: Generate Requirements Json
+        run: npm run req-gen
+      - name: Create Pull Request
+        id: cpr
+        uses: peter-evans/create-pull-request@v4
```
benjamin-shen, line 27: Used [create-pull-request](github.com/marketplace/actions/create-pull-request) github action. Please review this file!

zachary0kent:
Is there a reason we wouldn't want `update-courses` to be scheduled to run periodically?

benjamin-shen:
@zachary0kent I thought about that and I think we would want more granular control over when the workflow is run. It's possible that the new course roster is released at different times for different semesters, so there's no consistent schedule we would be able to follow. It's also usually top-of-mind anyway during pre-enroll season, so I don't think there's a huge benefit to scheduling it (but maybe it should be documented somewhere).

.github/workflows/cd-production.yml
```diff
@@ -3,12 +3,13 @@ on:
   push:
     branches:
       - beta-release
+      - release
```
benjamin-shen, line 6: I asked dev leads to add branch protection for the `release` branch (most importantly, require PRs). Along with `enforce-release-workflow` in `ci-policies.yml`, that should ensure only PRs from `master` can be pushed to the `release` branch. Then, we can make the switch and stop using the `beta-release` branch for prod deployments 😄

noschiff: WOOHOO!

benjamin-shen: Branch protection is now extended to all branches with suffix `release`

zachary0kent:
Could you fork CoursePlan and verify that the workflows work as expected?

zachary0kent (approved):

noschiff:
> It just occurred to me that we also need to add the semesters to `rosters.json`, eg. https://github.com/cornell-dti/course-plan/pull/727/files. I think it might be better to write an actual script for this. I'll simplify the workflow from this pr (just so we can test that run on workflow dispatch and the create PR action both work)

@benjamin-shen isn't that done automatically by the script?

benjamin-shen:
@noschiff oh yeah.... in that case I'll still update the script in a separate step because I do want to test the PR functionality

noschiff:
> @noschiff oh yeah.... in that case I'll still update the script in a separate step because I do want to test the PR functionality

what are you changing? i have a whole list of improvements i want to make in the future lol

PR #729, opened 2022-10-21
[easy] Fix requirements json from #727

benjamin-shen:
### Summary 

This pull request fixes the bad requirements json since #727 branched off an outdated source.

See the note in #628. Luckily it was easily caught since subsequent checks will fail (including when pushing to master). According to [this comment](https://github.com/actions/checkout/issues/15#issuecomment-524093065), changing the job so it runs on PR instead of pulls might fix this issue.

### Test Plan 

CI/CD


noschiff (approved):
Thanks Ben! We should definitely look into running the checks on what the code will be after we've merged the PR. The only downside is that this could make it a little confusing to understand why a check failed since it'd be run on code not in the branch if the branch is behind master.

PR #730, opened 2022-10-21
Function to generate user's course PDF

vaishnavi17:
### Summary 

In this PR, I write a typescript function `genPDF` to generate a PDF of the user's courses, grouped by semester. This function also directly downloads the file, so we don't need to write any additional code to download the PDF to their file system while connecting this to the frontend.

To test, I currently added a new GateKeep on the side-bar. If you click on that button, the PDF gets downloaded. In the next PR, I'll add the PDF generation toolcard on the toolcards page, after which the gatekeep could be removed.

Main features tested:
Sample PDF:
[Vaishnavi's CoursePlan-12.pdf](https://github.com/cornell-dti/course-plan/files/9834936/Vaishnavi.s.CoursePlan-12.pdf)

- [ ] Semesters sorted in chronological order
- [ ] Empty semesters displayed iff they are followed by at least one non-empty semester
Screenshot 2022-10-20 at 8 47 35 PM

- [ ] Page breaks display well for different types of tables
Screenshot 2022-10-20 at 8 47 35 PM Screenshot 2022-10-20 at 9 11 40 PM

- [ ] Different combinations of majors/minors/grad programs display correctly
Screenshot 2022-10-21 at 11 54 14 AM
Screenshot 2022-10-21 at 11 54 58 AM
Screenshot 2022-10-21 at 11 55 04 AM

- [ ] Multi-line course names and requirements display correctly (refer to screenshots below)
- [ ] Colour key shows up well (refer to screenshots above)
- [ ] "Downloaded from courseplan.io" and total credits count showed up at the footer of the last page (refer to attached pdf)
- [ ] Courses that don't fulfil any requirement look fine (refer to attached pdf)
- [ ] Every course on CP shows up in the PDF.. this one is kinda obvious and I don't have any screenshots to show that I tested it. But I did.
- [ ] AP/IB courses: Courses show up at the end of all the semesters in a separate table. Checked to see that exams that do/don't fulfil requirements both display correctly.

Screenshot 2022-11-01 at 7 55 37 PM
Example pdf file with AP/IB:

[Vaishnavi_CoursePlan-17.pdf](https://github.com/cornell-dti/course-plan/files/9915057/Vaishnavi_CoursePlan-17.pdf)

Some design notes:
I tried to put a lot of the constants up at the beginning of the function, like row height, font size, bubble colour codes, etc, so they can be easily modified if we want to. It is a bit hard to make everything super modular since PDF generation libraries are.. not the best.

The name of the PDF is `[first name]'s CoursePlan.pdf`. Edit: It is now `[first name]_CoursePlan.pdf` to avoid spaces and special characters as a best practice.

Remaining TODOs:

I will do these in a future PR.
- [ ] Currently, the swim test and PE requirement are grouped under the college requirement (for whatever college the user is in). For example, for me, the bubble says A&S credits). Kehui did define a colour code for "other courses", but currently it is going un-used due to this issue. IMO this is not a big deal and can be updated any time later - this issue occurs since the frontend displays the PE and swim requirements under the "College" tab.
- [ ] Add total credits count at the end of last page

### Test Plan 
refer to the sample PDF attached and other screenshots

### Blockers 

### Breaking Changes  


zachary0kent:
Wow, great work! Can you address the failing checks? I'll review after those are resolved.

noschiff:
I got this error in the console after enabling the gatekeep and clicking the "get PDF" button:
error

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+    url: 'https://courseplan.io',
+  });
+  doc.setDrawColor(0);
+  doc.line(
+    footerX + doc.getTextWidth('downloaded from '),
+    footerY + 3,
+    footerX + doc.getTextWidth('downloaded from courseplan.io'),
+    footerY + 3
+  );
+  const pdfName = `${store.state.userName.firstName}'s CoursePlan`;
```
noschiff, line 254: Can we remove the `'s` in the name? I think special characters might make the file name a little weird. For example, the file you link in your PR downloads as `Vaishnavi.s.CoursePlan-12`.

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  for (let i = 0; i < sems.length; i += 1) {
+    if (sems[i].courses.length > 0) maxNonemptyIndex = i;
+  }
+  return sems.slice(0, maxNonemptyIndex + 1);
+};
+
+const getFulfilledReqs = (
+  course: FirestoreSemesterCourse
+): readonly [string[], string[], string[]] => {
+  let reqsFulfilled = store.state.safeRequirementFulfillmentGraph.getConnectedRequirementsFromCourse(
```
benjamin-shen, line 285: I think `const` is always preferred over `let`, if possible. So in this example, it would be better to just use a new variable instead of reassigning the variable a mutated value.

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  const pdfName = `${store.state.userName.firstName}'s CoursePlan`;
+  doc.save(pdfName);
+};
+
+const getCourseRows = (sem: FirestoreSemester): [string[][], string[][], string[][]] => {
+  const filteredCourses = sem.courses.filter(
+    (course): course is FirestoreSemesterCourse => !isPlaceholderCourse(course)
+  );
+  const rows = filteredCourses.map(course => {
+    const [reqs, groups, colours] = getFulfilledReqs(course);
```
benjamin-shen, line 263: Same here, prefer `colors` spelling

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  course: FirestoreSemesterCourse
+): readonly [string[], string[], string[]] => {
+  let reqsFulfilled = store.state.safeRequirementFulfillmentGraph.getConnectedRequirementsFromCourse(
+    { uniqueId: course.uniqueID }
+  );
+  reqsFulfilled = reqsFulfilled.filter(
+    req => !reqsToFilterOut.includes(store.state.userRequirementsMap[req].name)
+  );
+
+  const bubbleTextMap: Record<RequirementGroupType, (req: string) => string> = {
```
benjamin-shen, line 292: Can this just be a function with a switch statement? One benefit would be that you can group together major and minor

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+
+const estimateTableHeight = (body: string[][]): number => {
+  const courseCharPerLine = 50;
+
+  let height = rowHeight * 2;
+  body.forEach(row => {
+    const numberOfLines = Math.max(row[0].length / courseCharPerLine, row[2].split('\n').length);
+    height += numberOfLines * (rowFontSize + 4);
+  });
+  return height;
```
benjamin-shen, line 325: This might be cleaner with a `.reduce()` but just a nit

src/requirements/data/index.ts
```diff
@@ -77,41 +77,50 @@ const json: RequirementsJson = {
...
     },
     AR: {
       name: 'Architecture, Art and Planning',
       requirements: aapRequirements,
+      abbr: 'aap',
     },
     AS1: {
       name: 'Arts and Sciences [before Fall 2020]',
       requirements: casPreFA2020Requirements,
+      abbr: 'a&s',
```
benjamin-shen, line 97: Can you change this to `cas`?

src/requirements/data/index.ts
```diff
@@ -428,5 +437,6 @@ const json: RequirementsJson = {
 
 export const colleges = ['AG', 'AR', 'AS1', 'AS2', 'EN', 'HE', 'IL', 'BU'] as const;
 export type College = typeof colleges[number];
-
 export default json;
+
+export const getCollegeAbbr = (collegeCode: string) => json.college[collegeCode].abbr;
```
benjamin-shen, line 442: Will this throw an exception if `collegeCode` is not a key of `json.college`? I think it'd be good practice to add in little redundancy checks, for example
```suggestion
export const getCollegeAbbr = (collegeCode: string) => json.college[collegeCode]?.abbr ?? 'other';
```

src/export-plan/add-fonts.ts
```diff
@@ -0,0 +1,16 @@
...
...oAGwAcAB0AHgAfACAAIQAiACMAJAAlACYAJwAoACkAKgArACwALQAuAC8AMAAxADIAMwA0ADUANgA3ADgAOQA6ADsAPAA9AD4APwBAAEEAQgBEAEUARgBHAEgASQBKAEsATABNAE4ATwBQAFEAUgBTAFQAVQBWAFcAWABZAFoAWwBcAF0AXgBfAGAAYQEEAKMAhACFAL0AlgDoAIYAjgCLAJ0AqQCkAQUAigCDAJMBBgEHAI0AiADeAQgAngCqAPUA9AD2AKIArQDJAMcArgBiAGMAkABkAMsAZQDIAMoAzwDMAM0AzgDpAGYA0wDQANEArwBnAPAAkQDWANQA1QBoAOsA7QCJAGoAaQBrAG0AbABuAKAAbwBxAHAAcgBzAHUAdAB2AHcA6gB4AHoAeQB7AH0AfAC4AKEAfwB+AIAAgQDsAO4AugCwALEAuwDYANkBCQEKAQsBDAENAQ4BDwEQAREBEgETARQBFQEWALIAswC2ALcAxAC0ALUAxQCHAKsBFwC+AL8BGAEZAIwBGgEbARwBHQEeBmdseXBoMQZnbHlwaDIHdW5pMDBBMAd1bmkwMEFEB3VuaTAwQjIHdW5pMDBCMwd1bmkwMEI5B3VuaTIwMDAHdW5pMjAwMQd1bmkyMDAyB3VuaTIwMDMHdW5pMjAwNAd1bmkyMDA1B3VuaTIwMDYHdW5pMjAwNwd1bmkyMDA4B3VuaTIwMDkHdW5pMjAwQQd1bmkyMDEwB3VuaTIwMTEKZmlndXJlZGFzaAd1bmkyMDJGB3VuaTIwNUYERXVybwd1bmlFMDAwB3VuaUZCMDEHdW5pRkIwMgd1bmlGQjAzB3VuaUZCMDS4Af+FsAGNAEuwCFBYsQEBjlmxRgYrWCGwEFlLsBRSWCGwgFkdsAYrXFgAsAMgRbADK0SwCCBFsgNKAiuwAytEsAcgRbIIPQIrsAMrRLAGIEWyBzACK7ADK0SwBSBFsgYiAiuwAytEsAQgRboABQEUAAIrsAMrRLAJIEWyAxQCK7ADK0QBsAogRbADK0SwCyBFsgpJAiuxA0Z2K0SwDCBFsgtqAiuxA0Z2K0SwDSBFsgxyAiuxA0Z2K0SwDiBFsg1SAiuxA0Z2K0SwDyBFsg47AiuxA0Z2K0SwECBFsg8hAiuxA0Z2K0SwESBFshAeAiuxA0Z2K0RZsBQrAAAAAU/He7gAAA==';
```
benjamin-shen, line 5: What is this font? Can it be not hard-coded in this ts file?

benjamin-shen (changes requested):
Looks so good!!! 🤩 I made an initial pass with small comments, let me know when they are resolved and I'll take a second pass. They should all be pretty low effort to fix but I'm happy to help.

I also encountered the same error as Noah.

For the failing CI pipeline, make sure to pull from master and double check that you changed all the requirement json types (`npm run type-check`). You might also need to sign the CLA.

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  doc.text(
+    `${store.state.onboardingData.gradSem} ${store.state.onboardingData.gradYear}`.trim(),
+    448,
+    93.2
+  );
+
+  doc.setTextColor(0);
+
+  let sems = sortedSemesters(store.state.semesters, false);
+  sems = trimEmptySems(sems);
```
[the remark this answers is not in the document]
noschiff: @benjamin-shen gunning for the PR co-author 👀 :)

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  return [rows.map(row => row[0]), rows.map(row => row[1]), rows.map(row => row[2])];
+};
+
+const trimEmptySems = (sems: readonly FirestoreSemester[]): readonly FirestoreSemester[] => {
+  let maxNonemptyIndex = -1;
+  for (let i = 0; i < sems.length; i += 1) {
+    if (sems[i].courses.length > 0) maxNonemptyIndex = i;
+  }
+  return sems.slice(0, maxNonemptyIndex + 1);
+};
```
zachary0kent, line 280:
```suggestion
const trimEmptySems = (sems: readonly FirestoreSemester[]): readonly FirestoreSemester[] => {
  const maxNonemptyIndex = sems.findLastIndex(sem => sem.courses.length > 0)
  return sems.slice(0, maxNonemptyIndex + 1);
};
```

src/requirements/data/index.ts
```diff
@@ -77,41 +77,50 @@ const json: RequirementsJson = {
     UNI: {
       name: 'University',
       requirements: universityRequirements,
+      abbr: 'other',
```
zachary0kent, line 80: I think `abbrev` might be clearer

src/export-plan/pdf-gen.ts
vaishnavi17: Huh, this is weird. What browser are you using? I tried with both Chrome and Safari and the quotation works for me. The link in my PR is just a file that I uploaded, so I think that might be on github's end.

Also, are you still getting that addImage error? i am not getting that on either chrome or safari.

noschiff:
@vaishnavi17 I'm still getting the error. I wonder if it has anything to do with the data in my account? You can use the [copy user data script](https://github.com/cornell-dti/course-plan/blob/master/scripts/copy-user-data.ts) to copy data from `nps39@cornell.edu` to your account if you want to look into it.

image

src/export-plan/pdf-gen.ts
noschiff: see my recent comment. I haven't actually been able to download it yet, so I'm not sure how it will be stored. However, I think using special characters in file names generally can be problematic.

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  getCollegeFullName,
+  getMajorFullName,
+  getMinorFullName,
+  sortedSemesters,
+  isPlaceholderCourse,
+  getGradFullName,
+} from '../utilities';
+import store from '../store';
+import { addFonts } from './add-fonts';
+import { getCollegeAbbr } from '@/requirements/data/index';
```
noschiff, line 13: `@` often cause a lot of issues with running typescript in the codebase for some reason
```suggestion
import { getCollegeAbbr } from '../requirements/data';
```

zachary0kent:
@vaishnavi17 Can you sign the CLA? Also I think I'm still getting the same error Noah and Ben encountered.

src/export-plan/pdf-gen.ts
noschiff: also the spaces can be annoying because they introduce `%20`

vaishnavi17: Hm, where do they introduce %20? I guess I can double check with the design folks about what file name they'd prefer. But maybe since this is meant for the end user it, I thought it'd make sense to capitalize it etc. Otherwise, I can just do `vaishnavi-course-plan.pdf`.

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+const reqsToFilterOut = ['A&S Credits'];
+
+const bubbleColourMap: Record<RequirementGroupType, (req?: string) => string> = {
+  College: (req?: string) =>
+    req && store.state.userRequirementsMap[req].sourceSpecificName === 'UNI'
+      ? '#1AA9A5'
+      : '#4F7D91',
+  Grad: () => '#1E8481',
+  Major: () => '#1E8481',
+  Minor: () => '#145351',
```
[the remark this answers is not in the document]
vaishnavi17: Hm, switch required a default for some reason, which I thought looked a bit less clean, since this is a complete set of cases for the data type. And I was getting a linter warning about it wanting a "return" at the end, even though I was returning in every case.

switch/case syntax in typescript can't be used as an expression I think, so it just makes things somewhat ugly. Like I can't do = return switch () {
...}

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  const doc = new jsPDF({ unit: 'pt', format: 'letter' });
+
+  addFonts(doc);
+  doc.setFont('ProximaNova-Regular', 'normal');
+
+  const tableX = (doc.internal.pageSize.width - tableWidth) / 2;
+
+  const img = new Image();
+  img.src = 'src/assets/cp-logo-pdf.png';
+  doc.addImage(img, 48, 30, 84, 23.25);
```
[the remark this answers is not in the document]
vaishnavi17: Yeah, I just got them from the Figma design. It would be a bit awkward to make them each a constant, since they're all just used once.

src/export-plan/pdf-gen.ts
vaishnavi17: same thing as above, the switch statement looks a bit awkward because it expects a default case, and also wants a return at the end of the switch statement?

noschiff (changes requested):
Can we modify the [requirements JSON generator](https://github.com/cornell-dti/course-plan/blob/vaish/pdf-gen/src/requirements/requirement-json-generator.ts) to exclude the `abbr`? I don't think it should be included.

src/requirements/data/index.ts
vaishnavi17: :( i do like a&s but since that's the general consensus I have changed it to cas

src/export-plan/add-fonts.ts
vaishnavi17: Yeah unfortunately this is how you add fonts in jspdf.... its the proxima nova one thats used elsewhere as well, but I need to add it as the base 64 version, and I need to use the ttf file version instead of the otf files that are in the assets folder already.

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+  );
+
+  doc.setTextColor(0);
+
+  let sems = sortedSemesters(store.state.semesters, false);
+  sems = trimEmptySems(sems);
+  const tableHeader = [['Course', 'Credits', 'Requirements Fulfilled']];
+  let startct = Math.max(firstTableY, programY + 20);
+
+  for (const sem of sems) {
```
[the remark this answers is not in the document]
vaishnavi17: Hmm I don't think so? Which parts are you suggesting can be pulled out?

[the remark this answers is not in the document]
zachary0kent: Yeah idk after looking at it more i'm not sure it's feasible

src/export-plan/pdf-gen.ts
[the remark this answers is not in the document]
benjamin-shen: I think a switch statement is still preferred. In other areas of the codebase we either throw an error or return a value that is never used.

src/export-plan/pdf-gen.ts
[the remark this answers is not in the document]
benjamin-shen: Would it be possible to make these numbers cleaner (eg. round to the nearest 5) and then refactor? I guess in a future PR

src/export-plan/add-fonts.ts
benjamin-shen: Hmmm can it be read from a file or something?

src/export-plan/pdf-generator.ts
```diff
@@ -0,0 +1,361 @@
...
+      return [
+        [`${course.code}: ${course.name}`, course.credits.toString(), reqs.join('\n')],
+        bubbles,
+      ];
+    });
+  return { body: rows.map(row => row[0]), bubbles: rows.map(row => row[1]) };
+};
+
+const trimEmptySems = (sems: readonly FirestoreSemester[]): readonly FirestoreSemester[] => {
+  if (sems.length === 0) return [];
```
zachary0kent, line 310: Is this check needed?

zachary0kent (commented):
Can we remove the empty file? Also, I generally think we should use objects wherever we're using tuples--I think it makes the code more self-documenting.

src/export-plan/pdf-gen.ts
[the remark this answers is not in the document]
vaishnavi17: even if I do round them to the nearest 5, this is just the position attribute for the image, so nothing else will be at the same coordinates as it. So I'm not sure if it makes sense to make these constants - it might even make the code less readable.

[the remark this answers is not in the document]
benjamin-shen: Are there other components that are positioned relative to this one? Or are all the major components placed absolutely?

I think the more important side effect of pulling out magic numbers (even to local variables) is to give them some semantic context, not just to refactor multiple instances of the constant. I think having a bunch of magic numbers and absolute positioning in this file makes it more difficult to maintain by other devs (eg. if we change the design, it might be the same LOE to write it mostly from scratch vs finding the right numbers to update).

src/export-plan/pdf-gen.ts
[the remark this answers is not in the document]
vaishnavi17: Hey, I switched the other map to switch/case but I think it is cleanest to keep this one like this. It's used in two places, and for one use case we have a requirement with a requirement group, and for the other case we just want to get the colour for a group like "grad", and this format works best.

src/export-plan/add-fonts.ts
vaishnavi17: i guess this is the file for the font? I don't see any point in making it async.

src/export-plan/pdf-gen.ts
```diff
@@ -0,0 +1,328 @@
...
+    footerX + doc.getTextWidth('downloaded from '),
+    footerY + 3,
+    footerX + doc.getTextWidth('downloaded from courseplan.io'),
+    footerY + 3
+  );
+  const pdfName = `${store.state.userName.firstName}'s CoursePlan`;
+  doc.save(pdfName);
+};
+
+const getCourseRows = (sem: FirestoreSemester): [string[][], string[][], string[][]] => {
```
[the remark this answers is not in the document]
vaishnavi17: done! thanks

src/requirements/data/index.ts
vaishnavi17: done

src/export-plan/pdf-generator.ts
vaishnavi17: If length was 0, it'd return sems.slice(0, 0), and I wasn't sure what the behaviour of that was on an empty list, since technically the index 0 doesn't exist, but also 0 >= 0, so maybe it would never even check that. let me know if you'd like me to remove it.

src/export-plan/pdf-gen.ts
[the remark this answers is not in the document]
benjamin-shen: alright

noschiff:
merging in master should fix the conflicts

noschiff:
also since the PR is also connecting to the front-end now (and the pdf generation is done client side), do you think it would be better to remove the "backend" from the PR name?

src/export-plan/pdf-generator.ts
benjamin-shen: You can always test it in the console! `[].slice(0, n)` seems to return `[]` for any `n`, so it's safe to remove.

zachary0kent: Yep--`a.slice(i, j)` is the sub array from i (inclusive) to j (exclusive).

src/export-plan/pdf-gen.ts
vaishnavi17: changed!

benjamin-shen (approved):
Yayy! Good job @vaishnavi17 and team 🎉

Could you remove "backend" in the PR title? Then it should be good to go!

PR #733, opened 2022-10-23
[easy] Rename human policy to health policy

benjamin-shen:
### Summary 

This pull request renames "human policy" to "health policy".

### Test Plan 

Human policy minor should be called health policy now
image


noschiff (approved):
How convenient that the abbreviation is still "HP"! Just tested having the Human Policy minor before this PR, and it seems to work fine for my account on this branch.

PR #734, opened 2022-10-25
Add Advisors Data

noschiff:
### Summary 

This PR adds the advisors data from @michaelfarkouh's research into the codebase. See #690 for an explanation of where this data is stored and how it is integrated into the app. Thanks @zachary0kent for helping to enter the data! This PR also fixes a misspelling in the mechanical engineering requirements data export, and it adds the front-end for no advisors.

### Test Plan 

Run `GK.enableTools()` in the console. Set your college, major, and minor, and verify that the advisors in the Advisors tool card are correct and the email links work.
For anyone looking at the code: GitHub doesn't want to show you the diff for `src/requirements/data/index.ts`, but you **should view it!**

### DIFF
--- a/src/components/Tools/AdvisorCard.vue
+++ b/src/components/Tools/AdvisorCard.vue
@@ -1,6 +1,9 @@
 <template>
   <div class="advisors">
-    <div class="advisors-content">
+    <div class="advisors-content" v-if="advisors.length === 0">
+      <span class="advisors-empty"> Sorry, you don't have an advisor yet. </span>
+    </div>
+    <div v-else class="advisors-content">
       <div class="advisors-heading-row">
         <span class="advisors-info advisors-heading-content">Type</span>
         <span class="advisors-info advisors-heading-content">Name</span>
@@ -78,6 +81,11 @@ export default defineComponent({
     flex-direction: column;
     justify-content: flex-start;
     margin: 0 1.5rem 0 1.5rem;
+  }
+
+  &-empty {
+    display: flex;
+    justify-content: space-around;
   }
 
   &-info {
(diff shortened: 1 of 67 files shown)


src/requirements/data/colleges/asFA2020.ts
```diff
@@ -100,3 +102,13 @@ const casFA2020Requirements: readonly CollegeOrMajorRequirement[] = [
 ];
 
 export default casFA2020Requirements;
+
+export const casAdvisors: AdvisorGroup = {
+  advisors: [
+    {
+      name: 'Jayla Greene',
+      email: 'jng65@cornell.edu',
+      checker: lastNameRange('https://as.cornell.edu/advising', ''),
```
benjamin-shen, line 111: Is this supposed to be a source?

noschiff: Yes, most likely! @zachary0kent I think you might be off the hook for this… I think @michaelfarkouh put this in the wrong column in the spreadsheet :)

zachary0kent:
Screen Shot 2022-10-25 at 8 59 21 PM
Do you think "college" and "major" should be title case here?

noschiff:
@zachary0kent 
> Do you think "college" and "major" should be title case here?

I don't think so… I like it lowercase.

zachary0kent (approved):
LGTM. Maybe in the future we could store advisor information in a json or something similar--would also make it easier for non-devs to add advisors.

benjamin-shen:
I agree, the only relation this has with requirements data is the mapping from college/major/minor to advisor. I think it really should be pulled out to a single file (especially since advisors also change pretty frequently).

src/components/Tools/AdvisorCard.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div class="advisors">
-    <div class="advisors-content">
+    <div class="advisors-content" v-if="advisors.length === 0">
+      <span class="advisors-empty"> Sorry, you don't have an advisor yet. </span>
```
benjamin-shen, line 4: I think you should confirm this copy @noschiff, imo the wording is pretty dev-centric and might sound strange to the user if they do have an advisor but we don't have the data.

noschiff: I took that from the design

noschiff: image
this was the design, but I'm happy to change it

PR #735, opened 2022-10-25
[easy] Write script outputs to ignored directory

benjamin-shen:
### Summary 

This pull request writes script outputs to an ignored directory (currently `scripts/out`). It also adds `serviceAccountDev.json` and `serviceAccount**` to the .gitignore.

### Test Plan 

Run the copy-user-data script and observe the output is gitignored.


noschiff (approved):
I wonder what could've inspired this PR! Looks good, but I'm going to have to make a quick PR to fix the documentation of the copy-user-data.ts script following this and #724.

PR #739, opened 2022-10-25
Update Copy User Data Script Documentation

noschiff:
### Summary 

This is a quick PR to update the documentation of [copy-user-data.ts](https://github.com/cornell-dti/course-plan/blob/db7bc6f912a6ac0f6854672b46e3b6f37343a5fa/scripts/copy-user-data.ts) to reflect that

- the dev service account should be named `serviceAccount.json` after the change in #724
- the output path will be relative to scripts/out after the change in #735 

### Test Plan 

Read the description of [copy-user-data.ts](https://github.com/cornell-dti/course-plan/blob/db7bc6f912a6ac0f6854672b46e3b6f37343a5fa/scripts/copy-user-data.ts) and ensure it makes sense and is correct.

### DIFF
--- a/scripts/copy-user-data.ts
+++ b/scripts/copy-user-data.ts
@@ -3,13 +3,17 @@
  * BE CAREFUL WHEN COPYING TO PROD!
  * Script to copy data from one user on production or dev to another user on dev.
  * Requires service accounts for database.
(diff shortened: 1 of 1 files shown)


noschiff:
@zachary0kent @benjamin-shen whoever fixes the service account, please feel free to rerun the workflow

benjamin-shen (approved):

PR #742, opened 2022-10-27
Prefer top-level await

zachary0kent:
Replaces the async `genPDF` function with top-level `await`s


zachary0kent:
Oops, I guess top-level `await` isn't compatible with firefox :(

PR #744, opened 2022-10-27
Wire PDF generation to export toolcard

zachary0kent:
### Summary 

Connects PDF generation to a basic frontend for the export toolcard and removes the pdf generation feature flags. For some reason I don't have access to the figma designs, so I just threw something together for now.

### Test Plan 

Enable tools and then click the download button.


zachary0kent:
> Do we want to merge into into the backend branch, or would it be better to merge this into master separately? Usually, I'd say isolating backend is a good idea, but the circumstances might just mean it's easier to merge this into Vaish's branch and avoid dealing with the feature flag.

We did the same thing with `user-choice`. I don't really have much of an opinion though.

zachary0kent:
> Thanks for working on the frontend!! I think overall the structure looks good to me, I just gotta change the name of genPDF to generatePDF so after those small naming changes are done you can edit them in this PR as well.
> 
> I'll test the frontend out more thoroughly by tonight and leave more feedback if I have any:)

Thanks! I don't have access to the designs though, so feel free to tinker with it to improve it.

PR #747, opened 2022-10-27
Remove Unnecessary Information from Requirements JSON

noschiff:
### Summary 

Removes the `abbr` field added to the requirements JSON.

### Notes

We don't want the abbr field to be in the requirements JSON. So, we can just remove it from the JSON generator. However, this brings up a bigger issue with our infrastructure. The data in index.ts is treated as the JSON data in terms of types. There's no way currently to include extra data in index.ts without it also being part of the JSON type. Our solution is to make the type optional so it doesn't have to appear in the JSON and we can remove it, but that isn't ideal.

Here's an example. If we want abbr to be in index.ts, we need to add it to `CollegeRequirements<R>`.
image

However, the JSON uses this type.
image

The req JSON generator returns a MutableRequirementsJSON that (I believe) becomes readonly when the data is read from the JSON. So, we need to add abbr to the MutableRequirementsJSON type so that we guarantee that the abbr is in the JSON.
image

This is how we would use an actual type. As you can see, including something in index.ts requires us to include it in the JSON, which isn't what we want. I think we should try to decouple index.ts from the JSON, so the JSON uses the index.ts data, rather than it being the data.

### Test Plan 

Run the req generator script and verify the abbr are not in the generated JSON.

### DIFF
--- a/src/requirements/requirement-json-generator.ts
+++ b/src/requirements/requirement-json-generator.ts
@@ -341,14 +341,14 @@ const generateDecoratedRequirementsJson = (): DecoratedRequirementsJson => {
       sortRequirementCourses
     );
   Object.entries(university).forEach(([universityName, universityRequirement]) => {
-    const { requirements, advisors, ...rest } = universityRequirement;
+    const { requirements, advisors, abbr, ...rest } = universityRequirement;
     decoratedJson.university[universityName] = {
       ...rest,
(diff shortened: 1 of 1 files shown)


zachary0kent (approved):

zachary0kent:
I might've waited for a review from @vaishnavi17 before merging

PR #748, opened 2022-10-27
Create Export Schedule Tool Card

noschiff:
### Summary 

This PR creates the Export Schedule tool card and adds it to the Tools page. After this PR is merged into master, we can hook up the download button to the pdf generation in #730. If we merge this PR, we should close #744.

### Test Plan 

Enable the tools page by running `GK.enableTools()` in the javascript console. Look at the Export Schedule tool card.

image


vaishnavi17:
Thanks so much for making the PDF gen frontend! I checked responsive-ness by changing browser width and the behaviour looked fine to me, and the design also looks really good! When I merge the backend PR in, I'll replace the console call with the `generatePDF()` call.

vaishnavi17 (approved):

PR #750, opened 2022-10-28
Relocate Requirement Progress Calculation

noschiff:
### Summary 

Moves requirement fulfillment progress calculation to requirement-frontend-computation.ts so that the progress tracker can access these methods.

### Test Plan 

Verify the progress bars on the sidebar still work.


benjamin-shen (approved):
LGTM

PR #751, opened 2022-10-28
Create Progress Tracker Tool Card

noschiff:
### Summary 

This PR creates the Progress Tracker tool card. The progress tracker lists your progress in all your programs on the right, and combines them all together into a blue progress bar.

Remaining TODOs:

- [ ] Fix color inconsistency between sidebar and progress tracker 

This was built with help from @andxu282, @zachary0kent, and @jerrry1123.

### Test Plan 

Run `GK.enableTools()` in the console. Add majors and minors and test!

### Notes 

We currently combine the requirements all together for the total progress, which isn't really ideal because it weights all reqs equally irrespective to how many courses or credits they correspond to. We also use safe progress only right now.


src/components/Tools/ProgressTracker.vue
```diff
@@ -0,0 +1,214 @@
...
+import {
+  groupedRequirementDangerouslyFulfilled,
+  groupedRequirementTotalDangerousRequirementProgress,
+  groupedRequirementTotalRequired,
+  groupedRequirementTotalSafeRequirementProgress,
+} from '@/requirements/requirement-frontend-computation';
+
+export default defineComponent({
+  computed: {
+    requirementProgressBundles(): GroupedRequirementFulfillmentReportWithProgress[] {
```
zachary0kent, line 43: I feel like this shouldn't be in the Vue component

noschiff: I initially had all the different properties bundled into one function, but I got worried that that would cause performance issues with the req sidebar because only parts of it are used at a time.

noschiff:
Our req sidebar ignores self check, and we use the same code. So, it looks like I finished the Applied Math major when I haven't, but I don't know if it's worth changing right now. Just something to think about.

src/requirement-types.d.ts
```diff
@@ -165,3 +165,10 @@ type GroupedRequirementFulfillmentReport = {
   readonly specific: string;
   readonly reqs: readonly RequirementFulfillment[];
 };
+
+type GroupedRequirementFulfillmentReportWithProgress = GroupedRequirementFulfillmentReport & {
+  readonly dangerouslyFulfilled: number;
+  readonly totalRequired: number;
+  readonly safeProgress: number;
+  readonly dangerousProgress: number;
+};
```
zachary0kent, line 174:
```suggestion
type GroupedRequirementFulfillmentReportWithProgress = GroupedRequirementFulfillmentReport & Readonly<{
  dangerouslyFulfilled: number;
  totalRequired: number;
  safeProgress: number;
  dangerousProgress: number;
}>;
```

zachary0kent:
Screen Shot 2022-10-28 at 7 22 10 PM
It looks kinda weird how bright the emoji is here

zachary0kent:
Screen Shot 2022-10-28 at 7 22 53 PM
The gap between the tracker and the text here is pretty big, but idk how much that matters. I don't think it looks too bad, but might be better if it were smaller.

src/requirement-types.d.ts
noschiff: you, andrew, and I all have like a 1/3 of the commits on this branch so just commit any changes you want lol

benjamin-shen (commented):
Mostly LGTM, but why are we using the dangerously fulfilled progress instead of the safely fulfilled progress? I think we should be using the safe progress, since ideally the end goal is for users to resolve conflicts and opt in/out such that there is no additional unknown information (i.e., no difference between dangerous and safe progress).

noschiff:
> > We also use safe progress only right now.
> 
> 
> 
> Mostly LGTM, but it seems like we are using dangerously fulfilled progress instead of the safely fulfilled progress. I think we should be using the safe progress, since ideally the end goal is for users to resolve conflicts and opt in/out such that there is no additional unknown information (i.e., no difference between dangerous and safe progress).

I thought we used safe, but it seems someone changed it. Wanna chime in @andxu282?

andxu282:
> > We also use safe progress only right now.
> 
> Mostly LGTM, but it seems like we are using dangerously fulfilled progress instead of the safely fulfilled progress. I think we should be using the safe progress, since ideally the end goal is for users to resolve conflicts and opt in/out such that there is no additional unknown information (i.e., no difference between dangerous and safe progress).

So the reason why I changed it to dangerouslyFulfilled is that on the Plans page, at the top here when we show the number of requirements fulfilled, we're showing them the number of dangerouslyFulfilled as opposed to the safeProgress. So, I thought it would be best if we made them match.
Capture d’écran 2022-10-30 à 10 41 55 AM

benjamin-shen:
@andxu282 That makes sense to me. Though I personally feel think we should show the safe progress for the progress tracker.

I think in your screenshot, safe fulfillment is equal to dangerous fulfillment. Here's an [example](https://github.com/cornell-dti/course-plan/pull/622#issuecomment-1001255926) of dangerous fulfillment being at 2/8, and safe fulfilment being at 1/8. It kind of makes sense to show 2/8 as the text on the reqs sidebar, but I don't necessarily want to show a user with a ton of conflicts (say 8/8 dangerous but 5/8 safe) a fully completed progress tracker -- I would rather them see 8/8 once they resolve those conflicts. Maybe we could also ask for some design input?

zachary0kent:
> @andxu282 That makes sense to me. Though I personally feel think we should show the safe progress for the progress tracker.
> 
> I think in your screenshot, safe fulfillment is equal to dangerous fulfillment. Here's an [example](https://github.com/cornell-dti/course-plan/pull/622#issuecomment-1001255926) of dangerous fulfillment being at 2/8, and safe fulfilment being at 1/8. It kind of makes sense to show 2/8 as the text on the reqs sidebar, but I don't necessarily want to show a user with a ton of conflicts (say 8/8 dangerous but 5/8 safe) a fully completed progress tracker -- I would rather them see 8/8 once they resolve those conflicts. The color also implies safe progress (whereas orange is dangerous).
> 
> Maybe we could also ask for some design input?

I don't like displaying progress inconsistently--that seems confusing to the user, especially since we haven't even rolled out conflicts yet.

noschiff:
> Screen Shot 2022-10-28 at 7 22 53 PM
> 
> The gap between the tracker and the text here is pretty big, but idk how much that matters. I don't think it looks too bad, but might be better if it were smaller.

What do you think now?

zachary0kent:
> > Screen Shot 2022-10-28 at 7 22 53 PM
> > The gap between the tracker and the text here is pretty big, but idk how much that matters. I don't think it looks too bad, but might be better if it were smaller.
> 
> What do you think now? @zachary0kent

Looks a lot better now

benjamin-shen:
> I don't like displaying progress inconsistently--that seems confusing to the user, especially since we haven't even rolled out conflicts yet. I think it would be better to update the progress tracker in the future after we do release user choice.

@zachary0kent ok. Then I think we should at least have a TODO item.

There probably will need to be some design input. Here are the two issues:
- inconsistent colors. blue should mean safe and orange should mean dangerous
  - one solution is to update the progress tracker with two colors, and show both safe and dangerous (in the same way we do in the req sidebar). I think this makes it more cluttered than necessary
  - another solution is to change the text in progress tracker (this pr) to be the safe progress. then this results in the second issue below.
- inconsistent text. we should show the user the same text in the req sidebar and the progress tracker (or at least unambiguous text)
  - one solution is to add both the safe and dangerous progress as text, if safe progress is not equal to dangerous progress. though this might be an overwhelming amount of info for the user
  - another solution is to change the text in both places to be safe progress. then we would need to change the current copy in the req bar, and it arguably touches the first issue above
  - another solution is to keep the dangerous progress as text in the progress tracker (what's currently implemented in this pr). then this results in the first issue above.

So if we don't change the progress tracker design, and we don't change the req bar progress, then there will be an inconsistency for the user.

zachary0kent:
> > I don't like displaying progress inconsistently--that seems confusing to the user, especially since we haven't even rolled out conflicts yet. I think it would be better to update the progress tracker in the future after we do release user choice.
> 
> @zachary0kent ok. Then I think we should at least have a TODO item.
> 
> There probably will need to be some design input. Here are the two issues:
> 
> * inconsistent colors. blue should mean safe and orange should mean dangerous
>   
>   * one solution is to update the progress tracker with two colors, and show both safe and dangerous (in the same way we do in the req sidebar). I think this makes it more cluttered than necessary
>   * another solution is to change the text in progress tracker (this pr) to be the safe progress. then this results in the second issue below.
> * inconsistent text. we should show the user the same text in the req sidebar and the progress tracker (or at least unambiguous text)
>   
>   * one solution is to add both the safe and dangerous progress as text, if safe progress is not equal to dangerous progress. though this might be an overwhelming amount of info for the user
>   * another solution is to change the text in both places to be safe progress. then we would need to change the current copy in the req bar, and it arguably touches the first issue above
>   * another solution is to keep the dangerous progress as text in the progress tracker (what's currently implemented in this pr). then this results in the first issue above.
> 
> So if we don't change the progress tracker design, and we don't change the req bar progress, then there will be an inconsistency for the user.

That sounds good to me. It shouldn't be too hard adding a dangerous progress bar in the future--I added a TODO item.

benjamin-shen (approved):

PR #753, opened 2022-10-29
[Easy] Add exhaustiveness check

zachary0kent:
### Summary 

Modifies `.eslintrc` to enforce exhaustiveness checks on switch statements. This allows us to safely omit unreachable default cases where we would otherwise do something like `throw new Error()`.


zachary0kent:
Couldn't get this to work very nicely...

PR #754, opened 2022-11-01
Create workflow to backup Firestore

benjamin-shen:
### Summary 

This pull request adds a workflow to back up the production Firestore (I followed [this guide](https://fireship.io/snippets/firestore-automated-backups/)).

Instructions to import: https://firebase.google.com/docs/firestore/manage-data/export-import

#### TODO
Use the gcloud [auth](github.com/marketplace/actions/authenticate-to-google-cloud) Github action. We may eventually want to move away from service account keys in the entire project ([Workload identity federation documentation](https://cloud.google.com/iam/docs/workload-identity-federation)).

### Test Plan 

Run the workflow after it gets merged to master.

### Notes 

The job is scheduled to run every 4:05am on January and July 11th.

You can also manually invoke it in the Actions tab. Backups are encouraged before running data migration scripts on prod. But please be aware of the potential cost of backing up data frequently (eg. run backups at least a day before running data migration scripts).
> Caution: Exporting data from Cloud Firestore will incur one read operation per document exported. However, these reads will not appear in the usage section of the console. Make sure you understand this before setting up recurring exports to avoid an unexpected bill.


.github/workflows/backup-firestore.yml
```diff
@@ -0,0 +1,21 @@
+name: Backup Firestore
+
+on:
+  schedule:
+    - cron:  '5 4 11 1,7 *' # 4:05am on 1/11 and 7/11
```
benjamin-shen, line 5: Why not 4:00am on the 1st or 15th?

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

> Note: The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.

noschiff:
can you document where the data gets stored and how this workflow works?

benjamin-shen:
Thanks for asking @noschiff!

The location of the data is currently in the TPM onboarding guide, since Firestore backups have been done in the past. I plan on moving a lot of this information out of the guide to the dev notion.

I think the workflow is mostly self-documenting. Here's the [action](https://github.com/marketplace/actions/set-up-gcloud-cloud-sdk-environment) I used. Here's the [reference](https://cloud.google.com/sdk/gcloud/reference/firestore/export) for `gcloud firebase export`. The workflow authenticates and sets up the specified project using the service account key and then backs up the firestore data to the specified bucket. Did you have any other questions about the workflow?

noschiff:
> Thanks for asking @noschiff!
> 
> The location of the data is currently in the TPM onboarding guide, since Firestore backups have been done in the past. I plan on moving a lot of this information out of the guide to the dev notion.
> 
> I think the workflow is mostly self-documenting. Here's the [action](https://github.com/marketplace/actions/set-up-gcloud-cloud-sdk-environment) I used. Here's the [reference](https://cloud.google.com/sdk/gcloud/reference/firestore/export) for `gcloud firebase export`. The workflow authenticates and sets up the specified project using the service account key and then backs up the firestore data to the specified bucket. Did you have any other questions about the workflow?

@benjamin-shen ah, I didn't realize that we already had documentation in the TPM onboarding doc. And thanks for the clarification on what's being run -- I didn't understand where the `gcloud firestore` came from / what it was doing.

zachary0kent (approved):

benjamin-shen:
The 2022-11-17 backup was successful! It runs pretty quickly 😀

PR #758, opened 2022-11-02
Small PDF generation bug fixes

vaishnavi17:
### Summary 
1. There was a bug where, if a user has a major like CS, adds a course that for e.g. fulfils a tech elective, and then removes the CS major, there is an error because the safe requirements graph doesn't remove the tech elective from the graph, but the tech elective no longer exists in the user requirements map. Currently, I have just made a fix to filter out all such requirements, but this should definitely also be addressed later to see what's causing the requirements to not be removed from the safe requirements graph if a major is removed.

Reason for the error:
This is the function to get the reqs fulfilled by a course a user has taken. We see that the first step is to use the saferequirementfulfillmentgraph to get all the requirements fulfilled by that course. Strangely, there is a bug that even when the CS major is removed, the graph still contains edges to the CS major requirements like "tech elective". Hence, a requirement like tech elective can be part of the list `reqsFulfilled`.

Screenshot 2022-11-02 at 7 32 06 PM

Then, to get the details like requirement name and the group it's part of (engineering, major, minor, etc,), I use this list `reqsFulfilled` in the following piece of code. The error happens because the CS requirement like "tech elective" has already been removed from the `userRequirementsMap`. It is easy to fix by just filtering out requirements that aren't in `userRequirementsMap`, but we should look into fixing the issue of edges not being removed from saferequirementfulfillmentgraph.

Screenshot 2022-11-02 at 7 32 20 PM

2. Fixed the issue where too many majors/too long majors can overflow pdf. Current policy is if a major is overflowing a line, the whole major gets shifted to the next line.

3. Added emoji for AP/IB creds 🏫

### Test Plan 
1. Tested this on the test server - file downloads fine after removing major, and only college requirements show up in the table.
2. Examples of different scenarios:
Screenshot 2022-11-02 at 5 24 06 PM

Screenshot 2022-11-02 at 5 19 25 PM

3. tested by downloading file. and checking for emoji.
4. Made the table headers left aligned, as requested by kehui.

Screenshot 2022-11-02 at 7 44 20 PM


zachary0kent:
Can you update the test plan to include what previously triggered the bug?

vaishnavi17:
> Can you update the test plan to include what previously triggered the bug and how we can ensure it's fixed?

yeah, I'd previously included a brief explanation of what happened, but I've included a code screenshot now to explain line-by-line what was breaking.

benjamin-shen (approved):
Could you maybe move the utility helper functions to another file? Other than that, lgtm! Thanks for the fixes :))

vaishnavi17:
> Could you maybe move the utility helper functions to another file? Other than that, lgtm! Thanks for the fixes :))

hm, most of them use the global consts in the file tho.

src/tools/export-plan/types.ts
```diff
@@ -0,0 +1,12 @@
+// represents a coloured bubble that shows the group a requirement falls in
+export type bubbleData = {
```
benjamin-shen, line 2: Can you capitalize type names?

src/tools/export-plan/utilities.ts
```diff
@@ -0,0 +1,104 @@
+import { isPlaceholderCourse, isCourseTaken } from '../../utilities';
+import { getCollegeAbbrev } from '../../data';
+import store from '../../store';
+import { pdfColors } from '@/assets/constants/colors';
+import { semesterRows, bubbleData } from './types';
+
+export const trimEmptySems = (sems: readonly FirestoreSemester[]): readonly FirestoreSemester[] => {
+  if (sems.length === 0) return [];
```
benjamin-shen, line 8: you can remove this line

PR #761, opened 2022-12-02
Refactor requirement graph builder

benjamin-shen:
### Summary 

This pull request refactors the requirement graph builder to a pipeline design pattern. For more information, please see https://courseplan.notion.site/FA22-Arbitrary-Opt-In-Choices-9bbc7111e52f476c9fdb1775299c0d18.

This is also a good first step in reorganizing the requirements directory.

Original folder structure:
```
src/requirements/
  - requirement-graph.ts
  - requirement-graph-builder.ts
  - requirement-graph-builder-from-user-data.ts
  ...
```

New folder structure:
```
src/requirements/
  - graph/
    - processor/
      - index.ts
      - definition.ts (processor interface definition)
      - build-initial-graph.ts
      - add-toggleable-choices.ts
      - add-selectable-choices.ts
      - add-opt-out-choices.ts
      - add-arbitrary-opt-in-choices.ts
      - remove-constraint-violation-edges.ts
    - index.ts
    - types.ts
    - builder.ts (current code in this file will be deprecated, but it is kept for the tests)
  - requirement-graph-builder-from-user-data.ts
  ...
```

After this PR, multiple processes can be composed to go from graphA to graphB. You can think of it like a more modular/flexible version of the original graph building algorithm. It also enables us to easily test individual units of the system.

Remaining TODOs:

- [ ] Implement the complete proposed flow in a future PR

### Test Plan 

Thoroughly test all requirements-related functionality. There should be no change from before!

### Notes
I am open to different naming for `processor`. Initially, it was `transformer` but I didn't like that since I wasn't mutating the base graph. Let me know if you have any suggestions!


src/requirements/graph/definition.ts
```diff
@@ -96,11 +94,13 @@ export default class RequirementFulfillmentGraph<
 
   public copy(): RequirementFulfillmentGraph<Requirement, Course> {
     const newCopy = new RequirementFulfillmentGraph<Requirement, Course>();
+    // newCopy.addGraph(this);
+    // return newCopy;
```
hahnbeelee, line 98: delete?

src/requirements/graph/processor/add-arbitrary-opt-in-choices.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+export const addArbitraryOptInChoicesParameters = <
+  Requirement extends string,
+  Course extends CourseWithUniqueId
+>(
+  graph: RequirementFulfillmentGraph<Requirement, Course>,
+  { userChoiceOnRequirementOverrides }: AddArbitraryOptInChoicesParameters<Requirement>
+): RequirementFulfillmentGraph<Requirement, Course> => {
```
hahnbeelee, line 40: OOC Is this a generic function? It's kinda hard to read/tell what's going on.

src/requirements/graph/processor/copy-graph.ts
```diff
@@ -0,0 +1,14 @@
+import GraphProcessor from './definition';
```
hahnbeelee, line 1: nit - it's really satisfying how all of your files in the processor folder are formatted similarly. is there anywhere else you can put this? a utils folder by chance?

src/requirements/graph/index.ts
```diff
@@ -0,0 +1,3 @@
+import RequirementFulfillmentGraph from './definition';
```
hahnbeelee, line 1: what's the design decision of putting this in `definition.ts` and exporting it out of `index.ts`? Why don't you just put it in `index.ts`?

src/requirements/graph/processor/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Compose zero or more graph processors on a requirement graph.
+ */
+export function process<Requirement extends string, Course extends CourseWithUniqueId>(
+  graph = new RequirementFulfillmentGraph<Requirement, Course>(),
+  ...processors: readonly GraphProcessor<Requirement, Course>[]
+) {
+  return processors.reduce((newGraph, processor) => processor.process(newGraph), graph);
+}
```
hahnbeelee, line 21: you went sicko mode on this

src/requirements/requirement-constraints-utils.ts
```diff
@@ -53,7 +54,6 @@ export const getConstraintViolationsForSingleCourse = <Requirement extends strin
         requirement,
         constraintViolatingRequirementsForCurrentRequirement
       );
-      constraintViolationsGraph.addRequirementNode(requirement);
```
hahnbeelee, line 56: porque?

benjamin-shen: unnecessary/redundant -- addEdge does this as the first step

src/requirements/graph/processor/add-arbitrary-opt-in-choices.ts
benjamin-shen: Yes. I just realized I named it incorrectly lol. But you're right, it probably would be more clear if I labeled it as a function.

src/requirements/graph/processor/copy-graph.ts
benjamin-shen: hmm I think it actually looks nice in the folder

src/requirements/graph/processor/add-arbitrary-opt-in-choices.ts
hahnbeelee: LOL

src/requirements/graph/index.ts
benjamin-shen: whoops, definitely an oversight! thanks for the catch

src/requirements/graph/processor/add-arbitrary-opt-in-choices.ts
```diff
@@ -0,0 +1,47 @@
+import GraphProcessor from './interface';
+import RequirementFulfillmentGraph from '..';
+import { CourseWithUniqueId } from '../types';
+
+export type AddArbitraryOptInChoicesParameters<Requirement extends string> = {
```
zachary0kent, line 5: Why is this bound necessary?

src/requirements/graph/processor/add-arbitrary-opt-in-choices.ts
```diff
@@ -0,0 +1,47 @@
...
+  Requirement extends string,
+  Course extends CourseWithUniqueId
+> implements GraphProcessor<Requirement, Course> {
+  private arbitraryOptInChoicesParameters: AddArbitraryOptInChoicesParameters<Requirement>;
+
+  constructor(arbitraryOptInChoicesParameters: AddArbitraryOptInChoicesParameters<Requirement>) {
+    this.arbitraryOptInChoicesParameters = arbitraryOptInChoicesParameters;
+  }
+
+  public process(graph: RequirementFulfillmentGraph<Requirement, Course>) {
```
zachary0kent, line 29: Is it possible to indicate at the type level whether each operation mutates the graph? I think that would be useful here, because some operations do while others don't.

src/requirements/graph/processor/add-opt-out-choices.ts
```diff
@@ -0,0 +1,48 @@
...
+  Course extends CourseForRequirementGraph
+>(
+  graph: RequirementFulfillmentGraph<Requirement, Course>,
+  { userChoiceOnRequirementOverrides }: AddOptOutChoicesParameters<Requirement>
+): RequirementFulfillmentGraph<Requirement, Course> {
+  graph.getAllCourses().forEach(course => {
+    userChoiceOnRequirementOverrides[course.uniqueId]?.optOut?.forEach(requirement => {
+      graph.removeEdge(requirement, course);
+    });
+  });
```
zachary0kent, line 45: Can we abstract this pattern somehow? It's repeated almost exactly in `addArbitraryOptInChoices`. Maybe we could define an operation that works on a supertype of `AddOptOutChoicesParameters` and `AddArbitraryOptInChoicesParameters`.

src/requirements/graph/processor/copy-graph.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Essentially a noop, but copies the graph. This can be useful for debugging.
+ */
+export default class CopyGraph<Requirement extends string, Course extends CourseWithUniqueId>
+  implements GraphProcessor<Requirement, Course> {
+  // eslint-disable-next-line class-methods-use-this
+  public process(graph: RequirementFulfillmentGraph<Requirement, Course>) {
+    return graph.copy();
+  }
+}
```
zachary0kent, line 14: Could this be the default implementation of `GraphProcessor` instead of creating a dedicate subclass?

zachary0kent (commented):
Sweet. Can we refactor `GraphProcessor` into a visitor pattern, or maybe have `GraphProcessor` be an abstract class with some default implementations?

src/requirements/graph/processor/add-opt-out-choices.ts
benjamin-shen: I tried adding an abstract class `AddOverriddenChoices` that each of these visitors extended, but it ended up adding more code. It should be okay since we're not planning on adding any more kinds of overridden choices.

benjamin-shen:
I could not add `graph.accept(visitor)` without introducing a dependency cycle. It is necessary for graph visitors to be dependent on the graph definition. So, the way we'll apply a graph visitor to a graph is something like `visitor.process(graph)`.

src/requirements/graph/processor/add-arbitrary-opt-in-choices.ts
benjamin-shen: We can possibly use an immutable `ReadonlyRequirementFulfillmentGraph` as the input graph. I think this can be tackled in a later PR, though.

benjamin-shen: Added in #762

PR #773, opened 2023-02-08
Added Applied Economics Minor

mirandayu131:
**Summary**
Added the Business minor for Applied Economics in the SC Johnson College of Business. 

_Information on the requirements is sourced from the following:_ 
[https://business.cornell.edu/programs/undergraduate/minors/applied-economics/]

**Test Plan**
Test AE minor!


zachary0kent:
@jerrry1123 @rohanmaheshwari430 might want to review this for dev portfolio.

noschiff:
make sure the minor is added to the req JSON!

src/data/index.ts
```diff
@@ -375,6 +376,12 @@ const json: RequirementsJson = {
     },
   },
   minor: {
+    AEM: {
+      name: 'Applied Economics and Management',
+      schools: ['AG', 'BU'],
```
noschiff, line 381: shouldn't every college except dyson be allowed to do the minor?

zachary0kent:
Remember to run `npm run req-gen`!

noschiff (changes requested):
Can we name this "Applied Economics" instead of AEM? The [minor](https://business.cornell.edu/programs/undergraduate/minors/applied-economics/) is officially applied economics, whereas AEM usually refers to the Dyson major (M for & Management).

src/data/index.ts
noschiff: @zachary0kent am I wrong about how this works

noschiff:
@mirandayu131 can you add more information to the PR overview? It'd be good to have a link to the source of requirements for the reviewer to reference, as well as a test plan. The test plan for adding a major is usually pretty brief though :)

noschiff (commented):
Looks great! I'll leave the approval + testing to one of the devs for their portfolio :)

src/data/index.ts
jerrry1123: i know that some colleges like engineering ahve their own version of their AEM minor

jerrry1123:
I did the testing plan and all the requirements look right, I just think that it might be better to rename the minor to non-engineering or something like that to differentiate this minor from the engineering business minor.

jerrry1123 (approved):
I did the testing plan and all the requirements look right, I just think that it might be better to rename the minor to non-engineering or something like that to differentiate this minor from the engineering business minor.

src/data/index.ts
zachary0kent: I believe this refers to which college offers the minor

noschiff: where do we store which colleges are eligible? @zachary0kent

zachary0kent: I don't believe we do? I'm not of such restrictions typically existing anyways.

PR #777, opened 2023-02-10
Add Dependency Bump to User Choice

mirandayu131:
The most recent commit on master corresponds to a frontend bug fix, which I need on user-choice. Used git cherry-pick <hash> to selectively copy Zak's commit.


zachary0kent:
> just curious: why cherry pick this rather than rebase user-choice onto master? @zachary0kent thoughts?



> just curious: why cherry pick this rather than rebase user-choice onto master? @zachary0kent thoughts?

So we can worry about conflicts later

PR #779, opened 2023-02-16
Added Earth and atmosphere Minor

rohanmaheshwari430:
### Summary 

In this PR, we add the requirements for the Earth And Atmospheric Sciences minor to the data/minors directory. 

Details about the requirements to fulfill this minor can be found here: [EAS Minor Requirements](https://www.eas.cornell.edu/eas/programs/undergraduate-programs/undergraduate-minors/earth-and-atmospheric-sciences-minor)

### Test Plan 

ran `npm run test -- -u` to update the requirements snapshot

ran `npm test` to ensure that the addition of the minor and its requirements did not affect previous infrastructure.


src/data/minors/earth-atmo.ts
```diff
@@ -0,0 +1,53 @@
...
+    perSlotMinCount: [3],
+    slotNames: ['Course'],
+  },
+  {
+    name: 'Additional Courses: 3000 or above',
+    description:
+      'Additional EAS courses at the 3000-level or higher. These courses may include, but are not limited to, additional courses from the above list, undergraduate research courses, and outdoor field courses.',
+    source:
+      'https://www.eas.cornell.edu/eas/programs/undergraduate-programs/undergraduate-minors/earth-and-atmospheric-sciences-minor',
+    fulfilledBy: 'self-check',
```
noschiff, line 45: I don't understand why this is a self check

rohanmaheshwari430: I was thinking that the description implies that the course can be from many different categories (undergraduate research, outdoor field, etc), so it would be tough to check all of those conditions. Instead, I will make a checker that checks if the course is 3000+

zachary0kent (commented):
Tbh, I am not sure of the best way to handle this case. I think it's okay to merge until we have a better solution.

zachary0kent (approved):
Looks good! Added the EAS major and poked at the requirements a bit.

PR #780, opened 2023-02-17
Added Standardized Abbreviations for Colleges, Majors, and Minors

jerrry1123:
### Summary 
- Added abbreviations for all colleges, majors, minors, and grad requirements
- added said abbreviations into the progress tracker toolcard

Remaining TODOs:

None

### Test Plan 

Test that the abbreviations are working correctly for both the export toolcard and the progress tracker toolcard.

### Notes 

There's a bug with the Applied Econ Minor because it's not in index.ts yet, we should add that soon.


src/requirements/decorated-requirements.json
```diff
@@ -217243,97 +217283,12 @@
...
-          ]
-        }
-      ]
-    },
     "AEROSPACE": {
       "name": "Aerospace Engineering",
       "schools": [
         "EN"
       ],
+      "abbrev": "Aero Eng",
```
noschiff, line 217291: can we exclude the abbrev from the computed json like we do with advisors?

noschiff (changes requested):

src/requirements/decorated-requirements.json
jerrry1123: it was included in req-gen which i had to run to pass the requirements check

src/requirements/types.ts
```diff
@@ -26,6 +26,7 @@ export type Major<R> = Readonly<{
   /** College requirements that have been "specialized" for this major */
   specializations?: readonly R[];
   advisors?: AdvisorGroup;
+  abbrev: string;
```
noschiff, line 29: can we make this an optional field and exclude it from the generated JSON?

noschiff (changes requested):

src/requirements/decorated-requirements.json
noschiff: yeah but we can exclude it from req-gen

src/components/Tools/ProgressTracker.vue
```diff
@@ -124,6 +124,22 @@ export default defineComponent({
       }
     },
   },
+  methods: {
+    generateRequirementProgressString(req: string, reqGroup: string): string {
```
zachary0kent, line 128: This name is too long. For example, the prefix `generate` doesn't convey any additional information.

src/components/Tools/ProgressTracker.vue
```diff
@@ -124,6 +124,22 @@ export default defineComponent({
...
+        case 'College':
+          return getCollegeAbbrev(req) + ' ' + reqGroup + ' Requirements';
+        case 'Major':
+          return getMajorAbbrev(req) + ' ' + reqGroup + ' Requirements';
+        case 'Minor':
+          return getMinorAbbrev(req) + ' ' + reqGroup + ' Requirements';
+        case 'Grad':
+          return getGradAbbrev(req) + ' ' + reqGroup + ' Requirements';
+        default:
+          return '';
```
zachary0kent, line 139: Can we refactor this big switch to use a function that returns the correct abbreviation given a requirement?

src/components/Tools/ProgressTracker.vue
```diff
@@ -124,6 +124,22 @@ export default defineComponent({
       }
     },
   },
+  methods: {
+    generateRequirementProgressString(req: string, reqGroup: string): string {
+      switch (reqGroup) {
+        case 'College':
+          return getCollegeAbbrev(req) + ' ' + reqGroup + ' Requirements';
```
zachary0kent, line 131: Use string interpolation instead

zachary0kent (changes requested):
You are going to have to exclude the abbreviations from the generated JSON for the checks to pass. Check out how Vaish did this in #730.

zachary0kent (changes requested):
We need to exclude the abbreviations from the generated requirements json.

zachary0kent (commented):
There shouldn't be any diff on [src/requirements/decorated-requirements.json](https://github.com/cornell-dti/course-plan/pull/780/files#diff-38bfbe5fd0f55f8bc0869e1e39f5a52bfc0624e24709f1a69017aa7c448f3b1c)

jerrry1123:
The only diff right now is moving the applied econ minor after the aerospace engineering minor instead of before, which might be because i switched the order in another file. should i move it back or leave it as is?

zachary0kent:
> The only diff right now is moving the applied econ minor after the aerospace engineering minor instead of before, which might be because i switched the order in another file. should i move it back or leave it as is?

Oh I see, nevermind. It's alphabetized now which is better.

zachary0kent (changes requested):
It seems the abbreviations appear correctly on the progress tracker toolcard, but not the exported schedule; see 'isst2'.
Screenshot 2023-02-24 at 2 01 56 PM

jerrry1123:
Just added the codes to the export PDF! @zachary0kent

src/requirements/requirement-json-generator.ts
```diff
@@ -355,22 +355,22 @@ const generateDecoratedRequirementsJson = (): DecoratedRequirementsJson => {
...
-    const { requirements, advisors, ...rest } = minorRequirement;
+    const { requirements, advisors, abbrev: abbr, ...rest } = minorRequirement;
     decoratedJson.minor[minorName] = {
       ...rest,
       requirements: decorateRequirements(requirements),
     };
   });
   Object.entries(grad).forEach(([gradName, gradRequirement]) => {
-    const { requirements, advisors, ...rest } = gradRequirement;
+    const { requirements, advisors, abbrev: abbr, ...rest } = gradRequirement;
```
zachary0kent, line 373:
```suggestion
    const { requirements, advisors, abbrev, ...rest } = gradRequirement;
```

zachary0kent (approved):
Looks good! Just left one small suggestion.

src/data/index.ts
```diff
@@ -141,291 +141,339 @@ const json: RequirementsJson = {
...
       requirements: englishRequirements,
       advisors: englishAdvisors,
+      abbrev: 'English',
     },
     ENV: {
       name: 'Environmental Engineering',
       schools: ['EN'],
       requirements: envEngineeringRequirements,
       advisors: envEngineeringAdvisors,
+      abbrev: 'EnvEng',
```
noschiff, line 271: is there any reason some abbreviations are space separated while others are PascalCase?

PR #785, opened 2023-02-27
Added Game Design Minor

KaylinChan:
### Summary 
This PR adds the Game Design minor. More info about the minor can be found here: https://www.cs.cornell.edu/undergrad/minors/game-design-minor

#### Test Plan: 
- Test if CS 3152 and CS 4152 automatically fulfill requirements 1 and 2. 
- Test the "4 Additional Courses" requirement

#### Notes: 
- The online requirements list is an "includes but not limited to" list so I was unable to comprehensively validate the 4 additional courses requirement (there are many classes that students can petition for)
    - Therefore, the 4 additional courses requirement acts like the CS "major approved electives" requirement where the user can manually add to it but there are also classes that automatically satisfy it


src/data/minors/gamedesign.ts
```diff
@@ -0,0 +1,119 @@
...
+          'PSYCH 4270',
+          'PSYCH 4280',
+          'PSYCH 4331',
+          'PSYCH 4320',
+          'PSYCH 4360',
+          'PSYCH 4370',
+          'PSYCH 4650',
+          'PSYCH 4770',
+          'PSYCH 4780',
+        ]) ||
```
zachary0kent, line 98: Any reason we shouldn't pass all of these courses into a single call to `courseMatchesCodeOptions`? Also, I think it would be cleaner if we factored this list of courses out of the function and into a constant.

KaylinChan:
I updated the requirements by putting the long list of courses into a const, and renamed all the files/instances accordingly!

zachary0kent (approved):
Looks great! Tested out adding the game design minor and everything works as expected.

noschiff:
Congratulations on making your first real contribution! Welcome to CoursePlan, @KaylinChan! 🎉🚀

PR #788, opened 2023-03-03
Add data science minor

elizabeth-tang:
**Add data science minor to minors.**

This is a new minor offered starting Spring 2023: https://stat.cornell.edu/academics/undergraduate/data-science-minor/minor-requirements

Test Plan

- Test if ENGRD 2700 or ECE 3100 fulfill core statistics requirement.
- Test if the two subrequirements in "Categories" requirement work as expected - 4 total courses in 3 categories.

Notes

- There is no existing way to check against classes taken for a student's major.
   - Applies to computing requirement. If intro computing is required for major, then they cannot use intro computing for the req. Currently addressed as a checkerWarning.
   - Applies to Categories requirement. Courses cannot also be used in major core. Currently included in just the description as checkerWarning produces user-inconveniencing behavior.
- The minor is technically offered under the CIS school.
- No set information available on advisors yet. I tentatively included the undergraduate director for data science as the advisor.


zachary0kent:
@PabloRaigoza @mirandayu131 Y'all might want to review this for the upcoming dev portfolio.

mirandayu131 (commented):
Looks good to me! Do we need to include a checker warning so the student is not declaring both a Data Science Minor and Data Science Concentration in the InfoSci Minor?

PabloRaigoza (approved):
Good work. Are we sure that this minor is only available to EN, AG, and AS?

zachary0kent:
> Good work. Are we sure that this minor is only available to EN, AG, and AS?

These refer to the colleges that offer the minor, not those eligible to complete it.

mirandayu131 (approved):
Nice work!

PR #789, opened 2023-03-05
Integrate Replace Modal

mirandayu131:
Draft implementation for integrating the replace/remove modals.

- Implemented Remove Modal
- Replace requirement for courses that don't exist in current schedule

Remaining TODOs:

- Replace requirements for courses that already exist in current schedule


zachary0kent:
I think we are going to have to rethink the original implementation of the replace and remove modals.

PR #791, opened 2023-03-05
Refactor Teleport Modal to Use Composition API

zachary0kent:
### Summary 

This PR refactors `TeleportModal.vue` to use the new Vue composition API. I plan to create a modal similar to the Teleport Modal, but the options API which we currently use makes composing components very difficult. This PR also shows how the composition API allows us to make emits and props much, much more type safe. Looking ahead, I think a future refactor of the codebase to use the composition API would definitely be worth it.

### Test Plan 

E2E tests should assure us that this doesn't break anything--if it did, nearly every test would fail. Regardless, you can perform the usual tasks that use the teleport modal (such as adding a course) and ensure everything looks OK.

### DIFF
--- a/src/components/Modals/TeleportModal.vue
+++ b/src/components/Modals/TeleportModal.vue
@@ -49,66 +49,83 @@
     </div>
   </Teleport>
 </template>
-<script lang="ts">
-import { defineComponent, ref, PropType } from 'vue';
+
+<script setup lang="ts">
+import { computed, ref } from 'vue';
 import store from '@/store';
 
-export default defineComponent({
-  props: {
-    title: { type: String, default: '' },
-    contentClass: { type: String, required: true },
-    leftButtonText: { type: String, default: '' },
-    rightButtonText: { type: String, default: '' },
-    rightButtonImage: { type: String, default: '' },
-    rightButtonAlt: { type: String, default: '' },
-    rightButtonIsDisabled: { type: Boolean, default: false },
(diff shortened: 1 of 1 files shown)


zachary0kent:
@elizabeth-tang You can review this for dev portfolio if you want

noschiff:
beautiful… I'm crying, @zachary0kent. thank you 🙏🏻❤️

jerrry1123:
Looks good Zak! I'm wondering how this would interact with the current bug facing the profile page where the scrolling on the sidebar would sometimes be disabled after closing a teleport modal. Do you think any of this refactoring would potentially fix the issue or at least help us identify why it's happening? Since that PR isn't merged in yet it would be pretty hard to tell, but I tested this with similar teleport modals that were previously in the profile page and those seem to work fine.

zachary0kent:
> Looks good Zak! I'm wondering how this would interact with the current bug facing the profile page where the scrolling on the sidebar would sometimes be disabled after closing a teleport modal. Do you think any of this refactoring would potentially fix the issue or at least help us identify why it's happening? Since that PR isn't merged in yet it would be pretty hard to tell, but I tested this with similar teleport modals that were previously in the profile page and those seem to work fine.

This refactor should, by design, should not change any user-facing functionality. It could indeed though make these issues easier to identify through use of a simpler API.

jerrry1123 (approved):
I tested a few teleport modals and they all seemed to work properly, so this looks good to me!

PR #792, opened 2023-03-05
Set up NCF Model in Courseplan and Tested it on a Sample Dataset

jerrry1123:
### Summary 
- Added the code to train and evaluate a NCF model with given CSV files
- Added code to fetch a sample dataset into the appropriate CSV files

Remaining TODOs:
- Generate the appropriate CSVs from course data
- Turn evaluation of individual users into an API format so that Courseplan can call it
- Conduct testing on model hyperparameters 
- Incorporate pre-training to increase accuracy - see (https://github.com/microsoft/recommenders/blob/main/examples/02_model_collaborative_filtering/ncf_deep_dive.ipynb)
- Improve the metrics with the above methods (right now they're not great, but I can add more optimizations once we're sure the basics work)
### Test Plan 
**IMPORTANT**: Follow the installation instructions here: https://docs.google.com/document/d/1e53btTwYI6-3LAB2mH4aE1FpmUJ-crL6I2vPUfmjGmg/edit
Once done, run NCFTest.py through the terminal and make sure that the terminal provide feedback saying that training and predictions are finished. Training could take a while, so be patient, but predictions shouldn't take more than a minute or so. Please make sure this works on your system and Slack me if you run into any issues during installation or running the code so I can add fixes into the doc above.


zachary0kent (changes requested):
It seems like this PR is carrying along some uninvited guests from another branch

jerrry1123:
> It seems like this PR is carrying along some uninvited guests from another branch

Yes I'm trying my best to fix it right now.

PR #794, opened 2023-03-05
 Set up NCF Model in Courseplan and Tested it on a Sample Dataset

jerrry1123:
Summary
    Added the code to train and evaluate a NCF model with given CSV files
    Added code to fetch a sample dataset into the appropriate CSV files

Remaining TODOs:
Generate the appropriate CSVs from course data
Turn evaluation of individual users into an API format so that Courseplan can call it
Conduct testing on model hyperparameters
Incorporate pre-training to increase accuracy - see (https://github.com/microsoft/recommenders/blob/main/examples/02_model_collaborative_filtering/ncf_deep_dive.ipynb)
Improve the metrics with the above methods (right now they're not great, but I can add more optimizations once we're sure the basics work)

Test Plan

IMPORTANT: Follow the installation instructions here: https://docs.google.com/document/d/1e53btTwYI6-3LAB2mH4aE1FpmUJ-crL6I2vPUfmjGmg/edit
Once done, first run SampleDatasetFetcher.py to download the training and testing data, and then NCFTest.py through the terminal and make sure that the terminal provide feedback saying that training and predictions are finished. Training could take a while, so be patient, but predictions shouldn't take more than a minute or so. Please make sure this works on your system and Slack me if you run into any issues during installation or running the code so I can add fixes into the doc above.

### DIFF
--- a/src/components/Modals/CourseMenu.vue
+++ b/src/components/Modals/CourseMenu.vue
@@ -16,14 +16,14 @@
         </div>
         <img
           class="courseMenu-arrow"
-          src="@/assets/images/sidearrow.svg"
+          src="@/assets/images/downarrow.svg"
           alt="arrow to expand edit course color"
         />
 
         <div
           v-if="displayColors"
           class="courseMenu-content courseMenu-colors"
-          :class="{ 'courseMenu-colors--left': isLeft }"
+          :style="{ zIndex: zIndexColors }"
         >
           <button
             v-for="(color, index) in colors"
@@ -69,13 +69,13 @@
         </div>
         <img
           class="courseMenu-arrow"
-          src="@/assets/images/sidearrow.svg"
+          src="@/assets/images/downarrow.svg"
           alt="arrow to expand edit course credits"
         />
         <div
           v-if="displayEditCourseCredits"
           class="courseMenu-content courseMenu-editCredits courseMenu-centerCredits"
-          :class="{ 'courseMenu-editCredits--left': isLeft }"
+          :style="{ zIndex: zIndexEditCredits }"
         >
           <div
             v-for="credit in makeCreditArary()"
@@ -129,6 +129,8 @@ export default defineComponent({
       displayColors: false,
       displayEditCourseCredits: false,
       tooltipColor: '',
+      zIndexColors: 1,
+      zIndexEditCredits: 1,
     };
   },
   computed: {
@@ -165,6 +167,12 @@ export default defineComponent({
     },
     setDisplayColors(bool: boolean) {
       this.displayColors = bool;
+      if (bool) {
+        this.zIndexColors = 3;
+        this.zIndexEditCredits = 1;
+      } else {
+        this.zIndexColors = 1;
+      }
     },
     setDisplayColorTooltip(bool: boolean, color: string) {
       if (bool) {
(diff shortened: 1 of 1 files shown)


zachary0kent (changes requested):
I don't think we should put a massive amount of example training data in the codebase. Instead, can you write a shell script that downloads the training data, performs all of the steps in the instructions document, runs the code, and then cleans up the now unneeded training data?

jerrry1123:
> I don't think we should put a massive amount of example training data in the codebase. Instead, can you write a shell script that downloads the training data, performs all of the steps in the instructions document, runs the code, and then cleans up the now unneeded training data?

I can remove it right now and just instruct testers to run the data download file, but in the final iteration we will probably need the CSVs for courses in the codebase so we're not fetching it again for every user

zachary0kent:
> > I don't think we should put a massive amount of example training data in the codebase. Instead, can you write a shell script that downloads the training data, performs all of the steps in the instructions document, runs the code, and then cleans up the now unneeded training data?
> 
> I can remove it right now and just instruct testers to run the data download file, but in the final iteration we will probably need the CSVs for courses in the codebase so we're not fetching it again for every user

I don't see why this needs to be the case--can't we generate the training CSVs on the fly if we only train the model every so often. Also, it is just not feasible to send a CSV of this size to the client.

jerrry1123:
> > > I don't think we should put a massive amount of example training data in the codebase. Instead, can you write a shell script that downloads the training data, performs all of the steps in the instructions document, runs the code, and then cleans up the now unneeded training data?
> > 
> > 
> > I can remove it right now and just instruct testers to run the data download file, but in the final iteration we will probably need the CSVs for courses in the codebase so we're not fetching it again for every user
> 
> I don't see why this needs to be the case--can't we generate the training CSVs on the fly if we only train the model every so often. Also, it is just not feasible to send a CSV of this size to the client.

mb You're probably right now that I think about it again

zachary0kent:
Regardless, can you write a shell script that downloads the training data and performs these steps automatically?

jerrry1123:
> Regardless, can you write a shell script that downloads the training data and performs these steps automatically?

Yes I already did and added it to the testing instructions

zachary0kent:
> > Regardless, can you write a shell script that downloads the training data and performs these steps automatically?
> 
> Yes I already did and added it to the testing instructions

I'm more asking for a shell script that I can run that will automatically run these instructions.

jerrry1123:
> > > Regardless, can you write a shell script that downloads the training data and performs these steps automatically?
> > 
> > 
> > Yes I already did and added it to the testing instructions
> 
> I'm more asking for a shell script that I can run that will automatically run these instructions.

I'm not sure if I can run a shell script since I'm on Windows. It's probably not a great idea regardless since the installation instructions include a lot of stuff outside of the command console, and there are various other packages that may need to be installed depending on the user's system. As for the training loop, it is just two python files, so it seems a bit redundant to put those into a shell script

zachary0kent (commented):
Will this code need to be in the repo once the recommendations algorithm is complete? I'm more in favor of creating a `recommendations` branch into which we'll merge our WIP recommendations work. Then, once it's complete, we can merge `recommendations` into `master`.

PR #796, opened 2023-03-06
courses + meta data script

rohanmaheshwari430:
### Summary 

This pull request is the first step towards the courses migration project. In this PR we:

1. Create a courses collection for all courses
2. Create a mapping from courses to available rosters (in chronological order)
3. Create a mapping from crseId to catalogNbr (which is subject + code)

Remaining TODOs:

- [ ] Develop an interface between the Firebase data and the Frontend
- [ ] Implement said interface
- [ ] Remove/scrub any previous references to courses.json

### Test Plan 

courses:
Tested the new script against the `old courses-json-generator.ts` and ensured that the same amount of courses were retrieved. All did sanity checks on roster subject pairs that had 0 or 1 courses to ensure that they were retrieved and properly added. 

course -> available roster mapping:
Performed sanity checks like ensuring courses like CS 1110, BTRY 3080 had every semester in their roster list. Also checks if only Fall/only Spring courses had only Fall/Spring rosters (i.e JAPAN 1101, JAPAN 2202). 

Example sanity check:

Screen Shot 2023-03-06 at 11 50 12 AM

crseId -> catalogNbr mapping:

Performed sanity checks on random courses by checking the entry against the class roster website.

### Notes

- This PR adds a few new collections to the Firebase. They do not interfere with existing data/schemas.


scripts/migration/courses-populate.ts
```diff
@@ -0,0 +1,171 @@
...
+  const subjects: readonly { roster: string; subject: string }[] = await fetch(
+    `https://classes.cornell.edu/api/2.0/config/subjects.json?roster=${roster}`
+  )
+    .then(res => res.json())
+    .then(jsonRes => jsonRes.data.subjects)
+    .then(jsonSubjects =>
+      jsonSubjects.map(jsonSubject => ({ roster, subject: jsonSubject.value }))
+    );
+  return subjects;
+};
```
zachary0kent, line 58:
```suggestion
const retrieveAvailableSubjects = (roster: string): Promise<readonly { roster: string; subject: string }[]> =>
  fetch(`https://classes.cornell.edu/api/2.0/config/subjects.json?roster=${roster}`)
    .then(res => res.json())
    .then(jsonRes => jsonRes.data.subjects)
    .then(jsonSubjects =>
      jsonSubjects.map(jsonSubject => ({ roster, subject: jsonSubject.value }))
    );
```

scripts/migration/courses-populate.ts
```diff
@@ -0,0 +1,171 @@
...
+/* Retrieves and formats available courses for a {roster: string, subject: string} object */
+const retrieveAvailableCourses = async (roster: string, subject: string) => {
+  const courses: CourseFullDetail[] = await fetch(
+    `https://classes.cornell.edu/api/2.0/search/classes.json?roster=${roster}&subject=${subject}`
+  )
+    .then(res => res.json())
+    .then(jsonRes => jsonRes.data.classes)
+    .then(jsonCourses => jsonCourses.map(jsonCourse => courseFieldFilter(jsonCourse)));
+  return courses;
+};
```
zachary0kent, line 69:
```suggestion
const retrieveAvailableCourses = (roster: string, subject: string): Promise<CourseFullDetail[]> => {
  fetch(`https://classes.cornell.edu/api/2.0/search/classes.json?roster=${roster}&subject=${subject}`)
    .then(res => res.json())
    .then(jsonRes => jsonRes.data.classes)
    .then(jsonCourses => jsonCourses.map(jsonCourse => courseFieldFilter(jsonCourse)));
```

scripts/migration/courses-populate.ts
```diff
@@ -0,0 +1,171 @@
...
+
+/* Helper functions to clean and format class roster data */
+const cleanField = (value: string | null | undefined) =>
+  value?.replace(/\u00a0/g, ' ') || undefined;
```
zachary0kent, line 73: iirc @noschiff already made something like this

scripts/migration/courses-populate.ts
```diff
@@ -0,0 +1,171 @@
...
+        .set({ rosters: [roster] });
+    }
+    /* Mapping that maps from a course's crseId to their course code */
+    await crseIdToCatalogNbrCollection
+      .doc(course.crseId.toString())
+      .set({ catalogNbr: `${subject} ${course.catalogNbr}` });
+    /* Writing each course to the 'courses' collection */
+    await coursesCollection.doc(roster).collection(subject).doc(course.catalogNbr).set({ course });
+  });
+};
```
zachary0kent, line 140: At a glance this might not behave as expected; the promise returned by `populateCourses` will immediately resolve
```suggestion
const populateCourses = (roster: string, subject: string, courses: CourseFullDetail[]) =>
  Promise.all(courses.map(async course => {
    /* Computing a mapping that maps a course to a list of semesters it is offered in */
    const courseAndRostersDoc = await availableRostersForCourseCollection
      .doc(`${subject} ${course.catalogNbr.toString()}`)
      .get();
    if (courseAndRostersDoc.exists) {
      await availableRostersForCourseCollection
        .doc(`${subject} ${course.catalogNbr.toString()}`)
        .update({ rosters: FieldValue.arrayUnion(roster) });
    } else {
      await availableRostersForCourseCollection
        .doc(`${subject} ${course.catalogNbr.toString()}`)
        .set({ rosters: [roster] });
    }
    /* Mapping that maps from a course's crseId to their course code */
    await crseIdToCatalogNbrCollection
      .doc(course.crseId.toString())
      .set({ catalogNbr: `${subject} ${course.catalogNbr}` });
    /* Writing each course to the 'courses' collection */
    await coursesCollection.doc(roster).collection(subject).doc(course.catalogNbr).set({ course });
  }));
```

zachary0kent (commented):
Thanks Rohan, this is great! Just left a couple suggestions & thoughts--lmk what you think.

scripts/migration/courses-populate.ts
rohanmaheshwari430: Yeah he did, I looked into his implementation and did not see any export for the his `cleanField`function so I was hesitant to use it. If you guys are cool with it, I can go ahead and export it so I can use it here?

scripts/migration/courses-populate.ts
rohanmaheshwari430: Yeah this is definitely more type safe and clear to the reader. Thanks for spotting this!

scripts/migration/courses-populate.ts
rohanmaheshwari430: Hmm. Could you explain a little bit why the `Promise.all() `is needed to wrap around the `courses.forEach()` loop? `courses` is just an array of data, not promises right? 

Outside of that, I do think the `async` in 118 should be removed, so thanks for that!

zachary0kent: `forEach` will not wait for the promise created for each iteration to resolve. My suggestion creates a promise for each course, and uses `Promise.all` to ensure each of those individual promises resolves before `populateCourses` resolves. Crucially, to do this, it also replaces `forEach` with `map`.

rohanmaheshwari430: Oh gotcha gotcha. you are referring to the async operations that are being applied to each course, and that `forEach` will not wait for those async operations to finish before moving onto the next course. That makes sense. Thanks!

zachary0kent: Yep. It's okay if the operations for each course finish in different orders (which they will now), but we want all of them to finish before `populateCourses` resolves.

scripts/migration/courses-populate.ts
zachary0kent: Yeah, or maybe pull in out into a utilities file?

zachary0kent: Yeah maybe pull it out into a utilities file and then import it from both scripts?

scripts/migration/courses-populate.ts
```diff
@@ -0,0 +1,172 @@
...
+const retrieveAvailableRosters = async () => {
+  const rosters: readonly string[] = await fetch(
+    'https://classes.cornell.edu/api/2.0/config/rosters.json'
+  )
+    .then(res => res.json())
+    .then(jsonRes => jsonRes.data.rosters)
+    .then(jsonRosters => jsonRosters.map(jsonRoster => jsonRoster.slug));
+
+  return rosters;
+};
```
zachary0kent, line 45:
```suggestion
const retrieveAvailableRosters = (): Promise<readonly string[]> => {
  fetch('https://classes.cornell.edu/api/2.0/config/rosters.json')
    .then(res => res.json())
    .then(jsonRes => jsonRes.data.rosters)
    .then(jsonRosters => jsonRosters.map(jsonRoster => jsonRoster.slug));
```

scripts/migration/courses-populate.ts
```diff
@@ -0,0 +1,172 @@
...
+/* Retrieves the subjects across for a roster and creates a list of {roster: string, subject: string} objects */
+const retrieveAvailableSubjects = (
+  roster: string
+): Promise<readonly { roster: string; subject: string }[]> =>
+  fetch(`https://classes.cornell.edu/api/2.0/config/subjects.json?roster=${roster}`)
+    .then(res => res.json())
+    .then(jsonRes => jsonRes.data.subjects)
+    .then(jsonSubjects =>
+      jsonSubjects.map(jsonSubject => ({ roster, subject: jsonSubject.value }))
+    );
```
zachary0kent, line 56:
```suggestion
const retrieveAvailableSubjects = async (
  roster: string
): Promise<readonly { roster: string; subject: string }[]> => {
  const res = await fetch(`https://classes.cornell.edu/api/2.0/config/subjects.json?roster=${roster}`);
  return res.json().data.subjects.map(({ value }) => ({ roster, subject: value }))
}
```
Sorry to make multiple suggestions, but is there a reason something like this wouldn't work? I don't see the need for a bunch of chained `then`s. If so, can you also remove the other unneeded chained `then`s?

scripts/migration/courses-populate.ts
noschiff: It's honestly not a good function at all. I wrote it in #720 just because it was a marginal improvement over what we had, and it was not the purpose of the PR.

noschiff: @zachary0kent also… this _seems_ to work fine because of the `/g`, but is it bad practice to use a global regex without replaceAll()?

scripts/migration/courses-populate.ts
```diff
@@ -0,0 +1,172 @@
...
+/** A helper function to generate a wait promise. Used for cooldown to limit API usage. */
+const wait = (time: number) =>
+  new Promise<void>(resolve => {
+    setTimeout(() => resolve(), time);
+  });
+
+/* Retrieves the rosters available on the class roster API */
+const retrieveAvailableRosters = async () => {
+  const rosters: readonly string[] = await fetch(
+    'https://classes.cornell.edu/api/2.0/config/rosters.json'
```
noschiff, line 38: Could we maybe refactor this into some constant like we do here? https://github.com/cornell-dti/course-plan/blob/0dba33372dd28b7e2f4d7ed30a3e80ef0e98aa88/src/requirements/courses-json-generator.ts#L9

noschiff (commented):
I don't think this should be in `scripts/migration/` since it's not a migration script; it's a population/generation script. The old script is in `src/requirements/`, and I don't think it should be there because it's not related to requirements (and it's a script), so I think maybe both this one and the old one could be moved to `scripts/population/` or `script/generate/`. What do you think?

scripts/migration/courses-populate.ts
rohanmaheshwari430: Yeah I can do that, it is probably better to use `async` `await` syntax.

rohanmaheshwari430:
> I don't think this should be in `scripts/migration/` since it's not a migration script; it's a population/generation script. The old script is in `src/requirements/`, and I don't think it should be there because it's not related to requirements (and it's a script), so I think maybe both this one and the old one could be moved to `scripts/population/` or `script/generate/`. What do you think?

Yeah, totally agree, will get that done..

scripts/migration/courses-populate.ts
rohanmaheshwari430: Yeah, will do.

scripts/migration/courses-populate.ts
rohanmaheshwari430: I will use it for now, and we can open another task to make a better `cleanField` function. This PR is already getting pretty big.

scripts/migration/courses-populate.ts
rohanmaheshwari430: Will update with proposed syntax below.

zachary0kent (approved):
Great work, thanks Rohan!

PR #798, opened 2023-03-07
Removed Conflicts Feature Flag

zachary0kent:
### Summary 

Removes the conflicts feature flag so that conflicts are visible by default on `user-choice`. To be consistent with the existing frontend flow (and pass the E2E tests), I also modified the logic of the add course modal so that only closes when the add button is clicked. Previously, the modal would close whenever you selected a course.

### Test Plan 

- Ensure that you can add a course as usual through the add modal
- Ensure that requirement conflicts are visible when adding a course creating a constraint violation


rohanmaheshwari430 (approved):
Thanks for doing this, saw you also made a lot of improvements to syntax and formatting too.

PR #801, opened 2023-03-17
Parsed Other Yes List

mirandayu131:
**Summary**

Parsed the PDF for Approved Liberal Studies Courses into a list of course entries that include subject, catalog number, and categories.

_Needs to be updated every semester!_
[https://apps.engineering.cornell.edu/liberalstudies/ApprovedLiberalStudies.pdf]

**Test Plan**
Use Jest tests such that valid courses are included and invalid courses (wrong types) are not excluded from the list.


noschiff (commented):
Great work, Miranda! Do we need to commit the PDF to the repo?

package.json
```diff
@@ -33,10 +33,13 @@
     "esbuild": "^0.15.5",
     "firebase": "^9.9.3",
     "firebase-admin": "^11.5.0",
+    "fs": "^0.0.1-security",
```
zachary0kent, line 36:
```suggestion
```
This isn't a "real" package on `npm`, so we can delete it

package.json
```diff
@@ -33,10 +33,13 @@
...
     "intro.js": "^3.3.1",
     "jspdf": "^2.5.1",
     "jspdf-autotable": "^3.5.25",
     "node-fetch": "^2.6.7",
+    "pdf-to-excel": "^0.3.2",
+    "pdfjs-dist": "^3.4.120",
```
zachary0kent, line 42: Can we change these to developer dependencies so they won't be bundled with the app sent to the client? You can first uninstall them, and then reinstall them with `npm install --save-dev` so that they are installed as dev dependencies.

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,80 @@
...
+        courseEntries.push(course);
+      }
+    } catch {
+      // If any data is invalid, exclude the Course from the result.
+    }
+  }
+  return courseEntries.filter(course => !Number.isNaN(course.catalogNbr));
+}
+
+// getPDF().then(pdf => console.log(genCourseEntries(pdf)));
```
zachary0kent, line 80:
```suggestion
```

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,80 @@
...
+  catalogNbr: number;
+  /** The liberal studies categories satisfied by the course */
+  categories: readonly string[];
+};
+
+// parses PDF into matrix
+export async function getPDF() {
+  const pdfPath = await readFile('./ApprovedLiberalStudies.pdf');
+  return genTextContextMatrix(pdfPath);
+}
```
zachary0kent, line 22: Do we use this anywhere?

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,80 @@
...
+
+// creates list of course entries
+export function genCourseEntries(data) {
+  // eslint-disable-next-line prefer-const
+  let courseEntries: Course[] = [];
```
zachary0kent, line 27:
```suggestion
  const courseEntries: Course[] = [];
```
We should avoid disabling the linter if possible. Here, `courseEntries` is never reassigned, so it can be declared with `const`

zachary0kent (commented):
Great work! I left a couple drive-by comments, and will give a more thorough review soon.

package-lock.json
```diff
@@ -19171,9 +21156,9 @@
       }
     },
     "node_modules/typescript": {
-      "version": "4.1.5",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.1.5.tgz",
-      "integrity": "sha512-6OSu9PTIzmn9TCDiovULTnET6BgXtDYL4Gg4szY+cGsc3JP1dQL8qvE8kShTRx1NIw4Q9IBHlwODjkjWEtMUyA==",
+      "version": "4.9.5",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.9.5.tgz",
+      "integrity": "sha512-1FXk9E2Hm+QzZQ7z+McJiHL4NW1F2EzMu9Nq9i3zAaGqibafqYwCVU6WyWAuyQRRzOlxou8xZSyXLEN8oKj24g==",
```
zachary0kent, line 21161: It looks like the version of `typescript` was bumped, causing these new type errors.

PabloRaigoza (commented):
Nice work. I liked it that you filtered the courses to make sure that they exist before returning them.

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,63 @@
...
+  let start = 0;
+  for (let i = 0; i < data.length; i += 1) {
+    try {
+      const v = data[i][0].str.trim();
+      if (v.slice(0, 5) === 'Notes') {
+        start = i + 1;
+      }
+    } catch {
+      // in case there is no list of approved courses
+    }
```
zachary0kent, line 28: What exception are we trying to catch here?

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,63 @@
...
+        // creditList = creditList.map((x: string) => parseInt(x, 10));
+        // credits = creditList;
+        col = 'course prefix';
+
+        const course = { subject, catalogNbr, categories };
+        courseEntries.push(course);
+      }
+    } catch {
+      // If any data is invalid, exclude the Course from the result.
+    }
```
zachary0kent, line 60: Same comment with this `try/catch`

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,63 @@
...
+      }
+    } catch {
+      // in case there is no list of approved courses
+    }
+  }
+  // assumes that the columns are ordered and there are no empty cells
+  for (let i = start; i < data.length; i += 1) {
+    try {
+      const v = data[i][0].str.trim();
+      if (col === 'course prefix' && v.length <= 5) {
```
zachary0kent, line 34: What is the `<= 5` check here for?

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,63 @@
...
+        col = 'credits';
+      } else if (col === 'credits') {
+        // let creditList = v.split('-');
+        // creditList = creditList.map((x: string) => parseInt(x, 10));
+        // credits = creditList;
+        col = 'course prefix';
+
+        const course = { subject, catalogNbr, categories };
+        courseEntries.push(course);
+      }
```
zachary0kent, line 57: Here, each string 'credits', 'category', etc. represents a different state. Instead of using a string literal, can we refactor this code to use an [enum](https://www.typescriptlang.org/docs/handbook/enums.html) with constructors `College`, `Category`, etc.?

zachary0kent (commented):
Looks great! Just left a couple small comments

scripts/other-yes-list/parse.ts
mirandayu131: It's the logic in order to know when there is a new row because only the prefix column will always less than 5 characters.

scripts/other-yes-list/parse.ts
mirandayu131: Not sure but it gives me an error for invalid data and the suggested fix is that I use try/catch.

zachary0kent: I think we should explicit handle the case that causes an error

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,83 @@
...
+          c = Col.Category;
+          break;
+        case Col.Category:
+          categories = v.split(', ');
+          c = Col.Credits;
+          break;
+        case Col.Credits: {
+          // let creditList = v.split('-');
+          // creditList = creditList.map((x: string) => parseInt(x, 10));
+          // credits = creditList;
```
zachary0kent, line 69:
```suggestion
```

zachary0kent (changes requested):
Can you delete all the commented code?

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,78 @@
...
+  for (let i = 0; i < data.length; i += 1) {
+    try {
+      const v = data[i][0].str.trim();
+      if (v.slice(0, 5) === 'Notes') {
+        start = i + 1;
+      }
+    } catch {
+      // in case there is no list of approved courses
+    }
+  }
```
zachary0kent, line 38:
```suggestion
  let start = 0;
  for (let i = 0; i < data.length; i += 1) {
      const v = data[i][0].str.trim();
      if (v.slice(0, 5) === 'Notes') {
        start = i + 1;
        break;
      }
  }
```
Would this work?

scripts/other-yes-list/parse.ts
```diff
@@ -0,0 +1,78 @@
...
+          courseEntries.push(course);
+          c = Col.CoursePrefix;
+          break;
+        }
+        default:
+          break;
+      }
+    } catch {
+      // If any data is invalid, exclude the Course from the result.
+    }
```
zachary0kent, line 75: Can we handle the error that this is catching explicitly?

package.json
```diff
@@ -68,12 +68,14 @@
     "eslint-plugin-vue": "^8.0.3",
     "jest": "^26.6.3",
     "minimist": "^1.2.6",
+    "pdf-to-excel": "^0.3.2",
```
zachary0kent, line 71:
```suggestion
```
Is this dependency used anywhere?

zachary0kent (changes requested):
Just have one more comment, but otherwise looks good!

scripts/other-yes-list/parse.ts
zachary0kent: @mirandayu131 I think we can still add a `break` here?

zachary0kent (changes requested):
I think (for this PR at least) we can revert the changes on `package.json` and `package-lock.json`.

zachary0kent (approved):
Looks good!!

PR #802, opened 2023-03-20
Migrate to Firebase Analytics API

PabloRaigoza:
### Summary 
Migrated previous analytics that used the gtag.ts to a Firebase Analytics API. This will hopefully integrate more smoothly with the code base.

This pull request is hopefully going to be a good starting place to begin implementing new analytics tracking in the future

- [x] Changed implementations for GTagEvent and GTagLoginEvent
- [x] Changed GTagEvent and GTagLoginEvent calls to use new logEvent and loginEvent calls

### Test Plan 

- Verify in Firebase backend analytics are appearing.
- Remove a course and test if it is show in backend.

### Notes 

- To implement new analytics tracking, use logEvent as opposed to GTagLogEvent


src/analytics.ts
```diff
@@ -209,5 +212,6 @@ export const GTagEvent = (gtag: GTag | undefined, eventType: EventType): void =>
     default:
       return;
   }
-  gtag.event(eventType, eventPayload);
+  if (analytics != null)
+    firebaseLogEvent(analytics, eventType, eventPayload), console.log('logged');
```
elizabeth-tang, line 216: Check me on this, but I think you just need to separate the console log and the firebase event with a semicolon. Otherwise, great job -- didn't even know there was so much analytics that went into CoursePlan!

elizabeth-tang: Also, I noticed some of the ansci minor files were added as part of this PR. Maybe move the changes just to be on the ansci branch?

PR #805, opened 2023-03-25
Course retrieval function and New course interface

rohanmaheshwari430:
### Summary 

In this PR, we create a global course retrieval function that takes in a roster, subject, and number and retrieves the desired course. If the course is not offered in that roster, we find the latest roster that it is offered in and retrieve that version of the course.  

We also define the new courses interface that will be used to populate the VuexStoreState. 

### Test Plan 

Tested the function by importing into BottomBarState.ts and calling it in `getDetailedInformationForBottomBar()`. I logged the output into the console to ensure that the course information that was retrieved when a user clicks on a course card is the same as the current implementation of course information (which is via the class roster API). 

Screen Shot 2023-03-25 at 3 49 29 PM

Line 43 is where we make the call to the new `getCourse` function.
Screen Shot 2023-03-25 at 3 50 08 PM

### DIFF
--- a/src/global-firestore-data/courses.ts
+++ b/src/global-firestore-data/courses.ts
@@ -18,19 +18,20 @@ export const getCourseWithSeasonAndYear = async (
   const roster = seasonAndYearToRosterIdentifier(season, year);
   const course = await getDoc(doc(coursesCollection, `${roster}/${subject}/${number}`));
   if (!course.exists()) {
-    const availableRostersForCourse = (
-      await getDoc(doc(availableRostersForCoursesCollection, `${subject} ${number}`))
-    ).data() as { rosters: string[] };
-    const lastRoster = availableRostersForCourse.rosters.length - 1;
-    const latestCourse = await getDoc(
-      doc(
-        coursesCollection,
-        `${availableRostersForCourse.rosters[lastRoster]}/${subject}/${number}`
-      )
-    );
-    return latestCourse.data()?.course as CornellCourseRosterCourseFullDetail;
+    return getLastOffering(subject, number);
   }
   return course.data()?.course as CornellCourseRosterCourseFullDetail;
+};
+
+const getLastOffering = async (subject: string, number: string) => {
+  const availableRostersForCourse = (
+    await getDoc(doc(availableRostersForCoursesCollection, `${subject} ${number}`))
+  ).data() as { rosters: string[] };
+  const lastRoster = availableRostersForCourse.rosters.length - 1;
+  const latestCourse = await getDoc(
+    doc(coursesCollection, `${availableRostersForCourse.rosters[lastRoster]}/${subject}/${number}`)
+  );
+  return latestCourse.data()?.course as CornellCourseRosterCourseFullDetail;
 };
 
 /**
--- a/src/global-firestore-data/courses.ts
+++ b/src/global-firestore-data/courses.ts
@@ -20,11 +20,7 @@ export const getCourseWithSeasonAndYear = async (
   number: string
 ): Promise<CornellCourseRosterCourseFullDetail> => {
   const roster = seasonAndYearToRosterIdentifier(season, year);
-  const course = await getDoc(doc(coursesCollection, `${roster}/${subject}/${number}`));
-  if (!course.exists()) {
-    return getLastOffering(subject, number);
-  }
-  return course.data()?.course;
+  return getCourse(roster, subject, number);
 };
 
 /**
@@ -42,15 +38,11 @@ export const getCourseWithCrseIdAndRoster = async (
     await getDoc(doc(crseIdToCatalogNbrCollection, `${crseId}`))
   ).data()?.catalogNbr as string;
 
-  const subjectAndNumberList = extractSubjectAndNumber(courseSubjectAndNumber);
-  const subject = subjectAndNumberList[0];
-  const number = subjectAndNumberList[1];
+  const subjectAndNumber = extractSubjectAndNumber(courseSubjectAndNumber);
+  const { subject } = subjectAndNumber;
+  const { number } = subjectAndNumber;
 
-  const course = await getDoc(doc(coursesCollection, `${roster}/${subject}/${number}`));
-  if (!course.exists()) {
-    return getLastOffering(subject, number);
-  }
-  return course.data()?.course;
+  return getCourse(roster, subject, number);
 };
 
 /**
@@ -68,6 +60,26 @@ const getLastOffering = async (subject: string, number: string) => {
     doc(coursesCollection, `${availableRostersForCourse.rosters[lastRoster]}/${subject}/${number}`)
   );
   return latestCourse.data()?.course;
+};
+
+/**
+ * This function attempts to retrieve the course from the specified roster. If it does not
+ * exist, it will retrieve the course from the last roster it was offered in.
+ * @param roster The roster from which the course should be retrieved
+ * @param subject The subject of the course
+ * @param number The number of the course
+ * @returns `Promise<CornellCourseRosterCourseFullDetail>`
+ */
+const getCourse = async (
+  roster: string,
+  subject: string,
+  number: string
+): Promise<CornellCourseRosterCourseFullDetail> => {
+  const course = await getDoc(doc(coursesCollection, `${roster}/${subject}/${number}`));
+  if (!course.exists()) {
+    return getLastOffering(subject, number);
+  }
+  return course.data()?.course;
 };
 
 /**
@@ -89,11 +101,13 @@ const seasonAndYearToRosterIdentifier = (season: FirestoreSemesterSeason, year: 
  * @param courseCode The course code (EX: 'CS 1110')
  * @returns A string[] containing the subject and number (EX: ['CS', '1110'])
  */
-const extractSubjectAndNumber = (courseCode: string): string[] => {
+const extractSubjectAndNumber = (courseCode: string): { subject: string; number: string } => {
   if (courseCode.split(' ').length !== 2) {
-    throw Error(`Invalid course format. Expected courseCode.split(' ') === 2`);
+    throw Error(
+      `Invalid course format. Expected course to be of form subject and number. EX: CS 1110`
+    );
   } else {
-    return courseCode.split(' ');
+    return { subject: courseCode.split(' ')[0], number: courseCode.split(' ')[1] };
   }
 };
 
--- a/src/global-firestore-data/courses.ts
+++ b/src/global-firestore-data/courses.ts
@@ -41,7 +41,7 @@ const seasonAndYearToRosterIdentifier = (season: FirestoreSemesterSeason, year: 
     Fall: 'FA',
     Spring: 'SP',
     Winter: 'WI',
-    Summer: 'SU'
+    Summer: 'SU',
   } as const;
 
   return `${seasonToSemesterMap[season]}${year - 2000}`;
--- a/src/global-firestore-data/courses.ts
+++ b/src/global-firestore-data/courses.ts
@@ -24,7 +24,7 @@ export const getCourseWithSeasonAndYear = async (
   if (!course.exists()) {
     return getLastOffering(subject, number);
   }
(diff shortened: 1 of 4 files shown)


src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,22 @@
+import { doc, getDoc } from 'firebase/firestore';
+import { coursesCollection, availableRostersForCoursesCollection } from '../firebase-config';
+
+export const getCourse = async (roster: string, subject: string, number: string) => {
```
zachary0kent, line 4: Instead of accepting a roster as a string like `FA20`, can this accept the season (of type `FirestoreSemesterSeason`), and year?

zachary0kent, line 4: Can we add a doc comment for this and annotate its return type?

zachary0kent (changes requested):
Just left a couple comments. Also, would it be possible to add a function that looks up a course by `crseId` and roster?

src/global-firestore-data/courses.ts
rohanmaheshwari430: > 

Yeah, I can do that

zachary0kent (commented):
Also, would it be possible to look up an array of courses for the same roster? I imagine this would probably be more efficient than looking up every course 1 by 1 due to the course data collection schema.

src/global-firestore-data/courses.ts
rohanmaheshwari430: Yerp

rohanmaheshwari430:
> Also, would it be possible to look up an array of courses for the same roster? I imagine this would probably be more efficient than looking up every course 1 by 1 due to the course data collection schema.



> Also, would it be possible to look up an array of courses for the same roster? I imagine this would probably be more efficient than looking up every course 1 by 1 due to the course data collection schema.



> Also, would it be possible to look up an array of courses for the same roster? I imagine this would probably be more efficient than looking up every course 1 by 1 due to the course data collection schema.

Would

rohanmaheshwari430:
> Also, would it be possible to look up an array of courses for the same roster? I imagine this would probably be more efficient than looking up every course 1 by 1 due to the course data collection schema.

Would you want that as a separate function, and by which mode do you want to retrieve the courses, crseId or subject + number?

zachary0kent:
> > Also, would it be possible to look up an array of courses for the same roster? I imagine this would probably be more efficient than looking up every course 1 by 1 due to the course data collection schema.
> 
> Would you want that as a separate function, and by which mode do you want to retrieve the courses, crseId or subject + number?

What do you think? I just imagine this would be useful for retrieving the courses in a user's semester, but don't have any specific feelings about the details. Maybe we could include both modes of retrieval, i.e. have each key by a union type of either crseId or subject and roster?

src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,83 @@
...
+    const latestCourse = await getDoc(
+      doc(
+        coursesCollection,
+        `${availableRostersForCourse.rosters[lastRoster]}/${subject}/${number}`
+      )
+    );
+    return latestCourse.data()?.course as CornellCourseRosterCourseFullDetail;
+  }
+  return course.data()?.course as CornellCourseRosterCourseFullDetail;
+};
```
zachary0kent, line 70: Do you think it would be useful to look up a function by a roster like this, i.e. by a string like `FA20`? I feel like it might expose too much of the internal details of our Firestore schema and could also be error-prone

src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,83 @@
...
+const seasonAndYearToRosterIdentifier = (season: string, year: number): string => {
+  const seasonToSemesterMap = new Map([
+    ['Fall', 'FA'],
+    ['Spring', 'SP'],
+    ['Winter', 'WI'],
+    ['Summer', 'SU'],
+  ]);
+
+  return `${seasonToSemesterMap.get(season)}${year - 2000}`;
+};
```
zachary0kent, line 83:
```suggestion
const seasonAndYearToRosterIdentifier = (season: FirestoreSemesterSeason, year: number): string => {
  const seasonToSemesterMap = {
    Fall: 'FA',
    Spring: 'SP',
    Winter: 'WI',
    Summer: 'SU'
  } as const;

  return `${seasonToSemesterMap[season]}${year - 2000}`;
};
```
This will allow the type system to ensure the lookup doesn't fail

rohanmaheshwari430:
> 

I originally wrote these functions with the Courses Interface and Migration project in mind. I think we should add other retrieval methods on a need basis. Especially if we aren't even going to use those functions at the moment. In my opinion, this is simplest approach as we will find out what retrieval methods we need to add when we start to mess with other parts of the courses infrastructure.

zachary0kent:
> > 
> 
> I originally wrote these functions with the Courses Interface and Migration project in mind. I think we should add other retrieval methods on a need basis. Especially if we aren't even going to use those functions at the moment. In my opinion, this is simplest approach as we will find out what retrieval methods we need to add when we start to mess with other parts of the courses infrastructure.

Makes sense!

src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,83 @@
...
+ * @returns `Promise<CornellCourseRosterCourseFullDetail>`
+ */
+export const getCourseWithCrseIdAndRoster = async (
+  roster: string,
+  crseId: number
+): Promise<CornellCourseRosterCourseFullDetail> => {
+  // use crseId to retrieve course subject and code
+  const courseSubjectAndNumber = await getDoc(doc(crseIdToCatalogNbrCollection, `${crseId}`));
+  const subject = courseSubjectAndNumber.data()?.split(' ')[0];
+  const number = courseSubjectAndNumber.data()?.split(' ')[1];
```
zachary0kent, line 53: Can we factor out this logic into a different function and also add error checking for the case when `courseSubjectAndNumber.data()?.split(' ')` has length < 2? We should avoid having this promise be rejected with say, an index out of bounds error. Maybe we could throw a more descriptive error instead?

src/global-firestore-data/courses.ts
rohanmaheshwari430: I do agree with what you are saying, but don't we already expose this since the BottomBar says "SP23 Lecture Information"?

Screen Shot 2023-03-25 at 4 54 37 PM

Like in this case. Let me know if what you are talking about is something different!

zachary0kent: Ah I see now, nevermind.

rohanmaheshwari430: For now, I will remove `getCourseWithCrseIdAndRoster()` and we can re-implement it when we need it! For Pablo's work this week, I don't think it will be necessary.

src/global-firestore-data/courses.ts
rohanmaheshwari430: Will note this down for when it is implemented in the future.

zachary0kent: Sounds good. For now, can we just make sure to throw a more descriptive error in this case? Of course, we can handle its propagation later.

src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,50 @@
...
+      await getDoc(doc(availableRostersForCoursesCollection, `${subject} ${number}`))
+    ).data() as { rosters: string[] };
+    const lastRoster = availableRostersForCourse.rosters.length - 1;
+    const latestCourse = await getDoc(
+      doc(
+        coursesCollection,
+        `${availableRostersForCourse.rosters[lastRoster]}/${subject}/${number}`
+      )
+    );
+    return latestCourse.data()?.course as CornellCourseRosterCourseFullDetail;
```
zachary0kent, line 31: Could we refactor this into a different function, like `getLastestOffering` or the like? I imagine this will be useful in other places as well.

zachary0kent (commented):
Looks good! Just one nit.

src/global-firestore-data/courses.ts
rohanmaheshwari430: For sure

src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,51 @@
...
+  year: number,
+  subject: string,
+  number: string
+): Promise<CornellCourseRosterCourseFullDetail> => {
+  const roster = seasonAndYearToRosterIdentifier(season, year);
+  const course = await getDoc(doc(coursesCollection, `${roster}/${subject}/${number}`));
+  if (!course.exists()) {
+    return getLastOffering(subject, number);
+  }
+  return course.data()?.course as CornellCourseRosterCourseFullDetail;
```
zachary0kent, line 23: Is this cast needed after annotating the return type?

rohanmaheshwari430: While we annotated the return type of the function, we never actually asserted/checked if the object being returned actually matches that type.  In fact, Typescript thinks the type is `any`, which makes sense because it can't possibly infer the type of what the Firestore is returning. 

Screen Shot 2023-03-25 at 11 27 11 PM

So I just like to use type assertion before returning as a sanity check.

zachary0kent: Does removing the cast produce a type error? Casts in typescript don't actually perform any runtime validation.

rohanmaheshwari430: It doesn't produce a type error, but is specifying the return type of the function enough? `course.data()?.course` is of type `DocumentData`, but our function returns `CornellCourseRosterCourseFullDetail`. This means we are just hoping that `DocumentData === CornellCourseRosterCourseFullDetail`, without checking at all.

zachary0kent: Yeah like what I'm saying is that typescript casts are just straight up removed during compilation to javascript, so even if `DocumentData === CornellCourseRosterCourseFullDetail`, the cast wouldn't fail.

rohanmaheshwari430: Gotcha, then I will just define `course` with type `CornellCourseRosterCourseFullDetail`

rohanmaheshwari430: Wait I can't do that either, nvm ill just nuke the type cast

zachary0kent (commented):
Also can we reduce the duplication between `getCourseWithSeasonAndYear` and `getCourseWithSeasonAndYear`?

src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @param courseCode The course code (EX: 'CS 1110')
+ * @returns A string[] containing the subject and number (EX: ['CS', '1110'])
+ */
+const extractSubjectAndNumber = (courseCode: string): string[] => {
+  if (courseCode.split(' ').length !== 2) {
+    throw Error(`Invalid course format. Expected courseCode.split(' ') === 2`);
+  } else {
+    return courseCode.split(' ');
+  }
```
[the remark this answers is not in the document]
rohanmaheshwari430: Betski

src/global-firestore-data/courses.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @returns A string[] containing the subject and number (EX: ['CS', '1110'])
+ */
+const extractSubjectAndNumber = (courseCode: string): { subject: string; number: string } => {
+  if (courseCode.split(' ').length !== 2) {
+    throw Error(
+      `Invalid course format. Expected course to be of form subject and number. EX: CS 1110`
+    );
+  } else {
+    return { subject: courseCode.split(' ')[0], number: courseCode.split(' ')[1] };
+  }
```
zachary0kent, line 111:
```suggestion
  const subjectAndNumber = courseCode.split(' ');
  if (subjectAndNumber.length !== 2) {
    throw Error(
      `Invalid course format. Expected course to be of form subject and number. EX: CS 1110`
    );
  } else {
    return { subject: subjectAndNumber[0], number: subjectAndNumber[1] };
  }
```

zachary0kent (approved):
Left one nit, but otherwise looks good!

PR #807, opened 2023-03-29
Validate parsed courses

mirandayu131:
Find the unique course ID of every course parsed from the other Approved Liberal Studies course list since some courses may be cross-listed. For example, both CS 2110 and ENGRD 2110 would have the same course ID. Courses not found in the course roster data are excluded.

Test plan:
- Check that courses excluded are not in the full list of courses

### DIFF
--- a/scripts/other-yes-list/parse.ts
+++ b/scripts/other-yes-list/parse.ts
@@ -1,0 +1,92 @@
+import { fullCoursesArray } from '../../src/assets/courses/typed-full-courses';
+
+export type Course = {
+  /** Corresponds to the "prefix" column on the PDF */
+  subject: string;
+  /** Corresponds to the "code" column on the PDF */
+  catalogNbr: number;
+  /** The liberal studies categories satisfied by the course */
+  categories: readonly string[];
+};
+
+export type CourseWithId = Course & {
+  crseId: number;
+};
+
+enum Col {
+  CoursePrefix,
+  CourseNumber,
+  Title,
+  College,
+  Category,
+  Credits,
+}
+
+// creates list of course entries
+export function genCourseEntries(data) {
+  const courseEntries: Course[] = [];
+  let c: Col = Col.CoursePrefix as Col;
+  let subject = '';
+  let catalogNbr = 0;
+  let categories = [];
+
+  // find the index of the first course (right after Notes)
+  let start = 0;
+  for (let i = 0; i < data.length; i += 1) {
+    const v = data[i][0].str.trim();
+    if (v.slice(0, 5) === 'Notes') {
+      start = i + 1;
+      break;
+    }
+  }
+  // assumes that the columns are ordered and there are no empty cells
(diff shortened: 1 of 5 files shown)


scripts/other-yes-list/parse.ts
```diff
@@ -81,3 +82,28 @@ export function genCourseEntries(data) {
...
+        courseWithId.catalogNbr === course.catalogNbr.toString()
+      ) {
+        const x: CourseWithId = {
+          subject: course.subject,
+          catalogNbr: course.catalogNbr,
+          categories: course.categories,
+          crseId: courseWithId.crseId,
+        };
+        validCourses.push(x);
+        break;
```
rohanmaheshwari430, line 104: could there ever be more than one `course` such that `courseWithId.subject === course.subject && courseWithId.catalogNbr === course.catalogNbr.toString()` is true? because by breaking after you find a match, you are saying that you are only looking for the first occurrence where that `if` condition passes.

rohanmaheshwari430 (commented):
nice work miranda!

scripts/other-yes-list/parse.ts
```diff
@@ -81,3 +82,28 @@ export function genCourseEntries(data) {
...
+export function getValidCourses(data) {
+  const courseEntries = genCourseEntries(data);
+  const approvedCourses = fullCoursesArray;
+
+  const validCourses: CourseWithId[] = [];
+  for (const courseWithId of approvedCourses) {
+    for (const course of courseEntries) {
+      if (
+        courseWithId.subject === course.subject &&
+        courseWithId.catalogNbr === course.catalogNbr.toString()
```
rohanmaheshwari430, line 95: could there ever be more than one `course` such that `courseWithId.subject === course.subject && courseWithId.catalogNbr === course.catalogNbr.toString()` is true? because by breaking after you find a match, you are saying that you are only looking for the first occurrence where that `if` condition passes.

zachary0kent (changes requested):
Great work! Can you change the base branch to `master` and wait until `parse-lib-studies` is merged to merge this?

scripts/other-yes-list/parse.ts
```diff
@@ -81,3 +82,28 @@ export function genCourseEntries(data) {
...
+          subject: course.subject,
+          catalogNbr: course.catalogNbr,
+          categories: course.categories,
+          crseId: courseWithId.crseId,
+        };
+        validCourses.push(x);
+        break;
+      }
+    }
+  }
```
zachary0kent, line 107:
```suggestion
  for (const { subject, catalogNbr, crseId } of approvedCourses) {
    for (const course of courseEntries) {
      if (subject === course.subject && catalogNbr === course.catalogNbr.toString()) {
        validCourses.push({ ...course, crseId });
        break;
      }
    }
  }
```

zachary0kent (changes requested):
Can you merge master into your branch? Then I think this should be good to merge :)

PR #811, opened 2023-04-14
Fetch Approved Liberal Studies Courses

mirandayu131:
Created a script that will (1) fetch the Other Yes List from its source, (2) parse it into a `CourseWithId[]`, and (3) write these courses to a JSON file to inspect later. 

Test Plan
- Test URL is https://apps.engineering.cornell.edu/liberalstudies/ApprovedLiberalStudies.pdf


elizabeth-tang (commented):
Great job Miranda!! I'm curious about your use of try catch with an empty catch in parse.ts. Just a thought, what would it mean if an exception does occur during genCourseEntries and would you possibly want to modify/rerun without the offending part of the data?

package.json
```diff
@@ -69,6 +69,8 @@
     "eslint-plugin-vue": "^8.0.3",
     "jest": "^26.6.3",
     "minimist": "^1.2.6",
+    "pdf-to-excel": "^0.3.2",
+    "pdfjs-dist": "^3.4.120",
```
zachary0kent, line 73: Do we use this dependency anywhere? If not, can we delete it?

zachary0kent (changes requested):
Looks good, except one nit!

zachary0kent (changes requested):
Also, can we modify the script to put the json in `src/assets/courses`? That's where our other json's with course information are located.

package.json
mirandayu131: We use pdf-to-excel but not pdfjs-dist so I removed that

zachary0kent (changes requested):
Can we also move the existing json to `src/assets/courses`? Thanks!

zachary0kent (approved):
Looks good!

PR #812, opened 2023-04-14
Set up NCF Model for Recommendation System

jerrry1123:
Summary
Added the code to train and evaluate a NCF model with given CSV files
Added code to fetch a sample dataset into the appropriate CSV files

Remaining TODOs:
Generate the appropriate CSVs from course data
Turn evaluation of individual users into an API format so that Courseplan can call it
Conduct testing on model hyperparameters
Incorporate pre-training to increase accuracy - see (https://github.com/microsoft/recommenders/blob/main/examples/02_model_collaborative_filtering/ncf_deep_dive.ipynb)
Improve the metrics with the above methods (right now they're not great, but I can add more optimizations once we're sure the basics work)

Test Plan

IMPORTANT: Follow the installation instructions here: https://docs.google.com/document/d/1e53btTwYI6-3LAB2mH4aE1FpmUJ-crL6I2vPUfmjGmg/edit
Once done, first run SampleDatasetFetcher.py to download the training and testing data, and then NCFTest.py through the terminal and make sure that the terminal provide feedback saying that training and predictions are finished. Training could take a while, so be patient, but predictions shouldn't take more than a minute or so. Please make sure this works on your system and Slack me if you run into any issues during installation or running the code so I can add fixes into the doc above.


zachary0kent (commented):
What parts of this code eventually need to reside in the repo for recommendations? I want to avoid merging proof of concept code into master if possible.

mirandayu131 (approved):
I used your model on the training data and it worked well for me. Prediction rates were very high

jerrry1123:
> What parts of this code eventually need to reside in the repo for recommendations? I want to avoid merging proof of concept code into master if possible. Maybe it would be better to do all recommendations work on one branch and then merge that branch into master when it's complete?

I'll set up a new branch with this code then and then merge it into rec-sys in a new PR.

jerrry1123:
On second thought, it is likely that we'll need the firebase script and some of the data preprocessing at a minimum inside our repo, so I think it's probably best to leave it in master and then remove whatever we don't need later since I'm not really sure how much we'll need to keep at this moment.

zachary0kent:
> On second thought, it is likely that we'll need the firebase script and some of the data preprocessing at a minimum inside our repo, so I think it's probably best to leave it in master and then remove whatever we don't need later since I'm not really sure how much we'll need to keep at this moment.

This seems like a good reason to do recommendations work on a different branch until it's complete no?

jerrry1123:
> > On second thought, it is likely that we'll need the firebase script and some of the data preprocessing at a minimum inside our repo, so I think it's probably best to leave it in master and then remove whatever we don't need later since I'm not really sure how much we'll need to keep at this moment.
> 
> This seems like a good reason to do recommendations work on a different branch until it's complete no?

done!

zachary0kent (approved):
Looks good!

PR #814, opened 2023-04-14
Added script to fetch user data into a usable CSV format

jerrry1123:
Summary

Added a script to get the user-course pairings in the CSV using the userID and courseID
Added options in the dataset processor to either use the dataset fetched this way or the MovieLens dataset

    movielens will mostly be used to evaluate changes in the model

Remaining TODOs:

Link the ML training and predictions to cloud

Test Plan

Check that the CSV generated makes sense with userIDs and courseIDs, no unexpected behaviour.


zachary0kent (changes requested):
I don't think it's a good idea (or necessary) to associate each user with a unique integer id in the Firestore. Could we hash the user's email instead, perhaps?

scripts/export-user-data.ts
```diff
@@ -0,0 +1,69 @@
...
+class Rating {
+  userId: number;
+
+  courseId: number;
+
+  rating: number;
+
+  timestamp: number;
+
+  constructor(id: number, course: number, rating: 1, timestamp: 0) {
```
zachary0kent, line 13: Why does rating have type `1` and timestamp have type `0`?

jerrry1123:
> I don't think it's a good idea (or necessary) to associate each user with a unique integer id in the Firestore. Could we hash the user's email instead, perhaps?

Just resolved both issues you brought up!

scripts/add-user-id.ts
zachary0kent, line 1: Can we remove this migration script?

src/user-data.d.ts
```diff
@@ -84,6 +84,7 @@ type FirestoreOverriddenFulfillmentChoices = {
 };
 
 type FirestoreUserData = {
+  readonly id: number;
```
zachary0kent, line 87:
```suggestion
```

zachary0kent (changes requested):

scripts/export-user-data.ts
```diff
@@ -0,0 +1,68 @@
+import fs from 'fs';
+import { semestersCollection } from './firebase-config';
+
+class Rating {
```
zachary0kent, line 4: Do you think making this a class is useful if it only has one method? Or are you planning on extending it with more in the future?

scripts/export-user-data.ts
```diff
@@ -0,0 +1,68 @@
...
+  rating: number;
+
+  timestamp: number;
+
+  constructor(id: number, course: number, rating: number, timestamp: number) {
+    this.userId = id;
+    this.courseId = course;
+    this.rating = rating;
+    this.timestamp = timestamp;
+  }
```
zachary0kent, line 18: I think having 4 unnamed numeric arguments is confusing. Could we pass in an object instead?

scripts/export-user-data.ts
```diff
@@ -0,0 +1,68 @@
...
+  const userCourseIDs: number[] = [];
+  const semesters = user.get('semesters');
+  for (const semester of semesters) {
+    const { courses } = semester;
+    for (const course of courses) {
+      userCourseIDs.push(course.crseId);
+    }
+  }
+  return userCourseIDs;
+}
```
zachary0kent, line 42:
```suggestion
  const userCourseIDs: number[] = [];
  const semesters = user.get('semesters');
  for (const { courses } of semesters) {
    for (const { crseId } of courses) {
      userCourseIDs.push(crseId);
    }
  }
  return userCourseIDs;
}
```

scripts/export-user-data.ts
```diff
@@ -0,0 +1,68 @@
...
+  const userSnapshot = await semestersCollection.get();
+  createCSVHeaders();
+  let id = 0;
+  for (const user of userSnapshot.docs) {
+    const courses = getUserCourses(user);
+    for (const course of courses) {
+      if (course !== undefined) {
+        const rating = new Rating(id, course, 1, 0);
+        rating.saveAsCSV();
+      }
```
zachary0kent, line 62: Why would `course` be `undefined`?

jerrry1123: I saw some examples when exporting the data with empty courses in the CSV, so I added the check just in case

zachary0kent (changes requested):
For now, could we use a cryptographic hash like SHA2 to generate a (almost certainly) unique integer for each user? It is more likely Earth will experience an extinction event than we are to encounter a SHA2 collision.

PR #815, opened 2023-04-16
Multiple plans backend

elizabeth-tang:
### Summary 

This pull request will support the backend for the multiple plans feature

- Add new data type: Plan that consists of a name for the plan and an array of semesters
- Update Vuex to keep track of plans and the current plan
- Add Firestore functions and Vuex mutations to support adding, deleting and modifying plans
- Wrote a migration script to update the plans field in user-semesters collection to reflect the new formatting of plans

### Test Plan 
- Would love some guidance here -- how does one actually test this 😢


zachary0kent (changes requested):
Overall looks good! Could you split this into two PRs--one for the migration script, and another for the Firestore functions?

PR #816, opened 2023-04-16
Multiple plans functions

elizabeth-tang:
Summary
This pull request will support the backend for the multiple plans feature

- Add new data type: Plan that consists of a name for the plan and an array of semesters
- Update Vuex to keep track of plans and the current plan
- Add Firestore functions and Vuex mutations to support adding, deleting and modifying plans

### Test Plan 
- Unit tests using Jest
- Eventually connecting to the frontend after modals are created

PR #817, opened 2023-04-16
migration script for multiple plans in firebase

elizabeth-tang:
### Summary 

- Wrote a migration script to support new data type Plan
- Associates all current plans with default name "Plan 1"

### Test Plan 

- Test with dev courseplan firebase to see if in user-semesters, plans:
     -  is an array with one plan
     -  the plan has name "Plan 1"
     -  the plan has all old semesters in the semesters field.

### DIFF
--- a/scripts/migration/plans-migration.ts
+++ b/scripts/migration/plans-migration.ts
@@ -1,0 +1,32 @@
+/* eslint-disable no-console */
+
+import { usernameCollection, semestersCollection } from '../firebase-config';
+
+/**
+ * Perform migration of semester to plans with a list of semesters
+ */
+async function runOnUser(userEmail: string) {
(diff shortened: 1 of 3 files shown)


jerrry1123:
Everything looks good and shoudl work properly. Just one thing - I don't know if it's necessary to keep the console log statements still, and if we do want to keep it, would it be possible to follow the eslint instructions? I found a website with how to do it: https://eslint.org/docs/latest/rules/no-await-in-loop

zachary0kent (approved):
Looks good! Ran the migration script on my account and everything works as expected.

zachary0kent (changes requested):
Actually, should we move the change where you added the `Plan` type in #816 and to this PR instead?

PR #819, opened 2023-04-24
Populate VuexStore from Course Data

PabloRaigoza:
### Summary 

This pull request handles added a function `popSemCourseData ` which repopulates courses with information from course data. In addition, I added function for a VuexAction to handle modify the VuexStore

- [x] Implemented functionality for handling VuexStore repopulation given list of `FirestoreSemester`

Remaining TODOs:

- [ ] Need to add functionality to replace old state modification of the VuexStore to new VuexAction

Depends on #805

### Test Plan 

- Unit tests is making sure the VuexStore is being repopulated correctly once replacing previous mutation calls with the VuexAction calls
- Creating a `FirestoreSemester` and making sure the courses within it are repopulated correctly with information from course data.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -355,7 +355,6 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
     if (data) {
       const semesters = getFirstPlan(data);
       const { orderByNewest } = data;
-      console.log('here');
       store.commit('setSemesters', semesters);
       updateDoc(doc(fb.semestersCollection, simplifiedUser.email), {
         plans: [{ semesters }], // TODO: andxu282 update later
--- a/src/store.ts
+++ b/src/store.ts
@@ -17,6 +17,7 @@ import {
 } from './utilities';
 import featureFlagCheckers from './feature-flags';
 import { setUserProperties } from './gtag';
+import getCourse from './global-firestore-data/courses';
 
 type SimplifiedFirebaseUser = { readonly displayName: string; readonly email: string };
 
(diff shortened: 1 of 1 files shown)


src/store.ts
```diff
@@ -32,6 +33,64 @@ type DerivedCoursesData = {
...
+  for (let i = 0; i < sem.courses.length; i += 1) {
+    const aCourse = sem.courses[i];
+    if (isFirestoreSemesterCourse(aCourse)) {
+      promises.push(popCourseCourseData(aCourse, sem));
+    }
+  }
+  const results = await Promise.all(promises);
+  for (const result of results) {
+    newCourses.push(result);
+  }
```
zachary0kent, line 86: Is there any need for `newCourses`? Can we just use `results`? Nice use of `Promise.all` btw!

src/store.ts
```diff
@@ -32,6 +33,64 @@ type DerivedCoursesData = {
...
+
+async function popSemCourseData(sem: FirestoreSemester): Promise<FirestoreSemester> {
+  const promises = [];
+  const newCourses = [];
+  for (let i = 0; i < sem.courses.length; i += 1) {
+    const aCourse = sem.courses[i];
+    if (isFirestoreSemesterCourse(aCourse)) {
+      promises.push(popCourseCourseData(aCourse, sem));
+    }
+  }
```
zachary0kent, line 82: Can we use `Array.map` here instead of a `for` loop?

src/store.ts
```diff
@@ -97,7 +156,18 @@ const store: TypedVuexStore = new TypedVuexStore({
     uniqueIncrementer: 0,
     isTeleportModalOpen: false,
   },
-  actions: {},
+  actions: {
+    setSemesters(context, semester: readonly FirestoreSemester[]) {
+      const newSemesters = [];
+      for (let i = 0; i < semester.length; i += 1) {
+        const curSemester = semester[i];
+        newSemesters.push(popSemCourseData(curSemester));
```
zachary0kent, line 164: Hm, `popSemCourseData` returns a `Promise<FirestoreSemester>` right? We should probably `await` this so we don't end up with an array of promises. Specifically, right now `newSemesters` has type `Promise<FirestoreSemester>[]`, but we want it to have type `FirestoreSemester[]`.

zachary0kent (changes requested):
Nice work adding our first Vuex action! Left a couple comments about `await`ing on promises.

mirandayu131 (approved):
Nice improvements on using promises instead. I like how you used Vuex handlers like subscribe and dispatch to populate courses.

mirandayu131 (changes requested):
Need to rerun npm tests?

zachary-kent:
@PabloRaigoza Can you try to get the tests passing here?

PR #820, opened 2023-04-26
Populate Firestore with specific roster's courses

rohanmaheshwari430:
### Summary 

In this PR, we write a script that will populate the `courses` collection with the courses of the latest roster. It will also update `available-rosters-for-course` collection by adding the latest roster to the rosters array. Finally, it will update the `crseid-to-catalognbr` collection for any courses whose crseId may have changed in the latest roster (unlikely but still a good measure to have). 

### Test Plan 

Performed checks using Firestore's query builder to ensure that each modified collection contained the appropriate data and that their schema was in line with existing data. Checked various courses and ensured that their offerings were in line with what is in the firestore.


zachary0kent (changes requested):
Is there any way we can refactor this to that the original population script calls the functions we added here?

scripts/population/populate-specific-roster.ts
```diff
@@ -0,0 +1,108 @@
...
+      `There are ${subjectCourses.length} courses in ${subject.subject} in ${subject.roster}.`
+    );
+    console.log(`We fetched ${subjectCount} out of ${rosterSubjectPairs.length} subjects`);
+  }
+
+  console.log(`There are ${coursesCount} total courses.`);
+};
+
+/* Enter the desired roster as a string argument (e.g 'FA21'). Fetches latest roster data by default */
+populate();
```
zachary0kent, line 108:
```suggestion
import parseArgs from minimist 

/* Enter the desired roster as a string argument (e.g 'FA21'). Fetches latest roster data by default */
populate(parseArgs(process.argv.slice(2)).roster);
```

zachary0kent: This way we can pass the roster in as a command line argument (we should also move the import to the top of the file)

zachary0kent (changes requested):
Just one suggestion!

scripts/population/populate-specific-roster.ts
rohanmaheshwari430: BETSKI

zachary0kent (approved):
Looks good!

PR #822, opened 2023-04-27
Newsletter Email Script

zachary0kent:
### Summary 

Implements a script that sends our newsletter to every user in the Firestore (or the users specified on the command line, if provided).

### Test Plan 

Unfortunately, y'all don't have the credentials needed to run the script. But I can definitely send a newsletter using the script to any skeptical reviewers!


noschiff (approved):
Amazing work, Zak! Thanks for taking the time to help our product and business team!
One small question -- is it necessary to keep the html in the repo?

zachary0kent:
> Amazing work, Zak! Thanks for taking the time to help our product and business team! One small question -- is it necessary to keep the html in the repo?

Probably not--I'm refactoring the script now so that the HTML source is a command line argument and the newsletter itself is untracked. I didn't realize how large it was until I saw the diff haha.

zachary0kent:
Unfortunately I think we'll far exceed our quota trying to email every user. I think we'll probably have to use something like mailchimp at this point.

PR #828, opened 2023-05-03
Implemented Clean Function

zachary0kent:
### Summary 

Implements a new and improved function to clean NBS while maintaining the structure of an object.

### Test Plan 

Jest


rohanmaheshwari430 (approved):
can u expand on what NBS is for new readers, otherwise LGTM!

PR #834, opened 2023-05-09
Update BottomBarCourseInfo.vue

rohanmaheshwari430:
### Summary 

Fix the class roster URL when a user clicks the "View Course Information on Roster" link in BottomBar. 

### Test Plan 

Visited the updated link for courses in two scenarios: 

1. Course is offered in the chosen semester, and the link takes user to the correct roster link.
2. Course is not offered in the chosen semester, and the link takes the user to the latest roster link.


zachary0kent (approved):
Looks good! Clicking on the link to a course's roster entry redirects us to the correct roster.

zachary0kent (approved):
Looks good! Clicking on the link to a course's roster entry redirects us to the correct roster.

PR #843, opened 2023-08-27
Implement 'serve:prod' Script

zachary-kent:
### Summary 

Implements a 'serve:prod' script, allowing devs to build and test CoursePlan locally for production.

### Test Plan 

Run `npm run serve:prod` and check that you can view your production data on CoursePlan.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "serve": "vite",
+    "serve:prod": "vite --mode production",
     "build": "vite build --mode production",
     "build:dev": "vite build --mode development",
     "serve:build": "serve --listen 8080 --single dist",


package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "serve": "vite",
+    "serve:prod": "vite --mode production",
```
rohanmaheshwari430, line 7: Is it possible that devs may accidentally modify the production data by spinning up a local build that uses production data? 

I don't suspect this happening since only the owner of the data can modify it (i.e the email check in the read/write rules in firestore), but I wanted to ask to make sure its been considered!

rohanmaheshwari430 (commented):
Could devs modify production data by building locally via this new command?

package.json
zachary-kent: Yeah it definitely is possible that someone could nuke their own prod data like this, they won't be able to modify anyone else's.

rohanmaheshwari430: Awesome!!

rohanmaheshwari430 (approved):

PR #845, opened 2023-08-28
Modify Workflows to Reference `main` as Default Branch

zachary-kent:
### Summary 

Updates workflows to refer to the default branch as `main`; after this is merged I will update the default branch in settings and delete `master`.

### DIFF
--- a/.github/workflows/ci-policies.yml
+++ b/.github/workflows/ci-policies.yml
@@ -10,10 +10,10 @@ jobs:
         run: |
           echo "Your head ref is ${{ github.head_ref }}."
           echo "Your base ref is ${{ github.base_ref }}."
-      - name: Fail if try to push release from non-master branch
-        if: ((github.base_ref == 'beta-release' || github.base_ref == 'release') && github.head_ref != 'master') && (github.event.pull_request.user.login != 'noschiff')
+      - name: Fail if try to push release from non-main branch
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-policies.yml
```diff
@@ -10,10 +10,10 @@ jobs:
         run: |
           echo "Your head ref is ${{ github.head_ref }}."
           echo "Your base ref is ${{ github.base_ref }}."
-      - name: Fail if try to push release from non-master branch
-        if: ((github.base_ref == 'beta-release' || github.base_ref == 'release') && github.head_ref != 'master') && (github.event.pull_request.user.login != 'noschiff')
```
rohanmaheshwari430, line 14: Does removing `github.event.pull_request.user.login != 'noschiff'` mean that anyone can merge the pull request into release now?

rohanmaheshwari430: I don't have any opinions on it, was just wondering what removing this line does.

rohanmaheshwari430 (approved):

.github/workflows/ci-policies.yml
zachary-kent: Removing this actually makes the check more conservative; that is, there are no cases where the failure would trigger before but not now.

PR #846, opened 2023-09-12
Migrate Other Yes Categories

mirandayu131:
The task is to add the categories for every course on the liberal studies other yes list to the course data collection. Currently, the type CornellCourseRosterCourseFullDetail records all of the information we maintain about a course in the course data collection. I added an optional property otherYesCategories, a string array that contains the categories associated with a course in the other yes list.

Test Plan

Need to use Firestore Jest


zachary-kent (commented):
Even if we can't use unit testing to verify this, can you still include an ad-hoc test plan?

zachary-kent (changes requested):
Can you add a test plan, even it's more ad-hoc than unit testing?

PR #849, opened 2023-09-20
Python CI + Black Pre-commit Hook

andxu282:
### Summary 
Adds python tests to the CI that includes
- [x] running pytest
- [x] running black linter to ensure no formatting errors
- [x] precommit hook for running black
- [x] ensure the `Pipfile` and `Pipfile.lock` are not out of sync

### Test Plan 
The CI Tests pass so I can merge this damn thing lol. Also tested the pre-commit hook by trying to commit something that was not formatted and seeing it reformatted
Screenshot 2023-09-21 at 5 55 56 PM

Push out of date pipfile to see if it's broken
Screenshot 2023-09-28 at 6 54 57 PM

PR #850, opened 2023-09-20
Add Pre-commit Hooks

andxu282:
### Summary 
Adds two pre-commit hooks to run prettier for JS files and black for python files before each commit to automatically format your code! Will add some documentation to the notion on how to set this up and any issues you might run into. There's a lot of changes due to upgrading prettier to v3.0.3

This pull request is the first step towards implementing feature Foo

- [x] Pre-commit hooks babyyyy
- [x] ~Need to fix the prettier, it's formatting things differently~ FIXED!

Depends on (https://github.com/cornell-dti/course-plan/pull/849)

### Test Plan 

Screenshot 2023-09-20 at 9 49 58 AM


zachary-kent (changes requested):
Awesome! Can you update prettier in a separate PR into `main` first so that the diff is easier to read here? Also, do you think it would be feasible to introduce precommit hooks into the codebase independent of your python work -- specifically, can we open a PR into `main` first that just adds a precommit hook for prettier, and then add the black precommit hook with your python work?

PR #851, opened 2023-09-20
Bump Prettier to 3.0.3

andxu282:
### Summary 
Bumps prettier to latest version 3.0.3

### Test Plan 
Ensure CI tests pass after running `npm run format`

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "jest": "^26.6.3",
     "minimist": "^1.2.6",
     "pdf-to-excel": "^0.3.2",
-    "prettier": "2.2.1",
+    "prettier": "3.0.3",
     "sass": "^1.54.5",
     "sass-loader": "^13.0.2",
     "serve": "^14.1.2",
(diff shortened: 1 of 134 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "jest": "^26.6.3",
     "minimist": "^1.2.6",
     "pdf-to-excel": "^0.3.2",
-    "prettier": "2.2.1",
+    "prettier": "3.0.3",
```
rohanmaheshwari430, line 76: Can you add `^` to the version so that `npm install` will only permit minor patch updates and not major patch updates to the prettier package? More on this [here](https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json).

rohanmaheshwari430 (approved):
LGTM! Thanks for adding this in! 🙏🏼

zachary-kent (changes requested):
Ah, sorry if I was unclear in my original comment. I was suggesting that we open a PR just for updating prettier, without adding pre-commit hooks. This way, the large diff from the prettier reformat would not clutter meaningful diff.

andxu282:
Oh yes, sorry, that makes total sense 🤦‍♂️ lol, I'll fix that!

PR #854, opened 2023-09-21
Add Prettier Pre-Commit Hook

andxu282:
### Summary 
Adds a pre-commit hook that runs prettier

### Test Plan 

Make some changes to ts file and make changes, see fixes when committing
image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.2.1
+    hooks:
+      - id: prettier
+        exclude: 'src/requirements/decorated-requirements.json'


zachary-kent (changes requested):
I'd like the behavior of the precommit hook to be consistent with `npm run format`, if possible. For example, the following code is fixed by `npm run format`, but not modified by the precommit hook:

```
export type ExamFulfillmentBase =
  {
    readonly courseId: number;
    readonly courseEquivalents?: Record<string, number[]>;
    readonly credits: number;
    readonly majorsExcluded?: string[];
  };
```

andxu282:
Wait I couldn't replicate this. Also, I'm assuming if there are discrepancies between the two, it's due to `.editorconfig`, which there doesn't seem to be a way for the pre-commit hook to follow https://stackoverflow.com/questions/71503396/how-to-use-editorconfig-for-code-formatting-in-git-pre-commit-hook

zachary-kent (approved):
Awesome, looks good to me!

PR #857, opened 2023-10-01
Refactor semesters to plans

elizabeth-tang:
### Summary 

Remove all uses of semesters in Vuex store and in Firebase data
- Replace references to semesters field in VuexStoreState with new schema using the new plan data type.
- Address edge case with rendering before Vuex store is set.

Remaining TODOs:

- Stub in rest of multiple plans functionality

### Test Plan 

- After running migration script on all users/just your username if testing, does the app load with 1 plan?
- Does adding a new plan show up for you on Firebase?
- Does adding a new semester or classes to a semester update the correct plan on Firebase?


zachary-kent (changes requested):
Wow, amazing work!! Everything is really coming together. Because the last commit is really independent of the other frontend changes, could you cherry-pick the last commit to a branch based of `main` and then open a PR for that? I'll properly review that PR.

PR #858, opened 2023-10-01
Refactor semesters to plans

elizabeth-tang:
**Summary**
Remove all uses of semesters in Vuex store and in Firebase data
- Replace references to semesters field in VuexStoreState with new schema using the new plan data type.
- Address edge case with rendering before Vuex store is set.

**Remaining TODOs:**

- Stub in rest of multiple plans functionality
**Test Plan**
- After running migration script on all users/just your username if testing, does the app load with 1 plan?
- Does adding a new plan show up for you on Firebase?
- Does adding a new semester or classes to a semester update the correct plan on Firebase?

### DIFF
--- a/src/components/Semester/Semester.vue
+++ b/src/components/Semester/Semester.vue
@@ -278,6 +278,7 @@ export default defineComponent({
       set(newCourses: readonly AppFirestoreSemesterCourseWithRequirementID[]) {
         const courses = newCourses.map(({ requirementID: _, ...rest }) => rest);
         editSemester(
+          store.state.currentPlan,
           this.year,
           this.season,
           (semester: FirestoreSemester): FirestoreSemester => ({
@@ -411,7 +412,14 @@ export default defineComponent({
     addCourse(data: CornellCourseRosterCourse, choice: FirestoreCourseOptInOptOutChoices) {
       const newCourse = cornellCourseRosterCourseToFirebaseSemesterCourseWithGlobalData(data);
       // Since the course is new, we know the old choice does not exist.
-      addCourseToSemester(this.year, this.season, newCourse, () => choice, this.$gtag);
+      addCourseToSemester(
+        store.state.currentPlan,
+        this.year,
+        this.season,
+        newCourse,
+        () => choice,
+        this.$gtag
+      );
 
       const courseCode = `${data.subject} ${data.catalogNbr}`;
       this.openConfirmationModal(`Added ${courseCode} to ${this.season} ${this.year}`);
@@ -429,7 +437,14 @@ export default defineComponent({
         };
 
         // add the course to the semeser (with no choice made)
-        addCourseToSemester(this.year, this.season, newCourse, () => choice, this.$gtag);
+        addCourseToSemester(
+          store.state.currentPlan,
+          this.year,
+          this.season,
+          newCourse,
+          () => choice,
+          this.$gtag
+        );
         this.closeCourseModal();
 
         const conflicts = store.state.courseToRequirementsInConstraintViolations.get(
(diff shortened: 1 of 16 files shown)


src/components/Semester/SemesterView.vue
```diff
@@ -107,7 +107,14 @@ export default defineComponent({
   },
   computed: {
     semesters(): readonly FirestoreSemester[] {
-      return store.state.semesters;
+      console.log(store.state.plans);
```
andxu282, line 110: get rid of console log

src/user-data.d.ts
```diff
@@ -77,15 +77,15 @@ type FirestoreCourseOptInOptOutChoices = {
...
+  readonly arbitraryOptIn: { readonly [requirement: string]: readonly string[] };
 };
 type FirestoreOverriddenFulfillmentChoices = {
   readonly [courseUniqueId: string]: FirestoreCourseOptInOptOutChoices;
 };
 
 type FirestoreUserData = {
   readonly name: FirestoreUserName;
-  readonly semesters: FirestoreSemester[];
+  // readonly semesters: readonlyFirestoreSemester[];
```
andxu282, line 88: delete

src/store.ts
```diff
@@ -283,15 +314,21 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
...
+      store.commit('setPlans', data.plans);
+      store.commit('setCurrentPlan', data.plans[0]);
+
+      const plan = getFirstPlan(data);
       const { orderByNewest } = data;
-      store.commit('setSemesters', semesters);
+      store.commit('setSemesters', plan.semesters);
       updateDoc(doc(fb.semestersCollection, simplifiedUser.email), {
-        plans: [{ semesters }], // TODO: andxu282 update later
+        plans: [plan], // TODO: andxu282 update later
```
andxu282, line 324: can u get rid of comment 🥲

src/store.ts
```diff
@@ -300,7 +337,7 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
       store.commit('setSemesters', [newSemester]);
       setDoc(doc(fb.semestersCollection, simplifiedUser.email), {
         orderByNewest: true,
-        plans: [{ semesters: [newSemester] }], // TODO: andxu282 update later
+        plans: [{ name: 'Plan 1', semesters: [newSemester] }], // TODO: andxu282 update later
```
andxu282, line 340: remove comment : )

src/store.ts
```diff
@@ -76,7 +78,7 @@ const store: TypedVuexStore = new TypedVuexStore({
       tookSwim: 'no',
     },
     orderByNewest: true,
-    semesters: [],
+    // semesters: [],
```
andxu282, line 81: delete

src/store.ts
```diff
@@ -36,7 +36,7 @@ export type VuexStoreState = {
   currentFirebaseUser: SimplifiedFirebaseUser;
   userName: FirestoreUserName;
   onboardingData: AppOnboardingData;
-  semesters: readonly FirestoreSemester[];
+  // semesters: readonly FirestoreSemester[];
```
andxu282, line 39: delete

andxu282 (commented):
Looking good, just some clean up items, I would do a command f for semesters just in case you missed something, did you ever end up fixing that bug btw?

zachary-kent (changes requested):
I think it would probably be easier and cleaner if you modified the `semesters` getter to return `plans[currentPlan].semesters`. That way, accesses to `store.state.semesters` wouldn't have to be changed throughout the codebase.

zachary-kent (changes requested):
Instead of passing the current plan into functions like `addCourseToSemester` and `deleteCourseFromSemesters`, can we import the vuex store and access the current plan directly inside of these functions?

zachary-kent (changes requested):
Looks good! Can you delete the spurious console statements?

src/components/Semester/Semester.vue
```diff
@@ -455,15 +470,28 @@ export default defineComponent({
       this.openConfirmationModal(`Added ${course.code} to ${this.season} ${this.year}`);
     },
     deleteCourseWithoutModal(uniqueID: number) {
-      deleteCourseFromSemester(this.year, this.season, uniqueID, this.$gtag);
+      deleteCourseFromSemester(
+        store.state.currentPlan,
```
zachary-kent, line 474: Why do we need to pass the current plan as an argument?

zachary-kent (changes requested):

src/components/Semester/Semester.vue
elizabeth-gs: I guess there's no case where you aren't deleting from the current plan currently but it was just the way I wrote the firestore functions to modify the plan that is passed in.

zachary-kent (changes requested):
Multiple plans dropdown openclose (failed)
It looks like the test is failing because the dropdown isn't opened after it's clicked.

PR #863, opened 2023-10-20
Parse GPT Output

andxu282:
### Summary 
Parses the GPT Output into a usable dictionary to represent the boolean expression of prerequisites.

- [x] implemented parser for parsing GPT output into dict

Remaining TODOs:

- [x] add test cases
- [x] make type change based on OR or AND clause

### Test Plan 
Will add under python as new unit tests

### DIFF
--- a/python/test_prereq_parse.py
+++ b/python/test_prereq_parse.py
@@ -1,4 +1,8 @@
-from prereq_parse import get_raw_prereqs_and_coreqs, get_prereqs_coreqs
+from prereq_parse import (
+    get_raw_prereqs_and_coreqs,
+    get_prereqs_coreqs,
+    parse_boolean_string,
+)
 import os
 import pytest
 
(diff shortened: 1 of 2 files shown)


zachary-kent (changes requested):
Awesome! Can you delete the commented code and add tests?

PabloRaigoza:
This seems like a very useful step into deploying a more rigorous requirement checking system!
I would only suggest in the test cases adding test cases where certain the the preconditions in your code may not be satisfied. For example 'and CHEM 2080'. Alternatively you can keep asking GPT for responses until preconditions are met.

python/test_prereq_parse.py
```diff
@@ -176,3 +180,174 @@ def test_prereqs_coreqs(index=None, shorten=False, verbose=False, hard=False):
...
+                    },
+                ],
+            },
+        ),
+    ]
+
+    correct = 0
+    total = len(test_cases)
+    for test in test_cases:
+        (boolean_string, answer) = test
```
zachary-kent, line 336:
```suggestion
    for (boolean_string, answer in test_cases:
```

python/test_prereq_parse.py
```diff
@@ -176,3 +180,174 @@ def test_prereqs_coreqs(index=None, shorten=False, verbose=False, hard=False):
...
+        try:
+            response = parse_boolean_string(boolean_string)
+        except:
+            print("PARSING ERROR")
+            continue
+        if response == answer:
```
zachary-kent, line 342:
```suggestion
        if response == answer:
```
Can we do `assert(response == answer)`?

zachary-kent (changes requested):

zachary-kent (approved):
Looks good!

PR #866, opened 2023-10-23
Calculating and Storing Requirement Statistics

PabloRaigoza:
### Summary 
This pull request calculates and stores the data of most common courses to fulfil a given requirement and its slots. It only stores the top 50 most popular courses because storage limitations on Firestore. In addition, this pull request is going to be become important for course recommendations based on most popular courses to fulfill a requirement.

I also cleaned up the code for 'scripts/firebase-config.ts' to be a little more consistant as well as add a new npm command to run most if not all of the cli checks.

- [x] Calculating Requirement Statistics
- [x] Soring Requirement Statistics

### Testing
I tested this script on the dev server by adding a courses on my account, and the changes did appear in Firestore. There are some users that error when attempting to calculate their requirement statistics, by this is likely due to corrupt data in the Firestore.

### DIFF
--- a/scripts/gen-req-full-stats.ts
+++ b/scripts/gen-req-full-stats.ts
@@ -3,91 +3,98 @@ import {
   semestersCollection,
   toggleableRequirementChoicesCollection,
   overriddenFulfillmentChoicesCollection,
+  courseFulfillmentStats,
 } from './firebase-config';
 
 import computeGroupedRequirementFulfillmentReports from '../src/requirements/requirement-frontend-computation';
+import { computeFulfillmentStats } from '../src/requirements/fulfillment-stats';
+import { createAppOnboardingData } from '../src/user-data-converter';
 
-import { RequirementFulfillment, CourseTaken } from '../src/requirement-types.d';
 
-const idRequirementFrequency = new Map<string, Map<number, number>[]>();
+import '../src/requirements/decorated-requirements.json';
 
+let idRequirementFrequency = new Map<string, Map<number, number>[]>();
+let counter = 0;
 /**
  * Prints to console.out the json string of frequencies each course take to
  * fulfill a given requirement
  */
-async function computeRequirementFullfillmentStatistics() {
-  await semestersCollection.get().then(semQuerySnapshot => {
-    // Query a user's semesters
-    semQuerySnapshot.forEach(async doc => {
+async function computeRequirementFullfillmentStatistics(_callback) {
+  const semQuerySnapshot = await semestersCollection.get();
+  semQuerySnapshot.forEach(async doc => {
       // obtain the user's semesters, onboarding data, etc...
-      const semestersAndPlans = doc.data();
-      const onboardingData = await onboardingDataCollection.doc(doc.id).get();
-      const toggleableRequirementChoices = await toggleableRequirementChoicesCollection
-        .doc(doc.id)
-        .get();
-      const overriddenFulfillmentChoices = await overriddenFulfillmentChoicesCollection
-        .doc(doc.id)
-        .get();
+      const semestersAndPlans = await doc.data();
+      const onboardingData = (await onboardingDataCollection.doc(doc.id).get()).data();
+      const toggleableRequirementChoices = (await toggleableRequirementChoicesCollection.doc(doc.id).get()).data() ?? {};
+      const overriddenFulfillmentChoices = (await overriddenFulfillmentChoicesCollection.doc(doc.id).get()).data() ?? {};
 
-      // For the given users semesters compute the requirements they fulfill
-      // and which course it took for them to fulfill them
-      const res = computeGroupedRequirementFulfillmentReports(
-        semestersAndPlans.semester,
-        onboardingData,
-        toggleableRequirementChoices,
-        overriddenFulfillmentChoices
-      );
-
-      res.groupedRequirementFulfillmentReport.forEach(currentGroup => {
-        const currentGroupReqs: RequirementFulfillment[] = currentGroup.reqs;
-
-        // For a given group of requirements, we iterate through all courses
-        // that were needed to fulfill each slot
-        currentGroupReqs.forEach(reqFulfillment => {
-          const key: string = reqFulfillment.requirement.id;
-
-          const fulfillment: readonly CourseTaken[][] = reqFulfillment.fulfillment
-            .safeCourses as readonly CourseTaken[][];
-
-          // Obtain the frequency list for this particular group's requirements
-          const freqList: Map<number, number>[] = idRequirementFrequency.has(key)
-            ? (idRequirementFrequency.get(key) as Map<number, number>[])
-            : [];
-
-          // Iterate over all slots in the requirement group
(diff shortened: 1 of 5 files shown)


scripts/gen-req-full-stats.ts
```diff
@@ -0,0 +1,109 @@
...
+          newOnboardingData,
+          toggleableRequirementChoices,
+          overriddenFulfillmentChoices
+        );
+        idRequirementFrequency = await computeFulfillmentStats(
+          res.groupedRequirementFulfillmentReport,
+          idRequirementFrequency
+        ); // compute fulfillment stats
+      } catch {
+        console.log(counter, ' : Error in computing fulfillment stats for user: ', doc.id);
```
elizabeth-tang, line 53: Delete logging statements/replace with a comment. Instead of printing to console when errors occur, maybe we can set it up to send an email to us? Not sure if this is necessary if the vast majority of computations are fine

scripts/gen-req-full-stats.ts
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  setTimeout(_callback, 3600 * 1000); // wait for all the data to be processed
+}
+
+/**
+ * Stores the computed requirement fulfillment statistics in firestore
+ */
+async function storeComputedRequirementFullfillmentStatistics() {
+  console.log('Keeping only the top fifty courses for each slot');
```
elizabeth-tang, line 66: console again lolz

scripts/gen-req-full-stats.ts
```diff
@@ -0,0 +1,109 @@
...
+  // iterate through all the keys in the idRequirementFrequency hashmap
+  for (const [reqID, slots] of idRequirementFrequency) {
+    const newSlots: Map<number, number>[] = [];
+    for (const slot of slots) {
+      const newSlot = new Map<number, number>();
+      const sorted = [...slot.entries()].sort((a, b) => b[1] - a[1]);
+      let count = 0;
+      for (const [course, freq] of sorted) {
```
elizabeth-tang, line 74: might be cleaner to iterate by index and do index < 50 instead of keeping track of count and breaking when it gets to 50.

scripts/gen-req-full-stats.ts
```diff
@@ -0,0 +1,109 @@
...
+        }
+        newSlot.set(course, freq);
+        count += 1;
+      }
+      newSlots.push(newSlot);
+    }
+    idRequirementFrequency.set(reqID, newSlots);
+  }
+
+  console.log('Storing fulfillment stats in firestore');
```
elizabeth-tang, line 86: change to comment pls!! would also be super helpful for future devs as documentation.

elizabeth-tang (changes requested):
This looks really good Pablo! I just had a few small suggestions about code cleaning and console statements. Excited to get started on "recommendations"!!

scripts/gen-req-full-stats.ts
zachary-kent: I'd disagree with this; I think it would be bad for the script to silently fail

scripts/gen-req-full-stats.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    let i = 0;
+    for (const slot of slots) {
+      const tempJson = {};
+      for (const [course, freq] of slot) {
+        tempJson[course] = freq;
+      }
+      json[i] = tempJson;
+      i += 1;
+    }
```
zachary-kent, line 114: Can you rename `tempJson` and `json` to be more descriptive

src/requirements/fulfillment-stats.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @todo
+ * None
+ */
+export default function computeFulfillmentStats(
+  groups: readonly GroupedRequirementFulfillmentReport[],
+  idRequirementFrequency: Map<string, Map<number, number>[]>
+) {
+  // Iterate over all groups
+  const res = idRequirementFrequency;
```
zachary-kent, line 31: Why do we need `res` here? If we're modifying `idRequirementFrequency` in place, do we have to return it?

scripts/gen-req-full-stats.ts
```diff
@@ -0,0 +1,125 @@
...
+          res.groupedRequirementFulfillmentReport,
+          idRequirementFrequency
+        );
+      } catch {
+        // There was an error computing the fulfillment stats for the user
+      }
+    }
+  });
+
+  setTimeout(_callback, 120 * 1000); // wait 2 minutes before storing the computed stats
```
zachary-kent, line 72: Why do we need to wait here?

scripts/gen-req-full-stats.ts
```diff
@@ -0,0 +1,125 @@
...
+      (await toggleableRequirementChoicesCollection.doc(doc.id).get()).data() ?? {};
+    const overriddenFulfillmentChoices =
+      (await overriddenFulfillmentChoicesCollection.doc(doc.id).get()).data() ?? {};
+
+    // if the user has obboarding data, semesters, etc...
+    if (
+      onboardingData !== undefined &&
+      semestersAndPlans.semesters !== undefined &&
+      toggleableRequirementChoices !== undefined &&
+      overriddenFulfillmentChoices !== undefined
```
zachary-kent, line 47: These conditions will always be true, no? Above, we set these to `{}` if they are `undefined`.

src/requirements/fulfillment-stats.ts
```diff
@@ -0,0 +1,65 @@
...
+      const { safeCourses } = reqFulfillment.fulfillment;
+
+      // Obtain the frequency list for this particular group's requirements
+      const freqList = res.get(key) ?? [];
+
+      // Iterate over all slots in the requirement group
+      // console.log(safeCourses.length);
+      for (let slotNumber = 0; slotNumber < safeCourses.length; slotNumber += 1) {
+        if (freqList.length === slotNumber) {
+          freqList.push(new Map<number, number>());
```
zachary-kent, line 47:
```suggestion
        if (freqList.length === slotNumber) {
          freqList.push(new Map());
```
I think the type parameters can be inferred here, no?

zachary-kent (changes requested):
Just a couple minor nits!

src/requirements/fulfillment-stats.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+        const currentCourseSlot = safeCourses[slotNumber];
+        const currentRequirementSlotFreq = freqList[slotNumber];
+
+        // Iterate over all courses taken to fulfill the req-slot
+        for (let j = 0; j < currentCourseSlot.length; j += 1) {
+          const currentCourseId = currentCourseSlot[j].courseId;
+          const pastFreq = currentRequirementSlotFreq.get(currentCourseId) ?? 0;
+          currentRequirementSlotFreq.set(currentCourseId, pastFreq + 1);
+        }
```
zachary-kent, line 46: Can you convert this into a for each loop?

zachary-kent (changes requested):
Looks mostly good, just a few small comments?

PR #871, opened 2023-11-13
firebase function to auto populate course collection

elizabeth-tang:
### Summary 

This pull request is meant to help automatically populate the course collection every semester instead of manually running the script

- Uses a built-in firebase cloud function

Remaining TODOs:

- Change frequency to every semester -- currently runs every 5 minutes for testing purposes

### Test Plan 
- ad hoc testing -- see if the courses are repopulated every 5 min and logs that it worked to firebase

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "esbuild": "^0.15.5",
         "firebase": "^9.9.3",
         "firebase-admin": "^11.5.0",
+        "firebase-functions": "^4.5.0",
         "intro.js": "^3.3.1",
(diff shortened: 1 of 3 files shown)


zachary-kent (changes requested):
Just to be clear; this is not currently running every five minutes right?

elizabeth-tang:
> Just to be clear; this is not currently running every five minutes right?

Yep :)

zachary-kent (changes requested):
Can you create a separate directory called `functions` with its own `package.json`? You can then install the `firebase-functions` dependency there.

PR #872, opened 2023-11-13
Script to Write Prereqs

andxu282:
### Summary 
Adds firebase config file in python and creates a script to write prereqs to firestore.

Remaining TODOs:
- [ ] Run script on dev
- [ ] Run script on prod

### DIFF
--- a/.github/workflows/ci-check.yml
+++ b/.github/workflows/ci-check.yml
@@ -61,6 +61,7 @@ jobs:
           pip install typing-extensions
           pip install pytest
           pip install --upgrade openai
+          pip install langchain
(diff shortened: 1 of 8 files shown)


zachary-kent (commented):
If it's WIP, you can also open as a draft

PabloRaigoza:
Exciting to see the new implementation for storing pre-req courses into firebase. I think the code you made is great first step to expanding our code base with Python scripts to store, write, and read from firebase. This code will definitely be useful for future scripts to use as reference if we need to use Python again for firebase interactions. Great job!

PR #873, opened 2023-11-13
Implementing Frontend Changes for Recomendations

PabloRaigoza:
### Summary 
This change is the second step to implementing recommendations. #866 implemented changes for storing popular courses for requirements. This PR implements the frontend for displaying course recommendations. The next step is to create calls to firebase to retrieve course recommendations to population.

TODO

- [ ] Make sure that displaying 50 courses (with a scrolling sidebar) is intended design.

### Test Plan 

Given a requirement, use the new frontend to make sure that all courses are present if we are able to query.

### DIFF
--- a/src/components/Requirements/RequirementSideBar.vue
+++ b/src/components/Requirements/RequirementSideBar.vue
@@ -81,29 +81,22 @@
         </div>
         <div class="see-all-padding-x py-3">
           <h1 class="title">{{ showAllCourses.name }}</h1>
-          <div class="see-all-pages" v-if="numPages > 1">
-            <span class="see-all-pageCount">{{ pageText }}</span>
-            <div class="see-all-buttonWrapper">
-              <button
-                class="see-all-button"
-                :class="{ 'see-all-button--disabled': !hasPrevPage }"
-                :disabled="!hasPrevPage"
-                @click="prevPage()"
-              >
-                <span class="see-all-button-text">Prev</span>
-              </button>
-              <button
-                class="see-all-button"
-                :class="{ 'see-all-button--disabled': !hasNextPage }"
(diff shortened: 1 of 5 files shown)


andxu282 (changes requested):
I think you can delete the `cypress/screenshots`

src/components/Requirements/RequirementSideBar.vue
```diff
@@ -198,8 +189,10 @@ export default defineComponent({
       numOfColleges: 1,
       showAllCourses: { name: '', shownCourses: [], allCourses: [] },
       shouldShowAllCourses: false,
-      showAllPage: 0,
+      // showAllPage: 0,
```
andxu282, line 192: remove comment

src/components/Requirements/RequirementSideBar.vue
```diff
@@ -270,25 +322,47 @@ export default defineComponent({
...
     },
+    /**
+     * @brief Returns tooltip text for courses
+     */
     getCoursesTooltipText() {
       return `<div class="introjs-tooltipTop"><div class="introjs-customTitle">These are your Courses</div><div class="introjs-customProgress">2/4</div>
       </div><div class = "introjs-bodytext">Drag and drop courses into your schedule! Click on them to learn more information like their descriptions.</div>`;
     },
+    /**
+     * @brief Modify this.state.showAllCourses to show all courses based on showAllCourses object
```
andxu282, line 363: I like these documentations

zachary-kent (changes requested):
Can we make this backwards compatible, if possible? That is, if there are no suggestions for a given requirement, then fall back to the previous approach,

PR #874, opened 2023-11-13
Requirement Migrations

nidhi-mylavarapu:
### Summary

- Implemented requirement migrations structure
- Adjusted types to accommodate migrations for majors
- Decorated migration requirements with courses with JSON generator
- Wrote documentation on how to add requirement migrations each year

### Test Plan

- Using the ECE Major, change entry year to post / pre 2021 and notice changes in Core Courses requirements
- Adding new migrations for other majors/minors and seeing changes in requirement bar
- Check that after adding new migrations, especially **Modify** migrations, that the order of requirements stays the same

### Notes 

Documentation in Notion describing how to add requirement migrations: https://www.notion.so/courseplan/Requirement-Migrations-126ffdc4c2274e15a9e40bb8432b791b?pvs=4

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -24,7 +24,7 @@ import csRequirements, { csAdvisors } from './majors/cs';
 import deaRequirements, { deaAdvisors } from './majors/dea';
 import easRequirements, { easAdvisors } from './majors/eas';
 import economicsRequirements, { economicsAdvisors } from './majors/econ';
-import eceRequirements, { eceAdvisors } from './majors/ece';
+import eceRequirements, { eceAdvisors, eceMigrations } from './majors/ece';
 import essRequirements, { essAdvisors } from './majors/ess';
 import englishRequirements, { englishAdvisors } from './majors/engl';
 import envEngineeringRequirements, { envEngineeringAdvisors } from './majors/envE';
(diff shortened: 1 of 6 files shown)


zachary-kent (dismissed):
Awesome work!! Can you add some more inline documentation? You'll need to remove "WIP" from the PR title for the pending check to pass.

PabloRaigoza:
Great work Nidhi! I think this is a great step for adding support to changing majors and minors. Documentation of this code looks great, I couldn't find any immediate problem with the code written. Excellent use of maps and filters. In the future have we considered adding support of this feature for changing college requirements? The ILR school changed their college requirements in 2022, it would be useful to be able to adopt similar method to encounter this case in the future.

elizabeth-tang:
Looks really good Nidhi! I'm curious about what we want to do with the requirements migrations after everyone that the former requirement applies to graduates. They don't necessarily take up a lot of space, but we might want to think about cleaning them up in our annual update, or at least sorting them by year so we can easily keep track of which migrations exist and are applicable.

Destaq:
Tested according to test plan and works great, well done Nidhi. Not doing a detailed review on account of all the feedback you received earlier.

Destaq (approved):

PR #877, opened 2023-12-04
multiple plans!

elizabeth-tang:
Finish implementing the multiple plans feature!

- Format modals according to Figma
- Add TextInputModal.vue that saves a text input. Component of CopyPlanModal.vue, and EditPlanModal.vue
- Write methods that will be used to call firestore functions
- Call firestore functions to create/edit/delete plans
- Confirmation modals when plans are correctly added/renamed/removed
- Fix bug so that Firebase doesn't randomly delete plans anymore
- Fix bug where the correct plan is not displayed

Remaining TODOs:
- Write Cypress e2e frontend tests

Documentation will be here: https://www.notion.so/courseplan/Multiple-Plans-c01b61f8a7a9473781539dc4ec24a5d9


andxu282:
this PR is quite large haha, is this meant to be branched off of something else other than main? like https://github.com/cornell-dti/course-plan/pull/858

andxu282:
Couple of things:
1. I think the renaming probably should disallow for empty strings, and also, I didn't get a screen recording of this, but when I first logged in, my plan was 'undefined' so maybe changing the default to be 'Plan #1' or something?
https://github.com/cornell-dti/course-plan/assets/53061040/3c51eb17-da5b-4e93-8465-e08528afb6e0

when the name is empty and I try to copy it, it looks like I don't have anything selected, even though I have the empty name plan selected, so disallowing the text to be empty solves this too!

Screenshot 2023-12-04 at 2 16 09 PM

2. A few frontend nits but the spacing on the left is a little small
Screenshot 2023-12-04 at 2 16 26 PM
Screenshot 2023-12-04 at 2 16 19 PM

3. I think it'd be best when creating a blank template that we still create 8 empty semesters instead of the user having to go in and making them one by one

Screenshot 2023-12-04 at 2 19 47 PM

Otherwise, this is looking awesome!! Very excited 😄

andxu282 (approved):
let's get this bread 🍞

PR #881, opened 2024-02-10
Schedule Builder Icon

nidhi-mylavarapu:
### Summary 

This pull request is to add the Schedule Builder icon for the upcoming schedule-building feature. Clicking on this icon will open a tab for users to enter, requirements, classes, and other details to help build a schedule. **This is just an addition of an icon - there is no functionality when it is clicked.**

Remaining TODOs:

- Horizontally center icon & text with Navbar!

### Test Plan 

Screenshot 2024-03-04 at 6 26 41 PM

- Need to pass front-end tests after fixing alignment


src/components/NavBar.vue
```diff
@@ -31,6 +31,12 @@
             <span>Profile</span>
           </div>
         </div>
+        <div class="navbar-buttonWrapper desktop">
+          <button class="navbar-iconWrapper schedule-builder-icon full-opacity-on-hover" />
+          <div class="navbar-iconText">
+            <span>Schedule Builder</span>
```
Destaq, line 37: The text `Schedule Builder` here is currently too wide to fit in the base leftmost side panel, which causes it to stretch out past the bounds (and take with it the other text below icons on the sidebar).

Like you mentioned @nidhi-mylavarapu — alignment is wonky.

One way this could be fixed is to give this specific span element the following CSS properties:

1. `text-align: center`
2. `overflow-wrap: anywhere`

However note that the latter will break up the words into letters, and doesn't look too aesthetic. Need to consult designers.

src/components/NavBar.vue
```diff
@@ -155,26 +161,35 @@ $mobile-navbar-height: 4.5rem;
     background-repeat: no-repeat;
     background-size: auto;
     background-position: center;
+    justify-content: center;
```
Destaq, line 164: I wonder if this is still necessary if you try the fix I suggested?

src/components/NavBar.vue
```diff
@@ -155,26 +161,35 @@ $mobile-navbar-height: 4.5rem;
...
   }
 
   &-buttonWrapper {
     cursor: pointer;
     margin-bottom: 1.5rem;
+    justify-content: center;
```
Destaq, line 170: See above.

Destaq (commented):
This isn't ready to merge yet because of the alignment issues, but that can be fixed with a few lines of CSS + maybe some widening of the sidebar.

One potential way of not widening the sidebar but just fixing as-is to pick a different set of words for 'Schedule Builder', like e.g. 'Build Schedule' (even this isn't perfect). But the idea is to use shorter words.

Destaq:
Also (forgot to mention this) need to update the cypress tests for accessibility I believe.

elizabeth-tang:
Hi Nidhi! Thanks for working on this so quickly. When you address the comments, could you also open a new PR into a separate branch for the "schedule builder" feature? That way we can merge the whole feature at once when we finish it!

andxu282:
@elizabeth-tang i know you wrote above to merge this into a feature branch, but could we feature flag this instead? sorry i didn't mention this to you earlier nidhi, you can read about how to do it here! if you need help i can help you tomorrow as well 
https://www.notion.so/Feature-Flags-85d9a6b6cb8540d1babdbfeeae435ef5

Issue #705, opened 2022-07-28
Reduce App Size with API to Send Requirement Data to Clients


noschiff:
All of our requirement computation is done on the front-end. In order to build the requirement fulfillment graphs, our app needs to know which courses will count for which requirements. Currently, we use a script [`req-gen`](https://github.com/cornell-dti/course-plan/blob/master/src/requirements/requirement-json-generator.ts) to take the information from our major, minor, and colleges typescript files (see [cs.ts](https://github.com/cornell-dti/course-plan/blob/master/src/requirements/data/majors/cs.ts) for an example), determine which courses fulfill the requirements in those files, and put all that information into a big JSON file: [decorated-requirements.json](https://github.com/cornell-dti/course-plan/blob/master/src/requirements/decorated-requirements.json). This JSON has a mapping from every requirement we handle to a list of courses that fulfill that requirement. As a result, it's quick to determine if a course fulfills a requirement on the front-end.

However, this JSON is incredibly large—currently 4.83 MB. The information in this JSON gets directly embedded into our compiled javascript script used to run the vue app. The entire JSON is basically verbatim in that file. When a user goes to courseplan for the first time, the browser has to download this file, which ends up being about 6 MB with the code used to run the vue app (see the screenshot at the bottom of this issue).

It's unnecessary to send the entire JSON to every client because the client only needs information about the requirements relevant to a user's majors, minors, and college. Most of the information in the JSON is unused by a given client since the client will only ever check the data for the user's necessary requirements. For example, a user who is majoring in computer science does not need to know which classes will fulfill the requirements for a history major. We should try to avoid sending unnecessary information because it slows the website load and requires the browser to store extra data.

[...]

PR #889, opened 2024-02-15
Switch to an external cloud function to store requirements JSON

Destaq:
### Summary 

In #705 we talked about using a firebase cloud function to send relevant sub-parts of the decorate requirements json file so that initial page speed is quicker. This PR takes steps towards implementing this:

- [x] instead of reading from `decorated-requirements.json` we instead have a store that contains the requirements data that is relevant to a specific user
- [x] data / the store is updated when a user changes majors etc. on their profile
- [x] working functionality when a new user is onboarded as well

Remaining TODOs:

- [ ] create the actual cloud function in firebase (I don't have perms for this, have been running locally)
- [ ] somehow (GitHub Action?) make it so that the JSON file which the cloud function it itself references is updated whenever `npm run req-gen` is done (aka when we have a new major or whatnot then we need to update the base requirements JSON file)
- [ ] update URL in `src/requirements/filter-from-api.ts` to match TODO 1

### Test Plan 

Buckle up this is going to be long.

First up you need to make a local server that you can call, since we don't have a firebase cloud function setup yet. (Once we have it setup then this will be quite easier to test.)

1. Make a new express app somewhere and add the following in `index.js`

```js
const express = require("express");
const fs = require("fs").promises; // Use fs promises API for simpler async handling
const path = require("path");

const app = express();
const port = 3000;

// Enable CORS for all routes
app.use((req, res, next) => {
    res.header("Access-Control-Allow-Origin", "*");
    next();
});

// Function to read and optionally filter the JSON data
async function readAndFilterData(filterParams) {
    const filePath = path.join(__dirname, "decorated-requirements.json");
    const jsonData = await fs.readFile(filePath, "utf8");
    const data = JSON.parse(jsonData);

    // Always include "university": "UNI"
    const filteredData = { university: data.university };

    // Filter based on provided query parameters
    ["grad", "minor", "major", "college"].forEach((param) => {
        if (filterParams[param]) {
            // If the parameter exists in the query string, and is found in data, add it to filteredData
            const paramValues = filterParams[param].split(",");
            filteredData[param] = {};
            paramValues.forEach((value) => {
                if (data[param] && data[param].hasOwnProperty(value.trim())) {
                    filteredData[param][value.trim()] =
                        data[param][value.trim()];
                }
            });
        } else {
            // If the parameter doesn't exist, or not found, include everything under it
            // *unless* the parameter is "skip-this", in which case we don't include it at all.
            if (param !== "skip-this") {
                filteredData[param] = data[param];
            }
        }
    });

    return filteredData;
}

// Express route to handle the GET requests
app.get("/requirements", async (req, res) => {
    try {
        const filteredData = await readAndFilterData(req.query);
        res.json(filteredData);
    } catch (error) {
        res.status(500).json({ error: "Failed to read or filter data" });
    }
});

app.listen(port, () => console.log(`Server running on port ${port}`));

```

2. At the same level as the above file, copy over the `decorated-requirements.json` file.
3. `node index.js`
4. Go to `src/requirements/filter-from-api.ts` and update `BASE_URL` to `http://localhost:3000/requirements`.
5. Now you can test things properly.

Note that if you want to test against a locally run firebase cloud function, you'll have to message me — took a lot of configuring and would be overwhelming to put things here.

### Notes 

*I really don't like this PR for a number of reasons.*

- Because we always only see a subset of the requirements, when you are onboarding / editing your major or minor or college we have to constantly ping the server
- This is supposed to make things go faster by returning a smaller requirements json file (true), but isn't there inherent lag in fetching?
- Since we have to wait until the store + site is loaded before we can call the server (since we need to know what major the user is for example), there is a short flicker (probably even bigger once we are not having a local server) as the requirements sidebar panel is filled with data (i.e. it goes from having 0/0 requirements satisfied for your college to everything else)
- It will be a pain to update the cloud function whenever the base requirements graph is generated
- We are passing around the requirements json file a bit within functions (this is necessary, I minimized it as much as I could but it still happened a few places) which might be inefficient

Note that the tests are failing because currently the virtual machine runner can't access the cloud function server since it's localhost.

### Breaking Changes  

Lots of internal reworking to prevent there from being cyclical dependencies. Means that some of the internal util "APIs" have changed.

### DIFF
--- a/src/containers/Profile.vue
+++ b/src/containers/Profile.vue
@@ -70,7 +70,6 @@ import OnboardingTransfer from '@/components/Modals/Onboarding/OnboardingTransfe
 import { setAppOnboardingData } from '@/global-firestore-data';
 import ProfileConfirmation from '@/components/Modals/ProfileConfirmation.vue';
 import { computeGradYears, SeasonOrdinal } from '@/utilities';
-import { getMajorFullName, getMinorFullName, getGradFullName } from '@/store-utilities';
 import store from '@/store';
 
 const placeholderText = 'Select one';
@@ -114,16 +113,19 @@ export default defineComponent({
       }
       return false;
     },
+    // FIXME: the requirement graph is not generated for the chosen major / minor / grad
+    // etc. and so cannot tell if the user has chosen a valid major / minor / grad.
     isInvalidMajorMinorGradError(): boolean {
-      return (
-        this.onboarding.major
-          .map(getMajorFullName)
-          .some((majorFullName: string) => majorFullName === '') ||
-        this.onboarding.minor
-          .map(getMinorFullName)
-          .some((minorFullName: string) => minorFullName === '') ||
-        (this.onboarding.grad ? getGradFullName(this.onboarding.grad) === '' : false)
-      );
+      return false;
+      // return (
+      //   this.onboarding.major
+      //     .map(getMajorFullName)
+      //     .some((majorFullName: string) => majorFullName === '') ||
+      //   this.onboarding.minor
+      //     .map(getMinorFullName)
+      //     .some((minorFullName: string) => minorFullName === '') ||
+      //   (this.onboarding.grad ? getGradFullName(this.onboarding.grad) === '' : false)
+      // );
     },
     isError(): boolean {
       if (!this.changed || this.isInvalidGraduationSemester || this.isInvalidMajorMinorGradError) {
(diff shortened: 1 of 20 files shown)


src/containers/Profile.vue
```diff
@@ -119,16 +113,19 @@ export default defineComponent({
       }
       return false;
     },
+    // FIXME: the requirement graph is not generated for the chosen major / minor / grad
```
Destaq, line 116: This wasn't super clear. Essentially what I mean here is that if you click to pick a new major (or college or minor or grad), we don't fetch from the cloud function to get the requirements for that major. Doing so would take say 0.5 seconds and so the warning would flash for 0.5 seconds, as the current requirement json has no information about whatever new choice you made and would hence think it is invalid.

PR #891, opened 2024-02-21
chore: added engrc 3027 as valid engrc

Destaq:
### Summary 

We are missing a class that can potentially fulfill an Engineering Communications requirement. Fixed this.

### Test Plan 

Drag in ENGRC 3027 with an Engineering College student and note that it counts.


zachary-kent (commented):
I think we should change the checker to accept any ENGRC course.

PR #892, opened 2024-02-29
[Schedule Generator] Add Schedule Modal

andxu282:
### Summary 
Adds the Schedule Modal to pop up when the "Generate" button is clicked.

Depends on [#880](https://github.com/cornell-dti/course-plan/pull/880)

### Test Plan 

https://github.com/cornell-dti/course-plan/assets/53061040/95b7433b-8a2c-4504-bcf9-9aa427cc1f8a

PR #893, opened 2024-03-02
[Schedule Generator] Courses Component

andxu282:
### Summary 
Implements the Courses Component and a placeholder for the Schedule Component.

Depends on [#892](https://github.com/cornell-dti/course-plan/pull/892)

### Test Plan 

Screenshot 2024-03-02 at 3 00 49 PM

### DIFF
--- a/src/components/ScheduleGenerate/Schedule.vue
+++ b/src/components/ScheduleGenerate/Schedule.vue
@@ -1,0 +1,25 @@
+<template>
+  <div class="generated-schedule-schedule">
+    <div>hello</div>
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue';
+
+export default defineComponent({});
+</script>
+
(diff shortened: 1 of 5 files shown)


elizabeth-tang:
Looks really good!! Just a thought -- what do we want to do when we resize the window? Do we want the same thing as what we have with other modals where they will just close out once it gets smaller than a certain size?

Also, this is a separate issue for when we start implementing the backend, but when they close out of this modal, are we still preserving the state of what requirements were already inputted by the user?

andxu282:
> Looks really good!! Just a thought -- what do we want to do when we resize the window? Do we want the same thing as what we have with other modals where they will just close out once it gets smaller than a certain size?
> 
> Also, this is a separate issue for when we start implementing the backend, but when they close out of this modal, are we still preserving the state of what requirements were already inputted by the user?

d

PR #895, opened 2024-03-04
address edge cases in multiple plans

elizabeth-tang:
### Summary 

This pull request addresses some bugs/design decisions found for multiple plans during our testing session

- disallow deleting last plan
- disallow creating two plans with the same name
- character limit for plans
- frontend bugs (colors, borders, etc.)

Remaining TODOs:

- highlight/placeholder text for renaming modal
- some more frontend stuff 

### Test Plan 

image

- Should show warning "A plan with that name already exists." if you try to add a plan with a name that already exists.
- Should not let you delete your last plan
- Should not let you type more than 15 characters.
- Please lmk if you find other issues too!! 

Documentation: https://www.notion.so/Bug-Bash-81ca898208934e9c8a97bcd494adfc34


src/components/Modals/TextInputModal.vue
```diff
@@ -5,10 +5,13 @@
       <div class="textInput-wrapper">
         <input
           class="textInput-userinput"
-          maxlength="charLimit"
+          maxlength="15"
```
nidhi-mylavarapu, line 8: the 15-character limit works for me! do we also want to require that the plan name cannot be empty?

src/components/Modals/EditPlanModal.vue
```diff
@@ -8,10 +8,12 @@
     @left-button-clicked="closeCurrentModal"
     @right-button-clicked="saveChanges"
     @plan-name="setPlanName"
+    @warn-state="isWarn"
     :isPlanModal="true"
     label="Name"
+    warning="A plan with that name already exists."
```
nidhi-mylavarapu, line 14: works!

src/components/Modals/EditPlanModal.vue
```diff
@@ -8,10 +8,12 @@
...
   >
-    <button class="editPlan-delete" @click="deletePlan">DELETE PLAN</button>
+    <button class="editPlan-delete" @click="deletePlan" :disabled="!canDelete">DELETE PLAN</button>
```
nidhi-mylavarapu, line 16: also works - can't delete last plan

nidhi-mylavarapu (commented):
overall all the changes mentioned in the PR work and it closes after saving changes! i think we should add a check that the plan name is nonempty

andxu282:
By the way, I think the PR snapshot is failing at the moment because of the outdated node version, if you rebase off `main` it should fix it

andxu282:
Maybe just disable the "Save Changes" button here if there is a dup?
Screenshot 2024-03-04 at 6 48 00 PM

PR #901, opened 2024-03-15
SP24 Multiple Plans Release

elizabeth-tang:
### Summary 

Multiple plans feature release + some small non-breaking changes

### Test Plan 

- Multiple Plans add, copy, edit 
- That everything works still with different plans


KaylinChan (approved):
Tested multiple plans functionality and looks good, able to add ENGRC 3027 and 3025 without problems, everything looks good!

andxu282 (approved):
damn the frontend tests passed 🤩

Destaq (commented):
Alright in general looks good to me; ENGRCs now work as expected and multiple plans functionality is there.

But there is that one bug that I mentioned on call right now. Say you have a base schedule that fulfills n/X requirements for your college (and m/Y for your major, etc.). You then go to make a new schedule. It fulfills 0/X, 0/Y as expected. However, as soon as you add in a new course it will jump to n/X, m/Y, etc.

Essentially it reads in the requirements for the base schedule to assign to the sidebar progress bars / requirements fulfillment. If you reload the badge / navigate between plans, the issue disappears. But whenever you add a new course, it will show up again.

I think this should be a lesson for all of us to not even *try* to push to prod on Fridays 😁

andxu282 (dismissed):

Destaq (approved):

Destaq (approved):
Still looks good 👍

KaylinChan (approved):
looks good, awesome work!!! :))

plumshum (approved):
Bugs from bug bash are gone. I like that we changed the copy feature to say "[plan name] copy" now. New major works with the multiple plans. However, when I make a copy of the my first plan, then delete a requirement on the copied plan, it also deletes it from my first plan. if i refresh, the progress bar says 6/6, but the bar itself, and the requirement checker aren't. I have to delete the requirement I orginally deleted from my copied plan, add it back in, then everything is back to normal.
image


Maybe for the pop-up explaining the multiple plans, we can specify that we can't change the majors between the plans, because I think users will intuitively expect to be able to switch majors among plans.

PR #904, opened 2024-03-18
[Schedule Generator] Schedules Component

andxu282:
### Summary 
Implements the frontend for the Schedules component of the Schedules modal.

Remaining TODOs:

- [x] Add gray lines
- [x] Make sure that the days don't move if they don't have classes
- [x] Compute the total number of pixels
- [x] Compute the end time as well to make the length of the class be reflected
- [x] Compute min and max hours
- [x] Compute total number of minutes

Depends on [#893](https://github.com/cornell-dti/course-plan/pull/893)

### Test Plan 

Screenshot 2024-03-18 at 9 38 08 PM

### DIFF
--- a/src/components/ScheduleGenerate/Schedule.vue
+++ b/src/components/ScheduleGenerate/Schedule.vue
@@ -2,15 +2,16 @@
   <div class="schedule-main">
     <div class="schedule-body">
       <div class="schedule-hours">
-        <div class="schedule-hour" v-for="hour in hoursRange">{{ hour }}</div>
+        <div class="schedule-hour" :key="hour" v-for="hour in hoursRange">{{ hour }}</div>
       </div>
       <div class="schedule-week">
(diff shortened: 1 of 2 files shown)


KaylinChan:
@andxu282 can you make the vertical lines lighter (#E5E5E5 maybe?) and show what a 2 hour class would look like? Also on my screen, the bottom isn't aligned but I'm not sure if it's a zooming in issue.

image

nidhi-mylavarapu (approved):
a work of art, andrew. i'm assumine we're adding all of the bottom buttons with page numbers / the download button later - other than that, looks perfect :)

elizabeth-tang:
I think Joanna mentioned this during one of our standups but could we also have vertical lines in the background of the schedule? Something like the gcal interface? I think it may help users more easily visualize when the classes actually are
image

PR #907, opened 2024-03-20
style: make course search modal scrollable

Destaq:
### Summary 

Long overdue: make it so that you can scroll when searching for courses when you are searching for them to add to a semester.

Algorithm: sort first by matching course codes, then by title.

### Test Plan 

Use eyes.

Screenshot 2024-03-20 at 6 33 06 PM

### Notes 

Right now I just came up with my own styles — scroll spans `max-height: 50vh;`. Works on both mobile and desktop but maybe there is something else that is better? Though right now Joanna gave this the ok since it is pretty much the same height as if you had 10 courses (the default).

Also there is "dead code" (I just commented stuff out). Sarah and Sophia would undoubtedly be disappointed given what we went over in Dev Sesh 2. However I reckon it's best to leave this up there just in case any bugs come up (I don't think we are going to be bug bashing this).

### DIFF
--- a/src/components/Modals/NewCourse/CourseSelector.vue
+++ b/src/components/Modals/NewCourse/CourseSelector.vue
@@ -51,8 +51,9 @@ const getMatchingCourses = (
   title.sort((first, second) => first.titleLong.localeCompare(second.titleLong));
 
   /* prioritize code matches over title matches */
+  return code.concat(title);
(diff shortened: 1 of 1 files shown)


KaylinChan (approved):
tested on mobile and web and looks great! thanks for the long overdue fix 😍

PR #913, opened 2024-03-25
[Schedule Generator] Schedule Fixes

andxu282:
### Summary 
Fixes a couple of frontend issues for the schedule component
- [x] the Courses and Schedules sections now use `display: flex` so they're always aligned
- [x] course heights are always fixed and the schedule component height is calculated based on the number of hours, so overflow will never occur.

https://github.com/cornell-dti/course-plan/assets/53061040/a3466cbd-4016-4db9-9c9c-6a1987a16b03

https://github.com/cornell-dti/course-plan/assets/53061040/88f3f2c3-0c62-4afe-8298-535be7bd7585

### DIFF
--- a/src/components/ScheduleGenerate/Schedule.vue
+++ b/src/components/ScheduleGenerate/Schedule.vue
@@ -175,7 +175,7 @@ export default defineComponent({
       margin-bottom: 1.8rem;
       justify-self: center;
     }
-    border-left: 1px solid #e5e5e5;
+    border-left: 1px solid $veryLightGray;
   }
   &-hours {
     display: flex;
--- a/src/components/ScheduleGenerate/Schedule.vue
+++ b/src/components/ScheduleGenerate/Schedule.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="schedule-main">
-    <div class="schedule-body">
-      <div class="schedule-hours">
+    <div class="schedule-body" :style="getTotalHeight()">
+      <div class="schedule-hours" :style="getTotalHeight()">
         <div class="schedule-hour" :key="hour" v-for="hour in hoursRange">{{ hour }}</div>
       </div>
       <div class="schedule-week">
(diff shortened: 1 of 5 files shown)


plumshum (commented):
The fixes overall look good to me. 
Your comment about what would happen if there are too many courses and it overflows, I think we should make the `Your Courses` component scrollable then. The height and width of the component shouldn't change.  
Seems like there are a few css colors that we can convert into scss variables aswell. (Line 178 in `Schedule.vue`.)

src/components/ScheduleGenerate/Schedule.vue
```diff
@@ -167,17 +175,16 @@ export default defineComponent({
       margin-bottom: 1.8rem;
       justify-self: center;
     }
-    border-left: 1px solid $inactiveGray;
+    border-left: 1px solid #e5e5e5;
```
elizabeth-gs, line 178: can we extract this into a color variable?

src/components/ScheduleGenerate/ScheduleGenerateModal.vue
```diff
@@ -27,25 +27,17 @@
...
           </div>
           <div class="schedule-generate-section-schedule">
             <div class="schedule-generate-subHeader schedule-generate-subHeader--indent">
-              <span class="schedule-generate-subHeader--font">Fall 2024</span>
-            </div>
-            <div class="schedule-generate-inputs">
-              <div class="schedule-generate-inputWrapper">
-                <schedule :classesSchedule="classesSchedule" />
-              </div>
+              Fall 2024
```
elizabeth-gs, line 38: This should probably be a prop (selectedSemester), like you have for `ScheduleGenerateSideBar.vue` that you pass in from `Dashboard.vue`.

elizabeth-gs (changes requested):
Looks really good! Just a few small nits:

- can we extend the lines and the times to fill the space? I've had classes late at night before (i.e. 7:30-8:45). If that exceeds the bounds of the modal, I think we should also make the component scrollable.

PR #915, opened 2024-04-15
Documentation for sched gen courses frontend

elizabeth-tang:
**Summary**
Documentation for implemenation of the frontend for the courses under each requirement for schedule generator
- `RequirementCourses` now has a new field in data, `courses`, that will contain all the courses for each requirement. T
- The courses are rendered as a list of `Course` components, each identified by an id
- isSchedGenCourse was added as an optional prop to `Course`,  and is used to specify the features in each course card (i.e. no course caution symbol, etc). 
image

**Remaining TODOs:**
- weird scroll issue on sidebar: if I set overflow-y: scroll, overflow-x no longer works so we can't see the CourseMenu
- once requirement ids are added, limiting the dropdown available courses to only the ones that fulfill the requirement

**Test Plan**
- Test clicking '+ Course' button and adding a course to a requirement
- Test deleting a course from a requirement

### DIFF
--- a/src/components/Course/Course.vue
+++ b/src/components/Course/Course.vue
@@ -42,7 +42,11 @@
         <div v-if="!compact" class="course-info">
           <span class="course-credits">{{ creditString }}</span>
           <span v-if="semesterString" class="course-semesters">{{ semesterString }}</span>
-          <course-caution v-if="!isReqCourse" :course="courseObj" :isCompactView="false" />
+          <course-caution
+            v-if="!isReqCourse && !isSchedGenCourse"
+            :course="courseObj"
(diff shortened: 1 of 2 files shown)


andxu282:
Thanks for working on this! If I understand correctly, this PR is adding support for filtering the available courses based on the requirement (once I get the PR out), right?

nidhi-mylavarapu (commented):
This is great! Super helpful for me to connect all the info to the algorithm - it might be easier for me to work off of this as I'm connecting to the frontend? Right now I'm trying to connect with the algo with a hard coded request which I'd later replace with a user inputted request - will probably use some of these changes instead

PR #917, opened 2024-04-15
Add search to requirements sidebar

Destaq:
### Summary 

Added search functionality to requirements sidebar. UI is based on that of `CourseSelector.vue`.

### Test Plan 

Look at the preview URL and play around with it.

### DIFF
--- a/src/components/Requirements/RequirementSideBar.vue
+++ b/src/components/Requirements/RequirementSideBar.vue
@@ -474,6 +474,7 @@ export default defineComponent({
       };
     },
     applyFilter() {
+      this.showAllPage = 0;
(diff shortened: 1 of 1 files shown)


plumshum (commented):
Ooo I really like this search bar implementation. I didn't encounter any bugs using the search bar. I have a few suggestions about the front-end and design.
1. Can we make the search box the same width as the course cards so it looks more aligned? The right end seems a bit off. 
image
2. If there are no courses that match what the user typed in the search bar, I think some sort of text should pop up instead, something like `No Courses Found`. It's providing insights for the user as to what happened.

PR #919, opened 2024-04-18
Support schedule downloading + pagination in schedgen modal

Destaq:
### Summary 

Finished out the frontend components for scheduler generator (currently UI-only). We now have a 'regenerate schedules' button as well as going through prev / next with a page indicator. Additionally, there is also a 'download schedule' button which is synced with the modal props & values to allow for seamless PDF export.

### Test Plan 

Take a look at the preview URL.


src/tools/export-plan/pdf-schedule-generator.ts
```diff
@@ -0,0 +1,506 @@
...
+  // rendering PDF header for page 2
+  programY = await renderSecondHeader(doc);
+
+  // Rendering tables now
+  await renderBaseTable(doc, classes, year, season, programY, tableX, 'right');
+
+  // rendering PDF footer
+  renderFooter(doc);
+
+  const pdfName = `${store.state.userName.firstName}_CoursePlan`;
```
andxu282, line 99: could you call this something else like "Generated_Schedule" to differentiate it from the other PDF download?

andxu282 (approved):
LOOKING FIRE! Just the one nit and it should be good to go!

src/tools/export-plan/pdf-schedule-generator.ts
Destaq: Shoot sorry I missed this comment somehow, fixed on another branch.

PR #925, opened 2024-04-29
[Schedule Generator] Frontend Fixes

andxu282:
### Summary 
Fixes a couple of frontend issues
- Sets up the frontend logic for the disabling of the generate button
- When generating, the "Generate" button grays out and the texts turns into "Generating..."
- The minimum window for the schedule is now 8AM - 4PM
- Fixed a bug where the maximum hour displayed would be 1 less than the actual range
- Checks the credit limit and disables "Generate Schedule" until the credit limit is between 12 and 22
- set the courses and schedule on the schedule modal to overflow: scroll

Screenshot 2024-04-29 at 2 55 45 PM

Screenshot 2024-04-29 at 2 56 40 PM

https://github.com/cornell-dti/course-plan/assets/53061040/2b211c0b-b406-4d10-9868-3a8ee91db79d

https://github.com/cornell-dti/course-plan/assets/53061040/420c6fc0-2ae3-4cbb-aa68-d0a62b441981


src/components/ScheduleGenerate/ScheduleGenerateModal.vue
```diff
@@ -575,6 +577,7 @@ input {
   display: flex;
   width: 246px;
   height: 35px;
+  font-family: 'ProximaNova';
```
nidhi-mylavarapu, line 580: some of the font-family is 'ProximaNova' and others are 'Proxima Nova' which is what's causing the varying fonts in the modal - i'm pretty sure the one that works for everyone is 'Proxima Nova' * actually i don't know if the font-family is needed here? i think we might have just added it to debug and figure out what's wrong with the font haha so we can prob get rid of it

src/components/ScheduleGenerate/ScheduleGenerateSideBar.vue
```diff
@@ -90,13 +91,15 @@ export default defineComponent({
     requirements: ReqCourses[];
     isConfirmationOpen: boolean;
     confirmationText: string;
-    creditLimit: number;
+    creditLimit?: number;
```
nidhi-mylavarapu, line 94: is there a reason we make creditLimit an optional prop?

nidhi-mylavarapu (commented):
everything looks good! we can just cleanup and get rid of the extra font families but everything look's great

src/components/ScheduleGenerate/ScheduleGenerateSideBar.vue
andxu282: it's undefined initially when the user hasn't put anything in yet

PR #937, opened 2024-05-06
Documentation Update for Schedule Generator

Destaq:
### Summary 

Added a lot of inline / method documentation to the schedule generator TypeScript backend algorithm. Also increased README.md by about 30% with additional details such as the exact implementation of the algorithm. Finally, added one line that randomizes courses' timeslots when being considered by the algorithm to ensure a higher-quality output.

### Test Plan 

Only documentation updates (no test plan required) except for one change: randomizing courses' timeslots in the algorithm to ensure that there is more variety in generated schedules. (Realized as I was writing up some of the documentation.)


Destaq:
I would appreciate help in figuring out why the prettier checker is failing.

elizabeth-tang:
> I would appreciate help in figuring out why the prettier checker is failing.

Could you rebase off of main first? Then we can see your changes more clearly and diagnose the issue!

Destaq:
See #938 .

PR #882, opened 2024-02-11
[Saved Courses] Add Icon to Sidebar

andxu282:
### Summary 
Adds the saved courses icon to the sidebar.

Remaining TODOs:

- [x] make the icon fill the entire square
- [x] on hover turn blue

Depends on #{number of PR}

### Test Plan 

Screenshot 2024-02-13 at 5 24 42 PM

https://github.com/cornell-dti/course-plan/assets/53061040/f66938be-b965-4e3d-a744-711c1661a0d4

PR #940, opened 2024-09-17
collections page empty front-end look

plumshum:
### Summary 

Created the basic frontend look for the sidebar Collection page.
- when implemented with backend remove the hardcoded collections 

This pull request is part of the Saved Course Feature
Here is the lastest demo:

https://github.com/user-attachments/assets/120047e0-e959-45c5-afe4-e1c1af76f638

- [x] empty front-end look for the sidebar
image
- [x] front-end look for multiple collections (Note: I hardcoded courses into the All collection, so there is no logic in place to also put it in the other collections. It's also why there's alot of duplicated code between the All collection and the dynamic collection. )
image
image
- [x] EditCollectionModal
image

**Breaking Changes**
- I made `Course.vue` have prop `:isSemesterCourseCard` so that I can change the course card look for the collection. 
- Technically, I'm using FirestoreSemesterCourse, but I feel like I should be using CornellRosterCourse 
- I decided to make direct copy of `TextInputModal` and `EditPlanModal` for collections so that I don't make any breaking changes with multiple plans, but I think we can make a tempalte `TextInputModal` that can work for any feature modals that require its functionality

Remaining TODOs:
- [ ] make Course Draggable 

Depends on #882 (It has the Saved Course Icon on the side to click into this page)

### Test Plan 

Make sure the frontend looks accurate with design


src/components/Modals/CollectionsSideBar.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+  <div class="sidebar">
+    <div class="sidebar-header">
+      <div class="separator-header"></div>
+      <h1 class="top header-title">Collections</h1>
```
Destaq, line 5: There's a bit of misalignment with the Figma designs here, this text's bottom padding is a bit more than it should be (nitpicking).

src/components/Requirements/RequirementSideBar.vue
```diff
@@ -244,15 +252,17 @@ export default defineComponent({
...
   props: {
     startTour: { type: Boolean, required: true },
     isDisplayingMobile: { type: Boolean, required: true },
     isMobile: { type: Boolean, required: true },
     isMinimized: { type: Boolean, required: true },
     startNewFeatureTour: { type: Boolean, required: true },
+    isDisplayingCollection: { type: Boolean, required: true },
   },
-  emits: ['showTourEndWindow', 'toggleMinimized'],
+  emits: ['showTourEndWindow', 'toggleMinimized'], // probably will remove emits with just using components
```
Destaq, line 265: Could you clarify what you meant here?

Destaq (commented):
Nice job implementing the skeleton for this, looks in-line with the rest of the site! I picked up on a couple small Figma issues that would be good to have addressed, but as soon as you do those should be good to merge. Well done with SCSS and finding a relevant icon~

src/components/Modals/CollectionsSideBar.vue
```diff
@@ -0,0 +1,97 @@
...
+    </div>
+    <div>
+      <button class="dropdown-item-button">
+        <img src="@/assets/images/dropdown.svg" alt="open dropdown" class="hover-image" />
```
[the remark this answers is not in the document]
plumshum: Yes that makes sense once it becomes reactive. Another thing to consider is the change in color as well (it will become blue).

src/components/Requirements/RequirementSideBar.vue
plumshum: Sorry that's an old comment. Originally, I made an 'emit' to open the sidebar, then completely scratching that idea.

src/components/Course/Course.vue
```diff
@@ -146,6 +165,14 @@ export default defineComponent({
       this.$emit('delete-course', this.courseObj.code, this.courseObj.uniqueID);
       this.closeMenuIfOpen();
     },
+    deleteCourseFromCollection() {
+      this.deletingCourse = true;
+      this.$emit('delete-course-from-collection', this.courseObj.code);
+      // wait to allow deletion
```
plumshum, line 171: Another way to do this?

PR #942, opened 2024-09-19
chore: normalize search

Destaq:
We now ignore whitespace (in addition to case) when searching for a new course in the course selector. You can confirm this by, on this branch, typing in e.g. `cs1110` and noticing that `CS 1110: ...` gets matched.

### DIFF
--- a/src/components/Modals/NewCourse/CourseSelector.vue
+++ b/src/components/Modals/NewCourse/CourseSelector.vue
@@ -44,11 +44,13 @@ const getMatchingCourses = (
   } else {
     filteredCourses = filter != null ? fullCoursesArray.filter(filter) : fullCoursesArray;
   }
+  const normalizedSearchText = searchText.toUpperCase().replace(/\s+/g, '');
   for (const course of filteredCourses) {
-    const courseCode = `${course.subject} ${course.catalogNbr}`;
-    if (courseCode.toUpperCase().includes(searchText)) {
+    const courseCode = `${course.subject}${course.catalogNbr}`.toUpperCase().replace(/\s+/g, '');
+    const courseTitle = course.titleLong.toUpperCase().replace(/\s+/g, '');
+    if (courseCode.includes(normalizedSearchText)) {
       code.push(course);
-    } else if (course.titleLong.toUpperCase().includes(searchText)) {
+    } else if (courseTitle.includes(normalizedSearchText)) {
       title.push(course);
     }
   }


src/components/Modals/NewCourse/CourseSelector.vue
```diff
@@ -44,11 +44,13 @@ const getMatchingCourses = (
...
+  const normalizedSearchText = searchText.toUpperCase().replace(/\s+/g, '');
   for (const course of filteredCourses) {
-    const courseCode = `${course.subject} ${course.catalogNbr}`;
-    if (courseCode.toUpperCase().includes(searchText)) {
+    const courseCode = `${course.subject}${course.catalogNbr}`.toUpperCase().replace(/\s+/g, '');
+    const courseTitle = course.titleLong.toUpperCase().replace(/\s+/g, '');
+    if (courseCode.includes(normalizedSearchText)) {
       code.push(course);
-    } else if (course.titleLong.toUpperCase().includes(searchText)) {
+    } else if (courseTitle.includes(normalizedSearchText)) {
```
plumshum, line 53: Thank you fixing this too, so now if you write "introtocomputing", results for "intro to computing" will show up

plumshum (approved):
I tested out the search text with different number of whitespaces, and it works for both the course code and the course title.

plumshum:
Can we also change the code atp so that if the user types both the course code and title, courseplan also shows the valid courses? right now if you write cs 1110: intro..., it will show nothing, but "cs 1110" and "introduction to computing" will show `cs1110`

nidhi-mylavarapu:
> Can we also change the code atp so that if the user types both the course code and title, courseplan also shows the valid courses? right now if you write cs 1110: intro..., it will show nothing, but separately writing "cs 1110" and "introduction to computing" will show `cs1110`

Def something we can look at in the future - i think for now doing it with course number as one possible search and course title as one possible search will work... especially because if they start typing cs 1110, they'll see the course they need before they finish typing anyway. but definetely something we can look at in the future

src/components/Modals/NewCourse/CourseSelector.vue
nidhi-mylavarapu: Thank you for the fix! tested with different course codes with and w/o spaces - works on my end as well.

nidhi-mylavarapu (approved):
thank you for this fix simon!! approved :)

Destaq:
> Can we also change the code atp so that if the user types both the course code and title, courseplan also shows the valid courses? right now if you write cs 1110: intro..., it will show nothing, but separately writing "cs 1110" and "introduction to computing" will show `cs1110`

Implemented!

PR #943, opened 2024-09-28
ci/cd: migrate to non-deprecated upload artifact version

Destaq:
Let's fix our CI/CD pipeline. Semantic commit says it all.


plumshum (approved):
This is the best change ever. Finally, all those cypress tests will work.

PR #944, opened 2024-10-02
collection backend

plumshum:
### Summary 

This pull request is the first step towards adding collection into Firebase for implementing the Save Courses Feature.
Implement functionality to save selected courses into collection(s) in Firebase.

updated documentation: https://www.notion.so/Saved-Courses-8154144312034ccc914a6a37ef5ff0a6
### Breaking Changes with Backend ###
- Every user has a `Plan 1` that's automatically added if the user has no plan, but the code doesn't seem to do this with the default Collection `All`, so that was breaking my entire Courseplan account. I was hoping a migration can fix it I also added a quick script for that.

- [x] add collection to firebase
- [x] when add a course to a collection, backend should be able to store that course into the collection
- [x] course is removed from semester view and removed from requirements  
- [x] collection can be named by user before creation 
- [x] front-end matches with figma design (fixed padding, highlighting and auto-boldening of the box)

Here's the latest demo:

https://github.com/user-attachments/assets/cdbf30d9-1a6d-4885-a467-55cabe0a93e4

Here's a Screenshot of the collections with a scrollbar and editability:
image
image

Here is a demo of adding one course into one newly created collection. (this is with the old front-end look, but the backend is the same)
image

Here is a demo of adding one course into multiple new created collections (this is with the old front-end look, but the backend is the same)
image

Remaining TODOs:
- [x] debugging saved courses (hopefully migration script can fix it!!)

Depends on #{number of PR}
#921 

### Notes 

Concluded to store savedCourses in FireStoreSemesterData. 

### Blockers 

### Breaking Changes  

- `store` used to have **collections** now it's renamed to **savedCourses** for clarity.

### DIFF
--- a/src/components/Modals/SaveCourseModal.vue
+++ b/src/components/Modals/SaveCourseModal.vue
@@ -25,11 +25,16 @@
 
     <div class="saveCourseModal-body">
       <div class="saveCourseModal-body-content">
-        <p v-for="(collection, index) in collections" :key="index">{{ collection }}</p>
-        <!--Must find all possible collections
-            Checkbox Style
-            Need a collections variable in firestore
-        -->
+        <div v-for="collection in collections" :key="collection">
+          <input
+            v-if:="!isDefaultCollection"
+            type="checkbox"
+            id="collection"
+            :value="collection"
+            v-model="checkedCollections"
+          />
+          <label for="collection">{{ collection }}</label>
+        </div>
       </div>
     </div>
   </teleport-modal>
@@ -44,19 +49,26 @@ export default defineComponent({
   components: { TeleportModal },
   props: {
     courseCode: { type: String, required: true },
-    isdefaultCollection: { type: Boolean, default: true },
   },
   data() {
     return {
-      collections: store.state.collections.map(collection => collection.name),
+      checkedCollections: [] as string[], // New data property to manage checked state
     };
   },
   computed: {
-    placeholderName() {
+    isDefaultCollection() {
+      const collections = store.state.collections.map(collection => collection.name);
+      return collections.length === 0;
+    },
+    collections() {
+      const collections = store.state.collections.map(collection => collection.name);
+      return collections.length === 0 ? ['No collections added yet'] : collections;
+    },
+    placeholder_name() {
       const oldcollections = store.state.collections.map(collection => collection.name);
       let newCollectionNum = 1;
       // eslint-disable-next-line no-loop-func
-      while (oldcollections.find(p => p === `Collection ${newCollectionNum}`)) {
+      while (oldcollections.find(p => p === `New Collection ${newCollectionNum}`)) {
         newCollectionNum += 1;
       }
       return `New Collection ${newCollectionNum}`;
@@ -64,8 +76,7 @@ export default defineComponent({
   },
   emits: {
     'close-save-course-modal': () => true,
-    'save-course': (courseCode: string, collections: string[]) =>
-      typeof courseCode === 'string' && typeof collections === 'object',
+    'save-course': (collections: string[]) => typeof collections === 'object',
     'add-collection': (name: string) => typeof name === 'string',
   },
   methods: {
@@ -73,11 +84,11 @@ export default defineComponent({
       this.$emit('close-save-course-modal');
     },
     saveCourse() {
-      this.$emit('save-course', this.courseCode, this.collections);
+      this.$emit('save-course', this.checkedCollections);
       this.closeCurrentModal();
     },
     addCollection() {
-      this.$emit('add-collection', this.placeholderName);
+      this.$emit('add-collection', this.placeholder_name);
     },
   },
 });
--- a/src/components/Modals/SaveCourseModal.vue
+++ b/src/components/Modals/SaveCourseModal.vue
@@ -14,7 +14,6 @@
     </template>
 
     <div class="saveCourseModal-header">
-      <!-- create a rectangular border-->
       <div class="saveCourseModal-header-text">
         <span> Collections </span>
         <button class="saveCourseModal-header-text-addButton" @click="addCollection">
@@ -23,20 +22,28 @@
       </div>
     </div>
 
-    <div class="saveCourseModal-body">
-      <div class="saveCourseModal-body-content">
-        <div v-for="collection in collections" :key="collection">
+    <div :class="['saveCourseModal-body']">
+      <div
+        :class="[
+          'saveCourseModal-body-content',
+          { 'default-collection': isDefaultCollection, scrollable: isNumCollectionGreaterThanFour },
+        ]"
+        v-for="(collection, index) in collections"
+        :key="collection"
+      >
+        <div class="saveCourseModal-body-content-collection">
           <input
-            v-if:="!isDefaultCollection"
+            v-if="!isDefaultCollection"
             type="checkbox"
-            id="collection"
+            :id="'collection-' + index"
             :value="collection"
             v-model="checkedCollections"
           />
-          <label for="collection">{{ collection }}</label>
+          <label :for="'collection-' + index">{{ collection }}</label>
         </div>
       </div>
     </div>
+    <div v-if="isNumCollectionGreaterThanFour" class="saveCourseModal-divider-line"></div>
   </teleport-modal>
 </template>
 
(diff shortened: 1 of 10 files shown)


src/components/Modals/SaveCourseModal.vue
```diff
@@ -17,19 +17,24 @@
...
     </div>
 
     <div class="saveCourseModal-body">
       <div class="saveCourseModal-body-content">
-        <p>{{ collection }}</p>
-        <!--Must find all possible collections
-            Checkbox Style
-            Need a collections variable in firestore
-        -->
+        <div v-for="collection in collections" :key="collection">
```
Destaq, line 28: We would guarantee that collection names are unique then — not tracking internally with IDs and whatnot? (Is a reasonable assumption to make.)

src/components/Modals/SaveCourseModal.vue
```diff
@@ -17,19 +17,24 @@
...
+          <input
+            v-if:="!isDefaultCollection"
+            type="checkbox"
+            id="collection"
+            :value="collection"
+            v-model="checkedCollections"
+          />
+          <label for="collection">{{ collection }}</label>
```
Destaq, line 36: Heads up, I noticed when testing that this label will always only toggle the very first collection when clicked. So e.g. the label for the fourth collection, when clicked, would toggle the checkmark for first collection.

I suspect that this is due to the non-uniqueness for `id` and `for` — could perhaps combine this with the indices or use `"collections"` literal value?

Additionally, would it be possible to have a little margin between the checkbox and the label? I appreciate that this is a draft PR though 👍 .

src/components/Modals/SaveCourseModal.vue
```diff
@@ -38,30 +43,52 @@
...
 export default defineComponent({
   components: { TeleportModal },
   props: {
     courseCode: { type: String, required: true },
-    collection: { type: String, default: 'No collections added yet' },
-    isdefaultCollection: { type: Boolean, default: true },
+  },
+  data() {
+    return {
+      checkedCollections: [] as string[], // New data property to manage checked state
```
Destaq, line 55: Good work with the type safety here~

src/user-data.d.ts
```diff
@@ -36,6 +36,7 @@ type FirestoreSemester = {
 };
 
 type FirestoreSemestersData = {
+  // readonly collections: readonly Collection[]; Note: Not sure where to store collections
```
Destaq, line 39: Hmm yes, this would be tricky. Happy to discuss with you and Nidhi sometime at dev sesh or work sesh!

src/store.ts
```diff
@@ -326,6 +331,7 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
       const plan = getFirstPlan(data);
       store.commit('setPlans', data.plans);
       store.commit('setCurrentPlan', plan);
+      // store.commit('setCollections', data.collections); Note: toggle this on and off to save collections progress after refresh
```
Destaq, line 334: Thank you for haring this here and in the main PR information section. Tiny quality of life improvement could be to add another `//` before the `Note: ` so you don't have to add that when toggling, but really just nitpicking.

src/gtag.ts
```diff
@@ -35,7 +37,9 @@ type EventType =
   | 'bottom-bar-view-course-information-on-roster' // User clicks View Course Information on Roster link on Bottom Bar
   | 'course-edit-color' // User edits the course color
   | 'subject-edit-color' // User edits the subject color
+  | 'delete-collection' // User deletes a collection
   | 'delete-course' // User deletes a course
+  | 'delete-course-collection' // User deletes a course from a collection
```
Destaq, line 42: Nice stuff, thank you for commenting all of these so semantically!

src/components/Modals/SaveCourseModal.vue
```diff
@@ -38,30 +43,52 @@
...
+    };
+  },
+  computed: {
+    isDefaultCollection() {
+      const collections = store.state.collections.map(collection => collection.name);
+      return collections.length === 0;
+    },
+    collections() {
+      const collections = store.state.collections.map(collection => collection.name);
+      return collections.length === 0 ? ['No collections added yet'] : collections;
```
Destaq, line 65: Clean way of doing things, nice approach! Took a minute to understand so perhaps a one-line piece of documentation could be useful, but definitely the most straightforward approach — doesn't really on any extra attributes or anything.

src/components/Modals/SaveCourseModal.vue
```diff
@@ -38,30 +43,52 @@
...
+    },
+    placeholder_name() {
+      const oldcollections = store.state.collections.map(collection => collection.name);
+      let newCollectionNum = 1;
```
Destaq, line 69: Could we store this in the state somehow? Iterating through `oldcollections` like this is probably discouraged by `eslint` for a reason. I know it goes counter to my advice from earlier, but in this case I feel like it would be reasonable to store the latest collection number.

Though that said, if this is just a placeholder route and a user would be forced to enter a name or something, then can leave as be (again, I appreciate that this is still a draft-level PR).

src/components/Semester/Semester.vue
```diff
@@ -408,6 +414,35 @@ export default defineComponent({
...
+      const { collections } = store.state;
+      const toEdit = collections.find(collection => collection.name === oldname);
+      const updater = (collection: Collection): Collection => ({
+        name,
+        courses: collection.courses,
+      });
+      if (toEdit !== undefined) {
+        editCollection(oldname, updater);
+      }
+      this.confirmationText = `${oldname} has been renamed to ${name}!`;
```
Destaq, line 440: Maybe we could force capitalization of the first letter here for UI purposes.

src/global-firestore-data/user-semesters.ts
```diff
@@ -138,6 +184,48 @@ export const deletePlan = async (name: string, gtag?: VueGtag): Promise<void> =>
...
 
+/** Add one course to multiple collections.
+ * This course is removed from the requirement choices.
+ */
+export const addCourseToCollections = (
+  plan: Plan,
+  year: number,
+  season: FirestoreSemesterSeason,
+  newCourse: FirestoreSemesterCourse,
+  collectionIDs: string[], // Array of collection IDs
```
Destaq, line 195: Rather redundant to have this comment~

src/global-firestore-data/user-semesters.ts
```diff
@@ -138,6 +184,48 @@ export const deletePlan = async (name: string, gtag?: VueGtag): Promise<void> =>
...
+  editCollections(oldCollections =>
+    oldCollections.map(collection => {
+      if (collectionIDs.includes(collection.name)) {
+        return { ...collection, courses: [...collection.courses, newCourse] };
+      }
+      return collection;
+    })
+  );
+
+  deleteCourseFromSemester(plan, year, season, newCourse.uniqueID);
```
Destaq, line 208: I'd be curious about why we are necessarily removing something from a collection when we add from the dashboard main area, feels a bit counterintuitive. First time I tried it was annoyed that I lost the course and had to go remember what it was and find it.

Not an implementation question, more system design, would be interested to hear why this approach was taken.

Destaq (commented):
Nice work Hannah, great work on the draft PR. Added in some general comments and suggestions. I know that this is still a draft PR, so they are not definitive but hopefully helpful a bit at least — if you have any questions let me know.

My main issue / concern would be with clicking the labels and having those being incorrectly synced to the corresponding checkboxes.

src/components/Modals/SaveCourseModal.vue
nidhi-mylavarapu: I was thinking the same...I think this is a reasonable thing to assume for now.

src/components/Modals/SaveCourseModal.vue
nidhi-mylavarapu: I agree that storing this in state somehow would be better... also (for a very weird niche case that would be kind of annoying) if a renamed collection was named Collection 4 even if the collection was the second collection added, then the newCollectionNum wouldn't be updated correctly

src/components/Semester/Semester.vue
nidhi-mylavarapu: For 'oldname'? Maybe we can ask design, but I'd think that we would have that be the exact same as the original name

src/store.ts
```diff
@@ -326,17 +332,21 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
       const plan = getFirstPlan(data);
       store.commit('setPlans', data.plans);
       store.commit('setCurrentPlan', plan);
+      store.commit('setSavedCourses', data.savedCourses); // Note: toggle this on and off to save collections progress after refresh
```
nidhi-mylavarapu, line 335: Thank you for the clarifying comment here!!

src/store.ts
```diff
@@ -326,17 +332,21 @@ export const initializeFirestoreListeners = (onLoad: () => void): (() => void) =
...
       const { orderByNewest } = data;
       store.commit('setSemesters', plan.semesters);
       updateDoc(doc(fb.semestersCollection, simplifiedUser.email), {
         plans: data.plans,
+        // savedCourses: data.savedCourses,
```
nidhi-mylavarapu, line 340: I saw the toggling comment from above... is there a more intuitive way (rather than toggling within the code) to handle the refresh progress? works for now, but something to consider.

nidhi-mylavarapu (commented):
Thank you for all of your hard work on this Hannah! looking forward to seeing how this aligns with the frontend :)

PR #945, opened 2024-10-04
Link Email Script to Firebase & Add Template

Destaq:
### Summary 

This pull request more or less finishes off the remaining styling and DB tasks for our emailing script. In particular, it does the following:

- [x] connects to Firebase to extract all users with their info
- [x] has a styled, mobile-friendly CSS + HTML template for sending out email reminders

Infrastructure-wise, progress has also been made in migrating DNS to `courseplan.io`, and updating the dummy `to` recipient to be `courseplan.io@gmail.com`.

Remaining TODOs:

- [ ] forward email replies to a slack channel (also involves infra setup)
- [ ] should update it so that enrollment timeslot is a function of when you entered the school, not when you graduate? (for those people who graduate 2+ semesters early) (wouldn't handle deferrals though)
- [ ] unsubscribe functionality and UI on courseplan repo

### Test Plan 

You can manually override or update the `BCC` recipients for the email sending template to include (just) yourself, and then `y`, `y` through the confirmation to have Resend send the email to your email address. Afterwards, you can confirm that styles do indeed match those under Joanna's Figma.

For testing the Firebase, you can use a dev service account key (labelled `serviceAccountKey.json` in the root folder) and run the script, modifying it to print out emails, to confirm that it properly reads users' emails in segments from the Firebase.

### Notes

TPM should test with production Firebase service account key.


scripts/email/README.md
```diff
@@ -52,14 +52,49 @@ Then in comparison to other email services like Sendgrid, Resend stood out for i
...
+]
+```
+
+This code filters the `USERS` data to create a `BCC` list of email addresses for current freshmen (in FA24) based on their expected graduation year and semester.
+
+By using this approach, we can easily create targeted email lists for different groups of students without manually maintaining separate lists.
+
 ## Further Notes
 
+- You **must** run the script from the root directory of the project.
```
nidhi-mylavarapu, line 97: Thank you for _such_ thorough documentation - future CP devs are going to appreciate this so much :)

scripts/email/execute_template.py
```diff
@@ -24,9 +24,26 @@ def send_emails(bcc_list: List[str]):
     bcc_chunks = chunk_list(bcc_list, 49)
     total_chunks = len(bcc_chunks)
 
-    print(
-        f"Starting email job with {len(bcc_list)} recipients in {total_chunks} batch(es)."
-    )
+    # Add option to list all emails
```
nidhi-mylavarapu, line 27: Just to clarify, these are print statements for cp devs to make sure the email job is running properly / commands for CP devs to follow when running the email job, correct? If so, do we get warnings if the email is not sent to certain users / can we add those checks?

scripts/email/playground/template.html
```diff
@@ -0,0 +1,90 @@
+<!-- NOTE: this is a playground file — not up to date with `current_freshman.py`. Some clients limit CSS. -->
```
nidhi-mylavarapu, line 1: HTML King 👑

scripts/email/helpers/firebase_users_loader.py
```diff
@@ -0,0 +1,76 @@
...
+firebase_admin.initialize_app(cred)
+
+db = firestore.client()
+
+
+def get_all_user_names() -> Dict[str, Tuple[str, str, str]]:
+    print("Fetching all user names from Firebase...", end="\n\n")
+    user_names = {}
+    docs = db.collection("user-name").get()
+    for doc in docs:
```
nidhi-mylavarapu, line 16: is it necessary to extract specific user information if we are just using "Hey, courseplanner" in the email header? Could be worth it to keep in the future if we decide to add personalization, but we can possibly omit this for now if it isn't needed

scripts/email/templates/versioned_preenroll/sp25/current_freshman.py
```diff
@@ -0,0 +1,136 @@
+import sys
```
nidhi-mylavarapu, line 1: do you mind clarifying the difference between this python file and the changes in scripts/email/playground/template.html ?

nidhi-mylavarapu (approved):
AMAZING work Simon!! Looks amazing, thanks for sending us the test email as well. The frontend looks great and you distributed all of the helper functions & scripts very clearly throughout the codebase. Just had a few clarifying comments for the PR. Can't wait to see this hit prod :)

scripts/email/execute_template.py
Destaq: Argh I don't get notifications for replies 💀 sorry for always being delayed in my responses here.

Yes, this is just general notifications. Added in a warning output.

scripts/email/playground/template.html
Destaq: Thanks but I wish I could do real programming more 😭

scripts/email/helpers/firebase_users_loader.py
Destaq: Good point. I'll comment this out for now.

scripts/email/templates/versioned_preenroll/sp25/current_freshman.py
Destaq: Yes certainly.

- Hosting images differently
- Some inline CSS changes, e.g. no use of negative margins as that gets stripped by Gmail
- Actual load in the template variables instead of leaving them in square brackets
- Footer updates I made due to again CSS limitations on email clients

I can resync the template to be similar to the `current_freshman.py` file, but it's really just a playground and would be annoying having to update everything in two places going forward.

PR #946, opened 2024-10-16
adding AI minor

nit0107:
### Summary 

- [x] added AI minor!

### DIFF
--- a/src/data/minors/ai.ts
+++ b/src/data/minors/ai.ts
@@ -1,0 +1,77 @@
+import { CollegeOrMajorRequirement, Course } from '../../requirements/types';
+import {
+  includesWithSingleRequirement,
+  courseMatchesCodeOptions,
+  ifCodeMatch,
+  courseMatchesCode,
+} from '../../requirements/checkers';
+import { AdvisorGroup } from '../../tools/advisors/types';
+import { lastNameRange } from '../../tools/advisors/checkers';
+
+const aiMinorRequirements: readonly CollegeOrMajorRequirement[] = [
+  {
+    name: 'Foundations of AI: Machine Learning',
+    description: 'CS 3780 or ECE 4200 or ORIE 4741 or STSCI 3740',
+    source: 'https://prod.cis.cornell.edu/undergraduate-opportunities/minors/artificial-intelligence/ai-minor-requirements',
+    checker: includesWithSingleRequirement(
+      'CS 3780',
+      'ECE 3200',
+      'ORIE 3741',
+      'STSCI 3740',
+      //TODO: add credit for prev codes ie: cs 3780 was formerly 4780
+    ),
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+    slotNames: ['Course'],
+  },
+  {
+    name: 'Foundations of AI: Reasoning',
+    description: 'CS 3700',
+    source: 'https://prod.cis.cornell.edu/undergraduate-opportunities/minors/artificial-intelligence/ai-minor-requirements',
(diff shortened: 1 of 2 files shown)


src/data/minors/ai.ts
```diff
@@ -0,0 +1,119 @@
...
+      'https://prod.cis.cornell.edu/undergraduate-opportunities/minors/artificial-intelligence/ai-minor-requirements',
+    checker: includesWithSingleRequirement('CS 3700'),
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+    slotNames: ['Course'],
+  },
+  {
+    name: 'Human-AI Interaction',
+    description: 'Add Core Course 3',
+    //TODO: add INFO 3450 exception for students grad in dec 2024 or may 2025
```
plumshum, line 43: Add a space between // and TODO.
We should to a `requirement migrations` to accommodate different major requirements based on entry year. You can check out this [documentation](https://www.notion.so/Requirement-Migrations-8f1684c570aa490bbd4dae9b4bd7bd7f) for it! This would be under a **modification**
Something along the lines of 
image

then follow the instructors for `index.ts` and running the commands in your terminal, then you should be all set!

plumshum: perhaps same idea for CS 4780/CS3780

src/data/minors/ai.ts
```diff
@@ -0,0 +1,119 @@
...
+    description: 'Add Core Course 4',
+    source:
+      'https://prod.cis.cornell.edu/undergraduate-opportunities/minors/artificial-intelligence/ai-minor-requirements',
+    checker: includesWithSingleRequirement('ENGRG 3605', 'INFO 1260', 'PUBPOL 4210'),
+    fulfilledBy: 'courses',
+    perSlotMinCount: [1],
+    slotNames: ['Course'],
+  },
+  {
+    name: 'AI Electives',
```
plumshum, line 62: @nidhi-mylavarapu make sure this cross listing (bug?) does not break the functionality. 
image

src/data/minors/ai.ts
```diff
@@ -0,0 +1,119 @@
...
+    ),
+    fulfilledBy: 'courses',
+    perSlotMinCount: [2],
+    slotNames: ['Course'],
+  },
+];
+
+export default aiMinorRequirements;
+
+// export const aiMinorAdvisors: AdvisorGroup = {
```
plumshum, line 113: instead, should we add cis.ai-minor@cornell.edu](mailto:cis.ai-minor@cornell.edu). it's not an actual person but it's  good contact information.

plumshum (commented):
Hi nice code so far! I think we should also remove some fields such as `Course` , `ifCodeMatch`, etc, since it was defined but never used. If you dont' see it on ur end on vscode, maybe it's because prettier isn't running? 
Also added some potential solutions for some of your TODOs 
There is a potential? bug for selecting the AI Electives. Will let Nidhi also look at it too. Good job!!

src/data/minors/ai.ts
nidhi-mylavarapu: thanks for the requirement migration details, hannah! yes - we would want to add something like this. make sure that the file for the AI minor has the _current_ requirements and the migration is for previous requirements

src/data/minors/ai.ts
nidhi-mylavarapu: Right, this was a weird bug -- grabbing a course from the first bucket shouldn't cause the courses in the second bucket to change to their cross listed names. I would look for another major or minor that requires you to select 2/n courses to fulfill a requirement and repeat that functionality.

PR #948, opened 2024-10-20
In schedule generator, only allow courses matching requirement to be added

Destaq:
### Summary 

Currently, in the schedule generator, if you select e.g. the "Mathematics" requirement you can still click add *any* school course to the "requirement block", even if it would not fulfill the Mathematics requirement. With the introduction of this PR, that issue is resolved — you can now only add courses to a requirement that actually fulfill that requirement.

### Test Plan 

Go to the builder page and pick a requirement.

image

Verify that only courses which would fulfill that requirement show up in the modal.

image

Note that this behavior does not apply for the "No Requirement" requirement block, as would be expected.

### Notes

We still have the UI oddity where you can select which requirement to match a course to if there are multiple options.

image

Or it could say 'not automatically fulfilling any requirement'. I think we should just get rid of this stuff in the modal as it is not relevant to schedule generator (though very relevant to the base dashboard).

Also, **please note** that the FWS example in the test plan will not show you any FWSes if you type in 'FWS'. This is because the migration script for this semester hasn't seemed to have actually found / counted FWS courses. If you look at [the Fall 24 brochure](https://fws.arts.cornell.edu/brochures/2024FA_Brochure.pdf), you'll notice that the FA24 specific courses (ones that don't happen in other semesters) do not show up in any new course modals. This is not an issue with this PR but rather with our typed course JSON consistency.

### DIFF
--- a/src/components/Modals/NewCourse/NewCourseModal.vue
+++ b/src/components/Modals/NewCourse/NewCourseModal.vue
@@ -53,19 +53,26 @@ import store from '@/store';
 import {
   getRelatedRequirementIdsForCourseOptOut,
   getRelatedUnfulfilledRequirements,
+  getFilterForRequirementFulfillment,
 } from '@/requirements/requirement-frontend-utils';
 import { specificRosterCoursesArray } from '@/assets/courses/typed-full-courses';
 import { seasonAndYearToRosterIdentifier } from '../../../user-data-converter';
 
 export default defineComponent({
   props: {
-    // TODO: filter by selectedRequirement for schedule generator
-    // selectedRequirement: { type: String, required: false, default: '' },
     year: { type: Number, required: false, default: undefined },
     season: {
       type: String as PropType<FirestoreSemesterSeason>,
       required: false,
       default: undefined,
+    },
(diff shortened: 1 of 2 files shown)


src/components/ScheduleGenerate/RequirementCourses.vue
```diff
@@ -26,19 +26,19 @@
       </button>
     </div>
     <div v-if="showDropdown">
-      <p class="requirement-header-subtext">Please select a major requirement.</p>
+      <p class="requirement-header-subtext">Please select a major/minor requirement.</p>
```
Destaq, line 29: Also added this as the possible requirement groups include those that apply to your minor.

src/components/Modals/NewCourse/NewCourseModal.vue
nidhi-mylavarapu, line 1: Thank you for this PR, Simon! I agree that we should probably remove that second UI step regarding courses that to not automatically fulfill requirements...for now, the filtering works great. Tested with a few requirements/courses. Thanks for the fix!

nidhi-mylavarapu (approved):

PR #956, opened 2024-10-30
style: update hover colors for view label

Destaq:
Fixed the hover color for the view config button.


plumshum (approved):
Nice job! Makes it much better to see.

PR #960, opened 2024-11-07
Notes Basic Frontend

nidhi-mylavarapu:
### Summary 

This pull request is the first step towards implementing the Notes Feature!

- [x] Added notes option to Course Card Menu with appropriate icon 
- [x] Created Note component and added note for **every** course card that slides out on selection and slides back in after clicking on arrow icon

Remaining TODOs:

- [ ] Align the Note text + note icon in the course card menu 
- [ ] Get correct color mapping for course card color + lighter note color
- [ ] I think Joanna was looking at alternative icons for notes so we may need to replace the current one.
- [ ] Add the sliding note card only when the note option is selected from the menu
- [ ] Adjust shadows / CIS to it's more exact - I don't think the margins are exactly the same as the Figma (but also the actual Course Cards are different dimensions in the designs so not sure what to change)
- [ ] expanded note card should make everything under it shift downward
- [ ] add arrow hover state 

### Test Plan 

Add a course card and select the note that peeks out from under the card. Type something, click the arrow to slide the note back under the card. 
- make sure that the course card has a peeking note under it that is a lighter version of the course card color. 
- if you change the color of the course card, the note changes color too
- after the note is expanded, it can only collapse after clicking the arrow icon (ask Joanna about all ways to expand the note  + collapse note)


src/components/Modals/CourseMenu.vue
```diff
@@ -1,6 +1,16 @@
 <template>
   <div class="courseMenu">
     <div class="courseMenu-content">
+      <div class="courseMenu-section" @click="openNoteModal(courseObj.code)">
+        <div class="courseMenu-left">
+          <img
+            class="courseMenu-icon"
+            src="@/assets/images/noteIconSmall.svg"
```
plumshum, line 8: I know we might change the image, but I believe the icon is a bit bigger than the other icons so it looks unaligned

src/assets/images/notes/arrow.svg
Destaq, line 1: We might want to make a transparent version of this SVG that then gets filled with the color appropriate for that course note color, as otherwise would need a set of half-a-dozen-plus various arrow SVG icons for all the possible course colors.

src/components/Course/Course.vue
```diff
@@ -1,96 +1,109 @@
...
-      v-if="isSaveCourseOpen"
-    />
-    <edit-color
-      :editedColor="editedColor"
-      @color-course="colorCourse"
-      @color-subject="colorSubject"
-      @close-edit-color="closeEditColorModal"
-      v-if="isEditColorOpen"
-    />
+  <div class="course-container">
```
Destaq, line 2: I know it was my suggestion to rearrange, but I would just be careful with the class names here post-reshuffling. It looks like the CSS logic for some of the z-indices gets in a kerfuffle on account of this change, causing the modal to be hidden under the next course card.

image

An easy fix would be to reduce the `z-index` from 1 to 0 for the course card. Can be a part of the scope of frontend fixes branched off this PR.

src/components/Course/Course.vue
```diff
@@ -1,96 +1,109 @@
...
+              <img :src="trashIcon" alt="delete course from collection" />
+            </button>
+          </div>
+          <div v-if="!compact" class="course-name">{{ courseObj.name }}</div>
+          <div v-if="!compact" class="course-info">
+            <span class="course-credits">{{ creditString }}</span>
+            <span v-if="semesterString" class="course-semesters">{{ semesterString }}</span>
             <course-caution
-              v-if="!isReqCourse && compact"
+              v-if="!isReqCourse && !isSchedGenCourse"
```
Destaq, line 75: Just for reference could you explain what this change was motivated by? Doesn't look immediately relevant to the notes feature.

src/components/Course/Course.vue
```diff
@@ -1,96 +1,109 @@
...
-      :isCompact="compact"
-      :courseColor="courseObj.color"
-      :courseCode="courseObj.code"
-      @open-edit-color-modal="openEditColorModal"
-      @delete-course="deleteCourse"
-      @edit-course-credit="editCourseCredit"
-      @open-save-course-modal="openSaveCourseModal"
-      :getCreditRange="getCreditRange || []"
-      v-click-outside="closeMenuIfOpen"
+    <Note
```
Destaq, line 98: Wow great work figuring out all of this tricky CSS stuff!

src/components/Course/Course.vue
```diff
@@ -262,27 +291,31 @@ export default defineComponent({
...
 
+.course-container {
+  position: relative;
+  padding-bottom: 20px;
+}
+
 .course {
+  box-shadow: 0px 0px 10px 4px rgba(0, 0, 0, 0.055);
+  position: relative;
+  z-index: 1;
```
Destaq, line 302: This is the `z-index` I was referencing previously — seems to mess with the course menu dropdown precedence.

src/components/Modals/CourseMenu.vue
Destaq: Yes that is the case, we'd have to rearrange all the icons sidebar style to be centered not right-aligned and have a grid or dynamic offset calculation otherwise — I'd just find a new icon honestly.

src/components/Modals/CourseMenu.vue
```diff
@@ -250,6 +260,9 @@ export default defineComponent({
       }
       return creditArray;
     },
+    openNoteModal(courseCode: string) {
+      this.$emit('open-note-modal');
```
Destaq, line 264: NB: checks not relevant as this will be used in later PRs.

src/components/Notes/Note.vue
```diff
@@ -0,0 +1,131 @@
+<template>
+  <div class="note" :class="{ expanded: isExpanded }" :style="noteStyle" @click="expandNote">
+    <div v-if="isExpanded" class="note-content">
+      <input v-model="noteText" placeholder="Add a note..." class="note-input" />
+      <img
+        src="@/assets/images/notes/arrow.svg"
+        alt="Arrow icon"
+        class="note-icon"
```
Destaq, line 8: Might want to change the naming for this to be more semantic, since we do already have a note icon in the menu dropdown.

Destaq (approved):
Nice work! A couple smaller issues, but should be good to branch off of for additional fixes and animations.

PR #962, opened 2024-11-15
fix: implement correct grouping for engineering liberal studies

import-brain:
### Summary

Currently, the engineering liberal studies requirement is represented incorrectly on CoursePlan. For the "courses from three separate groups" requirement listed on [the liberal studies policy](https://www.engineering.cornell.edu/students/undergraduate-students/advising/liberal-studies/liberal-studies-policy-students-entering), it's currently set up such that courses from any three separate liberal studies categories (not groups) check off the requirement. For example, adding SHUM 2208 (LA), ARTH 2600 (CA), and AMST 3981 (ALC) checks off the requirement, despite LA, CA, and ALC all being members of "Group 1" and thus only counting for one of the three groups under the requirement.

This pull request implements the liberal studies group structure to correct this (I'm not sure if I implemented it correctly though, there seems to be many more line deletions than additions in the requirements json). Currently, the code checks all categories (LA, CA, ALC, HA, etc.) separately.

In addition, the liberal studies groups were rearranged for Cornell engineers entering in Fall 2024, so I will make another PR soon to implement a migration for that.

CC @nidhi-mylavarapu


src/data/colleges/en.ts
```diff
@@ -10,22 +10,40 @@ import {
...
-  'FL',
-  'CE',
-  'ALC',
-  'SCD',
-  'HST',
-  'ETM',
-  'SSC',
-  'GLC',
-];
+const engineeringLiberalArtsGroups: Record<string, string[]> = {
```
Destaq, line 13: Thanks for the hard work migrating all of these categories manually! And I agree, a likely cause of this issue in the first place was poor wording with groups x categories (our fault or the school's 🤔). Good catch!

src/data/colleges/en.ts
```diff
@@ -10,22 +10,40 @@ import {
...
+  'Group 1': [
+    'CA-AAP',
```
Destaq, line 15: Hmmm I'm not sure if it is strictly necessary to go and add in the content after the hyphen for all of these categories. `hasCategory` checks for a substring match, and all the `CA`s are in `Group 1`, all the `HA`s in `Group 2` ... (they are disjoint sets) so such granularity might not be needed. @nidhi-mylavarapu not sure if you have any thoughts?

src/data/colleges/en.ts
```diff
@@ -111,37 +129,51 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
       'https://www.engineering.cornell.edu/students/undergraduate-students/advising/liberal-studies',
     checker: [
       (course: Course): boolean =>
-        engineeringLiberalArtsDistributions.some(
-          distribution => hasCategory(course, distribution) ?? false
+        Object.values(engineeringLiberalArtsGroups).some(group =>
```
Destaq, line 132: Heads up, would be more efficient to do this with a set or variable that's defined beforehand.

src/data/colleges/en.ts
```diff
@@ -111,37 +129,51 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
...
-      'Courses must be from 3 categories.': {
+      'Courses must be from 3 groups.': {
         checker: [
           (course: Course): boolean =>
-            (course.catalogDistr?.includes('LA') || course.catalogDistr?.includes('LAD')) ?? false,
-          ...engineeringLiberalArtsDistributions
-            .filter(it => it !== 'LA' && it !== 'LAD')
-            .map(distribution => (course: Course): boolean =>
-              hasCategory(course, distribution) ?? false
+            engineeringLiberalArtsGroups['Group 1'].some(
```
Destaq, line 143: Could you try mapping over the list of groups first so that it's more readable and there's less code duplication 🙏

src/data/colleges/en.ts
```diff
@@ -111,37 +129,51 @@ const engineeringRequirements: readonly CollegeOrMajorRequirement[] = [
...
+        perSlotMinCount: [1, 1, 1, 1, 1, 1],
+        slotNames: Object.keys(engineeringLiberalArtsGroups),
         minNumberOfSlots: 3,
       },
       'Courses must have at least 18 credits.': {
         checker: [
           (course: Course): boolean =>
-            engineeringLiberalArtsDistributions.some(
-              distribution => hasCategory(course, distribution) ?? false
+            Object.values(engineeringLiberalArtsGroups).some(group =>
```
Destaq, line 175: Same comment as before, but it's just nitpicking for the sake of clean code to be honest, you don't have to implement.

src/requirements/decorated-requirements.json
Destaq, line 1: This decreasing in size makes sense to me as the permutation group space (?) decreases with fewer keys.

Destaq (commented):
Hey Eric, good work catching this consistency issue. Hopefully not too many existing Engineering users were affected! There were a couple of code style things I commented on, but mainly nitpicking — great job and should be good to merge (ping or message me) once changes are in place.

src/data/colleges/en.ts
import-brain: I tried the groups without the after-hyphen content and it didn't filter correctly...not sure if it's something else though?

Destaq: Alright we can probably stick with just this then, I don't anticipate the categories changing much.

Destaq (approved):
Thanks for the updates!

PR #964, opened 2024-11-17
potential save course fix

plumshum:
### Summary 
- default collection functionality changed
- courses in default collection can only be removed by specifically removing that course from the default collection

This pull request is the first step potentially fixing save Courses default collection functionality 

### Test Plan 
- If no specific collection is chosen, then a course is only saved to the 'All/default collection'
- If a course is deleted from 'All/default collection', then every instance of that course is deleted from the specific collections
- *Note:* if a course is removed from any instance from the specific collections, that course will still remain in the 'All/default collection`
- There should never be any duplicate courses in a particular collection.

### Notes 
May not change the functionality based on future user reviews
### Blockers 

### Breaking Changes  

### DIFF
--- a/src/components/Modals/SaveCourseModal.vue
+++ b/src/components/Modals/SaveCourseModal.vue
@@ -162,6 +162,11 @@ export default defineComponent({
         collection => !this.checkedCollections.includes(collection)
       );
 
+      // If no specific collections were selected, and no prevoius selections were removed, add to 'All'
+      if (addedToCollections.length === 0 && deletedFromCollections.length === 0) {
+        console.log('added to All');
+        addedToCollections.push('All');
+      }
       this.$emit('save-course', addedToCollections, deletedFromCollections);
 
       this.closeCurrentModal();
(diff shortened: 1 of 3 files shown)


src/components/Modals/SaveCourseModal.vue
```diff
@@ -162,6 +162,11 @@ export default defineComponent({
         collection => !this.checkedCollections.includes(collection)
       );
 
+      // If no specific collections were selected, and no prevoius selections were removed, add to 'All'
+      if (addedToCollections.length === 0 && deletedFromCollections.length === 0) {
+        console.log('added to All');
```
nidhi-mylavarapu, line 167: Just to clarify the functionality: 
anytime you save _any_ course, it should end up in the "all" collection, whether you add it to another specific collection or not?

nidhi-mylavarapu: Also, on a completely separate note, should the other drop downs for collections close if you open a different one? That is, is it possible for all of the collection drop downs to be open at the same time?

nidhi-mylavarapu (changes requested):
I know this is a draft PR so there are still some more fixes but great progress so far!! thanks so much for being on top of it!

PR #970, opened 2024-12-09
frontend fixes

plumshum:
Some front-end fixes that were required by this notion task: https://www.notion.so/Front-end-fixes-1440ad723ce180a5a12edac6a3c30450
- fixed colors
- fixed hover states hopefully?
- fixed icons
- fixed note sizes
- fixed sliding + padding between courses with notes on semester
- fixed z-indexes
- **NEW SP25** semester view border moves down

### Summary 

Remaining TODOs:
- [ ] Backend + Design Spec Fixes, explaiend in this [notion file ](https://www.notion.so/Front-end-fixes-1440ad723ce180a5a12edac6a3c30450)
- [ ] Removing `nextTick()` later once the rendering is fixed.

Depends on #{number of PR}

### Test Plan 
Using svgState() and getLighterColor() is slower than just mapping specific colors to their lighter state (this is what we originall had). Need to experiment to make sure that these 2 functions don't slow time loading. 

**Bug Bashing:** Ensure the margin and padding between courses are correct
1. No notes at all
2. One course with a note (not expanded)
3. One course with a node (expanded)
4. Two or more courses, next to each other, with notes (both not expanded)
5. Two or more courses, next to each other, with notes (both expanded)
6. Two or more courses, next to each other, with notes (one of them is expanded) (x2 alternate)
7. Ensure the `+ Course` button never overflows from the SemesterView

### Blockers 

### Breaking Changes  


src/components/BottomBar/BottomBar.vue
```diff
@@ -65,7 +65,7 @@ export default defineComponent({
 .bottombar {
   display: flex;
   flex-direction: column;
-  z-index: 1;
+  z-index: 2;
```
plumshum, line 68: An issue we encountered in the front-end fixes for notes was the layering situation. 
Adjusting the z-index scores allowed the notes to appear under the courses at all time. Also, we had to adjust the z-index for the bottom bar, or else the note would appear above the bar with all the CUReview for each course.

src/components/Course/Course.vue
```diff
@@ -177,6 +177,8 @@ export default defineComponent({
     'save-note': (uniqueID: number, note: string) =>
       typeof uniqueID === 'number' && typeof note === 'string',
     'open-delete-note-modal': (uniqueID: number) => typeof uniqueID === 'number',
+    'note-state-change': (uniqueID: number, isExpanded: boolean) =>
```
plumshum, line 180: Changing the note state since the note can exist/notexist, be expanded/not expanded.

src/components/Notes/Note.vue
```diff
@@ -43,15 +61,22 @@
...
+    // NOTE: each node has an ID (just the courseID)
+    // Each note has three types of heights and translateY values: initial, expandedInitial (empty input), and expanded. For now,
+    // all translateY values are the same, but this can be changed in the future.
+    noteId: { type: String, required: true },
+    initialTranslateY: { type: String, default: '-10px' },
+    expandedInitialTranslateY: { type: String, default: '-10px' },
+    expandedTranslateY: { type: String, default: '-10px' },
+    initialHeight: { type: String, default: '20px' },
+    expandedInitialHeight: { type: String, default: '40px' },
+    expandedHeight: { type: String, default: '60px' },
```
plumshum, line 79: Another issue with the notes is that the dimensions were always changing when we expanded/not expand the note, alongside the padding & margin issues with the other components. A way to fix that is to have the actual height of the note (`initialHeight`, `expandedInitialHeight`, and `expandedHeight`) and a translation height (expanded--).

src/components/Notes/Note.vue
```diff
@@ -75,15 +100,41 @@ export default defineComponent({
...
+
+      const translateY = this.isExpanded
+        ? parseInt(trueExpandedTranslateY, 10)
+        : parseInt(this.initialTranslateY, 10);
+
+      const baseHeight = this.isExpanded
+        ? parseInt(trueExpandedHeight, 10)
+        : parseInt(this.initialHeight, 10);
+
+      const visibleHeight = baseHeight + Math.max(0, -translateY); // Add the visible part of the translated note
```
plumshum, line 128: The front-end can only see the `visibleHeight`. 
`translateY` is calculated based on the note state.

src/components/Notes/Note.vue
```diff
@@ -101,18 +152,24 @@ export default defineComponent({
 
       return `${date.getMonth() + 1}/${date.getDate()}`;
     },
+    isDisabled() {
+      // Disabled when not editing or when input is empty
+      return (!this.isEditing && this.initialNote !== '') || this.note.trim() === '';
```
plumshum, line 157: This is a specific edge case where we do not want to update/save the note if 
1. It's not in edit mode
2. Nothing is inputted into the initial note
3. The user changes the note to have nothing in the note (this is a case we forgot at first)

src/components/Notes/Note.vue
```diff
@@ -121,67 +178,96 @@ export default defineComponent({
...
+    updateSvgColor(noteId: string) {
+      const color = this.isDisabled ? 'grey' : this.color;
+      const opacity = this.isDisabled ? '0.5' : '1';
+
+      nextTick(() => {
+        const svgElement = document.querySelector(`.note-icon[data-id="${noteId}"]`);
+        if (!svgElement) return;
+        (svgElement as HTMLElement).style.setProperty('--svg-filter-color', color);
+        (svgElement as HTMLElement).style.setProperty('opacity', opacity);
+      });
```
plumshum, line 222: This function is explained the notion doc

src/components/Semester/Semester.vue
```diff
@@ -352,7 +357,20 @@ export default defineComponent({
...
+          const isExpanded = this.expandedNotes.get(course.uniqueID);
+          if (isExpanded) {
+            return acc + 3.5; // Expanded note height
+          }
+          return acc + 1.0; // Collapsed but visible note height
+        }
+        return acc;
+      }, 0);
+
+      return (this.courses.length + 1 + extraIncrementer) * factor + extraNoteHeight;
```
plumshum, line 373: This is to fix the padding issue between courses in a semester, when one of the course has a note underneath. We had to make the course height more dynamic (originally the height is always a x amount of pixels).

nidhi-mylavarapu (approved):
approving for merge (branching to do notes backend fixes)

PR #974, opened 2024-12-10
implement duplicate schedule checking in schedule generator

ejcheng:
### Summary 

This is a PR implementing duplicate checking for schedules in the CoursePlan schedule generator.

Problems, todos, and future considerations:

- [ ] Course time slots are not correct for Spring '25
- [ ] Need to add discussion/lab times
- [ ] Need a better algorithm than the current randomization one— my current approach with manually timing out the generation after 1 second isn't very optimal
- [ ] Update pagination bounds when number of possible schedules is less than 5


ejcheng:
https://www.npmjs.com/package/object-hash

src/components/ScheduleGenerate/ScheduleGenerateModal.vue
```diff
@@ -118,6 +130,7 @@ import Course, {
   Timeslot,
 } from '@/schedule-generator/course-unit';
 import Requirement from '@/schedule-generator/requirement';
+import _ from 'lodash';
```
plumshum, line 133: should move this line between line 120 and 121.

src/components/ScheduleGenerate/ScheduleGenerateModal.vue
```diff
@@ -191,7 +204,23 @@ export default defineComponent({
...
-      for (let i = 0; i < 5; i += 1) {
+      /**
+       * Checks if two `GeneratedScheduleOutput` objects `first` and `second` are equal.
+       */
+      function isEqual(first: GeneratedScheduleOutput, second: GeneratedScheduleOutput) {
+        if (first.totalCredits !== second.totalCredits || first.semester !== second.semester) {
+          return false;
+        }
+        const firstScheduleEntries = Array.from(first.schedule.entries());
+        const secondScheduleEntries = Array.from(first.schedule.entries());
```
plumshum, line 215: I believe there's a slight bug. It should be ` const secondScheduleEntries = Array.from(first.schedule.entries());` , or else you're just comparing the same entries from `first`, then this function always returns True. Let me know if that's not the case though!

plumshum: Can also add a length check to return `false` early aswell.

plumshum (changes requested):
Hi Eric. I left a few comments for this PR. I believe there is a slight bug for `isEqual`. Also, I think we should discuss how many pages we want the scheduler generator to output, because if we have more than 10+ pages, that might be too much for the user. But I also agree that we shouldn't always produce 5 pages. Good job!

src/components/ScheduleGenerate/ScheduleGenerateModal.vue
```diff
@@ -211,12 +211,46 @@ export default defineComponent({
...
+          const secondTimeslotKeys = secondTimeslots
+            .map(ts => `${ts.start}-${ts.end}-${ts.daysOfTheWeek.sort().join(',')}`)
+            .sort();
+
+          if (!_.isEqual(firstTimeslotKeys, secondTimeslotKeys)) {
+            return false;
+          }
+        }
+
+        return true;
```
plumshum, line 253: Great job with the bug fix for the first and second course. makes sense that we have to check equality between the schedules' length, the course timeslots' length, and the time-slot values themselves.

plumshum (approved):

src/components/ScheduleGenerate/ScheduleGenerateModal.vue
plumshum: also to make this merge work you should change the artifact version to pass the front-end tests

ejcheng: Thanks!

PR #976, opened 2025-02-20
Updated landing page images and modified footer/login components

NigelTatem:
Summary

This PR implements a complete redesign of the landing page with modern animations and improved user experience

[x] Implemented alternating section layouts with slide-in animations

[x] Added clear feature presentation with icons and supporting visuals

[x] Improved mobile responsiveness and content organization

[x] Enhanced visual hierarchy and content readability

[x] Added smooth scroll animations and progressive content reveal

Remaining TODOs:

- [x] punctuation fixes
- [x] laptop image cropping issue
- [x] Add some flare to the new features like celebration emojis, some design to make it more clear

Test Plan

- Tested responsive behavior across multiple devices and screen sizes
- Verified animations work smoothly in major browsers
- Confirmed mobile layout adjusts appropriately
- Checked text readability and image scaling
- Tested scroll performance with animations

Notes:

https://www.notion.so/Landing-Page-Animations-1a00ad723ce180699b92c6849f137ac2?pvs=4

PR #983, opened 2025-03-20
Nigels landing page

NigelTatem:
Look at closed pr #976


nidhi-mylavarapu (approved):

src/containers/Login.vue
```diff
@@ -122,20 +126,137 @@
...
-            class="hide schedule"
+            class="hide schedule-centered"
             src="@/assets/images/landing_new/img2.svg"
             alt="Plan preview"
             aria-label="plan preview"
+            style="position: absolute"
+          />
+        </div>
+      </div>
+
```
nidhi-mylavarapu, line 144: Thanks for keeping your code so organized! I'm glad all of the features have header comments - will make it much easier to add new features in the future.

src/containers/Login.vue
```diff
@@ -122,20 +126,137 @@
...
+            src="@/assets/images/landing_new/img7.svg"
+            alt="Notes preview"
+            aria-label="notes preview"
+            style="position: absolute"
+          />
+          <div class="comment comment-drag">
+            <img
+              src="@/assets/images/new_badge.svg"
+              alt="NEW!"
+              class="new-badge"
```
nidhi-mylavarapu, line 248: Great idea with adding the new badge!

src/components/Footer.vue
```diff
@@ -34,6 +34,9 @@ export default defineComponent({
 </script>
 
 <style scoped lang="scss">
+.container-fluid.footer {
+  margin-top: 400px;
```
nidhi-mylavarapu, line 38: In general, I think the scaling issue is because are explicitly using pixels here - we should probably calculate the spacing / margins for different devices

src/containers/Login.vue
```diff
@@ -334,15 +510,18 @@ input {
 }
 
 .new-way {
-  margin-top: 350px;
```
nidhi-mylavarapu, line 337: Changing to calculated spacing

src/containers/Login.vue
```diff
@@ -369,14 +548,94 @@ figure {
...
+    font-weight: 700;
+    font-style: normal;
+  }
+
   @media (max-width: $large-breakpoint) {
     text-align: center;
   }
+
+  // Drag and Drop section
+  &.drag-text {
```
nidhi-mylavarapu, line 568: I think this section just needs to be changed so that exact pixels aren't used and instead, they're calculated per device. I know @NigelTatem is already working on the scaling issue soon so I'm excited to see those changes and push soon!

src/containers/Login.vue
```diff
@@ -522,4 +910,166 @@ html {
...
+  &:nth-child(even) {
+    transform: translateX(200px); // Start further right
+
+    &.visible {
+      opacity: 1;
+      transform: translateX(0);
+      visibility: visible;
+    }
+  }
+
```
nidhi-mylavarapu, line 987: The sliding logic is great!! On my device, the text and images overlap because of the scaling issue, so we'll want to look out for that while testing your changes

nidhi-mylavarapu (changes requested):
Looks amazing Nigel! Can't wait to see the final landing page after the scaling bug is fixed :) Great job!!

ejcheng (commented):
Yeah, to add to Nidhi's comments, to fix the scaling issues, I think good first steps would be:

1. Instead of using the explicit pixel values (which don't scale well on mobile/different aspect ratio screen sizes), use CSS flexbox.
2. Some of the transform properties, especially when combined with absolute pixel values, can contribute to the "text off the page" issue, depending on the device you use. I'm not too well-versed in this, but I think changing your approach to not use `transform` would be better.
3. Adding media queries to explicitly target mobile devices— the current ones only apply the CSS properties under them to devices with a breakpoint of "large" or under

I think we should definitely discuss this during the next CoursePlan standup/work sesh, this is a pretty complicated frontend design and getting the CSS just right for this is going to be difficult.

PR #939, opened 2024-08-24
feat: migration for new cs major requirements

ejcheng:
### Summary 

Computer Science majors who enter Cornell in Fall 2024 have different requirements compared to those who entered Cornell before that: CS 3700 or CS 3780 is now a required course, there are only two required CS electives, and there is no Probability requirement.

This PR implements migrations in the CS major file to reflect these differences for future CS majors who entered Cornell in Fall 2024.

PR #986, opened 2025-03-23
feat: migration for new cs major requirements

ejcheng:
### Summary 

Computer Science majors who enter Cornell in Fall 2024 have different requirements compared to those who entered Cornell before that: CS 3700 or CS 3780 is now a required course, there are only two required CS electives, and there is no Probability requirement.

This PR implements migrations in the CS major file to reflect these differences for future CS majors who entered Cornell in Fall 2024.

Supersedes #939 because of weird fork prevention check

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Access CoursePlan at [courseplan.io](https://courseplan.io)!
 
-CoursePlan is a four-year academic planner for Cornell undergraduates developed by the Digital Tech & Innovation. CoursePlan helps undergraduates track their courses and their requirements automatically depending on their college, major, and minor. It aims to allow students view the big picture of their time at Cornell.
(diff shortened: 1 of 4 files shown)


plumshum:
`and there is no Probability requirement`
New CS majors will never understand the probability trauma now 😿

plumshum:
Switched between FALL23 AND FALL24 and requirements changed accordingly.
Not sure why there is still double count for CS 3700/4700 and CS 3780/CS4780

plumshum (approved):
I think migration is all set. A bit annoying that CS 3700and 3780 corresponding 4XXX also shows up but that'll probably change with new semesters coming around 
Tested to make sure different entry years worked.

ejcheng:
> I think migration is all set. A bit annoying that CS 3700and 3780 corresponding 4XXX also shows up but that'll probably change with new semesters coming around Tested to make sure different entry years worked.

Thanks!

PR #993, opened 2025-04-09
Fall 2025 Pre Enroll Giveaway

nidhi-mylavarapu:
### Summary 

This pull request is to implement the Fall 2025 Pre-Enroll Giveaway. 
- [x] Implement the initial giveaway modal
- [x] Implement the progress circle
- [x] Connect the progress circle with the user's actions
- [x] Create "Eligible to Enter Giveaway" and "Giveaway Entered" Modal

### Test Plan 
- [ ] Ensure that giveaway modal is easy to exit / does not show up repeatedly upon login
- [ ] Make sure progress bar updates as the user completes the steps
- [ ] "Eligible to Enter Giveaway" Modal shows up right after last step is completed
- [x] Make sure they can't enter giveaway more than twice (after they enter, the giveaway modal should be "Giveaway Entered"


ejcheng (dismissed):
Upon testing, I noticed the following things that we might want to change:

1. The modal appears every time I refresh the page.
2. Would it be possible to find a confetti animation with a higher "frame rate"? I feel like the animation looks a bit choppy right now.
3. It would be nice if the netid/instagram entry modal popped up immediately after the user generates a schedule (probably overlaid on top of the generated schedules modal)— I feel like some users might not notice they completed the three steps with the current implementation.

As for the fourth checkbox, I don't think users would be able to fill out the form to enter more than once if `entered` is true in Firebase, so your implementation is good.

However, as for the second checkbox, would it be possible to "straighten" the third of the circle that is blue in the progress indicator when the user has only completed one of the three steps. The logic for advancing the progress indicator works correctly, but I feel like the "diagonal offset" of the 1/3-filled-in circle looks a bit off.

ejcheng (approved):
LGTM!

PR #995, opened 2025-04-14
WIP: implement quality-of-life changes for pre-enroll

ejcheng:
See https://cornelldti.slack.com/archives/C05QHNY8TKL/p1744242627568859

Thanks @ambers7!

### DIFF
--- a/src/components/Requirements/RequirementSideBar.vue
+++ b/src/components/Requirements/RequirementSideBar.vue
@@ -174,6 +174,29 @@
                 </div>
               </template>
             </draggable>
+            <div class="py-3">
+              <div class="see-all-pages" v-if="numPages > 1">
+                <span class="see-all-pageCount">{{ pageText }}</span>
+                <div class="see-all-buttonWrapper">
+                  <button
+                    class="see-all-button"
+                    :class="{ 'see-all-button--disabled': !hasPrevPage }"
+                    :disabled="!hasPrevPage"
+                    @click="prevPage()"
+                  >
+                    <span class="see-all-button-text">Prev</span>
+                  </button>
+                  <button
+                    class="see-all-button"
+                    :class="{ 'see-all-button--disabled': !hasNextPage }"
+                    :disabled="!hasNextPage"
+                    @click="nextPage()"
+                  >
+                    <span class="see-all-button-text">Next</span>
+                  </button>
+                </div>
+              </div>
+            </div>
           </div>
         </div>
       </div>


src/components/BottomBar/BottomBarTab.vue
```diff
@@ -1,7 +1,11 @@
 <template>
   <button
     class="bottombartab full-opacity-on-hover"
-    :style="{ background: `#${color}` }"
+    :style="{
+      background: `#${Math.floor(Math.random() * 0xffffff)
+        .toString(16)
+        .padEnd(6, '0')}`,
+    }"
```
ejcheng, line 8: @cornell-dti/course-plan What do you think of this approach to ensure that no two tabs have the same color? It seems pretty questionable to me, but this is the best approach I could come up with for now.

src/assets/scss/_variables.scss
```diff
@@ -39,7 +39,7 @@ $backgroundBlue: #f6fafc;
 $primaryGray: #3d3d3d;
 $secondaryGray: #858585;
 $activeGray: #5b676d;
-$disabledGray: #cccccc;
+$disabledGray: #78848a;
```
nidhi-mylavarapu, line 42: this changes disabledGray everywhere - was the incorrect hexcode being used originally according to design?

src/components/BottomBar/BottomBarTab.vue
nidhi-mylavarapu: I agree it's a bit questionable - since it’s random, there’s still a chance of collision (and color contrast might be inconsistent too). if we really want to ensure uniqueness and consistent colors we could:
- generate colors from a fixed palette (e.g. COLORS = ['#FFAA00', '#4DA8DA', ...]) and assign them based on tab index or hash of the course ID
- or use a seeded random function so the same course/tab always gets the same color
that said, for a quick temporary solution this totally works - just maybe worth revisiting if it ends up looking chaotic in prod haha

src/components/Course/Course.vue
```diff
@@ -214,6 +216,18 @@ export default defineComponent({
...
+    requirementsString(): string {
+      let requirementsString = '';
+      this.requirements.forEach(req => {
+        requirementsString += `${req}, `;
+      });
+      if (requirementsString.length > 0) {
+        return requirementsString.substring(0, requirementsString.length - 2);
+      }
+      return 'testingRequirementsString';
+      // todo: if requirements string is too long, the course card text gets crunched together
```
nidhi-mylavarapu, line 228: For the requirementsString length issue - maybe we could:

Limit the number of requirements shown (e.g. 2 or 3), and if there are more, add a +X more indicator (e.g. "MATH, CS, +2 more")

or switch to rendring them as separate badges (like small pill components) which could wrap or scroll more gracefully on smaller course cards

Either way we might also want to set a max-width or text-overflow: ellipsis style if we’re sticking with plain text for now

nidhi-mylavarapu (changes requested):
For the TODOs:

For RequirementFulfillmentGraph.getConnectedRequirementsFromCourse(courseId), we could call this when rendering each course card, and pass the result down as a fulfilledRequirements prop to display somewhere (maybe in a tooltip or under the title?).

for the "master requirements page", maybe we can reuse the existing paginated course list component logic, but instead group/filter by unfulfilled requirements as the top-level sections instead of course categories

PR #1000, opened 2025-04-16
Eliminate Schedule Generator Duplicates

nidhi-mylavarapu:
### Summary 
Look at PR #998

PR #1001, opened 2025-04-19
Eliminate Schedule Generator Duplicates (#1000) Into Release

nidhi-mylavarapu:
merging improved schedule gen algorithm that removes duplicate schedules 
creds @ejcheng :)


plumshum (approved):
Same review as earlier for schedule checker 😁

PR #1012, opened 2025-08-27
feat: multiline notes

Destaq:
Added multiline notes to improve flexibility 🎉

### DIFF
--- a/src/components/Notes/Note.vue
+++ b/src/components/Notes/Note.vue
@@ -186,7 +186,7 @@ export default defineComponent({
       setTimeout(() => {
         // Grant some rendering time so that the height is figured out.
         this.$emit('height-change');
-      }, 50);
+      }, 100);
     },
     saveNote() {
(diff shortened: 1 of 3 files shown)


plumshum:
When creating a new note or editing a new note, the upload and check button are not in the right place
Screenshot 2025-08-27 130817

image

src/components/Notes/Note.vue
```diff
@@ -248,16 +276,19 @@ export default defineComponent({
   transform: translateY(100%);
   transition: opacity 0.3s ease, transform 0.3s ease;
```
plumshum, line 277: I think need something like `align-items: end` so that the checkbox and upload button are at the bottom

Destaq: Oh interesting that is cleaner than my solution, I will go test this change.

plumshum (approved):

PR #1028, opened 2025-12-05
Tooltip to show requirement a Semester course fulfills

plumshum:
When a user hovers over a course in the Semetser view, they can see what requirements and requirement Group it fulfills. 

image

- Fulfills <Requirement> (Requirement Group)

- If there are multiple requirements it is all listed out 

- All requirement groups are "College", "Major", "Minor" and "Grad"

- If the course IS NOT matched to any requirement the hover should not show up

Remaining TODOs:

- [x] horrible horrible z-index issues 😭 
- [x] final design review 

Depends on #{number of PR}

### Blockers 

### Breaking Changes  


plumshum:
**Note!!** Need to revamp designs again but the logic should be the same

src/components/Course/Course.vue
NigelTatem: Method for finding tips looks great here.

NigelTatem (approved):
This feature is a great quality-of-life boost for the Semester view. The hover instantly answers a question students ask all the time “what requirement does this actually count for?” and the way you list multiple fulfilled requirements feels clear and easy to scan. 

The logic for pulling requirement info is straightforward and ties neatly into the existing report structure. The only real polish left is the z-index layering so the tooltip doesn’t get lost behind other UI pieces, but that’s expected for this kind of element.

PR #1034, opened 2026-09-18
add climate change minor

jjasonguo:
### Summary 

This pull request is the first step towards implementing the climate change minor

-added the climate-change.ts requirements file with the relevant courses, descriptions, modules and registered it into index.ts
- regenerated decorated JSON

### Test Plan 
 1. created profile with climate change minor
 2. make a schedule that satisfies/doesn't satisfy the upper-level credits, category courses, and total credits to ensure proper functionality.
 

NOTE: BEE 2000 and EVS 2001 are the same class just crosslisted (courseid 367870 and the same title)
image


gselina (commented):
Screenshot 2026-09-18 at 5 01 12 PM
Tried climate change minor on my end. Looks good to me!
Just a note that LAW 4444 doesn't seem to be in full-courses.json yet so it won't count toward this requirement until course data is regenerated later.

